PR #128, opened 2012-04-02
added sonata.media.admin.groupname parameter to admin config

tamayg:


OskarStark:
can we close this? the change itself just change the file permissions

PR #277, opened 2013-02-21
Fix ImageProvider and FileProvider bugs on generating private url for reference

tiagojsag:
(lets see if third is the charm)
MediaProviderInterface\getDownloadResponse() accepts several methods for downloading files, using 'http' by default. When methods other than "http" are used, it uses generatePrivateUrl() to get the file path, which is bugged in two places:
- FileProvider always returns false, which is wrong, as the file does exist. It should only return false only if a thumbnail (!= reference) is requested
- ImageProvider always returns a thumbnail, even if "reference" is being requested.

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -253,7 +253,11 @@ class FileProvider extends BaseProvider
      */
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        return false;
+        if ($format == 'reference') {
+            return $this->getReferenceImage($media);
+        } else {
+            return false;
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


NeuralClone:
This issue is fixed in FileProvider but it still appears to be a problem in ImageProvider (master). The problem is exactly the same as @tiagojsag described above: ImageProvider always returns a thumbnail even if the "reference" format is being requested. This causes downloads to fail if a method other than http is used.

NeuralClone:
This issue is fixed in FileProvider but it still appears to be a problem in ImageProvider (master). The problem is exactly the same as @tiagojsag described above: ImageProvider always returns a thumbnail even if the "reference" format is being requested. This causes downloads to fail if a method other than http is used.

OskarStark:
Fixed in ``master`` through #920

PR #374, opened 2013-08-03
sets providerMetadata doctrine type as 'json_array' instead of unsupport...

javigzz:
Hi,

I was generating the schema for my database and got a dbalexception saying:

[Doctrine\DBAL\DBALException]
  Unknown column type "json" requested.

So I this fixed for me: changing it to 'json_array' according to Doctrine\DBAL\Types\Type

regards
javier


fnagel:
This seems undocumented, at least in https://sonata-project.org/bundles/media/master/doc/reference/installation.html

One needs to add
````
doctrine:
    dbal:
        # ...

        types:
            json: Sonata\Doctrine\Types\JsonType
````
to projects config.yaml.

Issue #370, opened 2013-07-30
Resizer per format


marcinziolo:
Don't you think that resizer should be assigned to concrete formate in formats section ?

For example I would like to create two thumbnails with different resizer for one media object.

  formats:
    simple: { width: 100, height: 80, resizer: sonata.media.resizer.simple}
    exact_size:   { width: 200, height: 150, resizer: my.bundle.resizer.custom} //first crop and then resize in order to get 200x150 thumbnail

Now I can only assign resizer per all thumbnails what isn't sufficient for my requirements.

aivus:
:+1:

bertrandg:
:+1:

djaney:
:+1:

ruscon:
:+1:

malenkiki:
It is logical and must have feature. I dont understand, that since 2013, it is not yet done.

OskarStark:
> It is logical and must have feature. I dont understand, that since 2013, it is not yet done.

We are an Open-Source project and if you need it, you can provide a PR, thats how it works. If its not done yet, the needs seems to be NOT that "must have"....

greg0ire:
@malenkiki if you want to understand, find out why _you_ didn't contribute it, then assume everyone had similar reasons.

andreasschacht:
@OskarStark @greg0ire I agree that everybody should provide a solution instead of complaining about, but there was already a PR #423  for this. It was closed for a reason I can not follow...
It looks like, provide additional data in this place isn't a good solution at all for you. Or is there a chance to reopen this?

greg0ire:
Hmmm I don't know if I understand the problem correctly because I don't use this bundle, so please tell me if I'm wrong:

- there are formats, that map to resizer services, that are called here: https://github.com/sonata-project/SonataMediaBundle/blob/d7c493472d15d352fd05835041100d67ddb7db8c/Thumbnail/FormatThumbnail.php#L82
- you would like to be able to call your own resizer service, but be able to pass it some more options.

In the case of the watermark, adding a watermark can't really be called "resizing", so maybe this is not the right abstraction?
This sounds awefully like a think the liipimaginebundle would be able to do, and we provide integration with it: https://github.com/sonata-project/SonataMediaBundle/blob/fabd14b939267ab9ef622963f14aed1cf26caa4c/Resources/doc/reference/extra.rst#liip-imagine-bundle-integration

It provides filters instead of resizers, so maybe use just that?

[...]

PR #423, opened 2013-10-28
Add "extra" configuration field in "formats" section

da-eto:
For custom resizer settings.
Mybe useful in situations like issue #370 when we need to implement custom logic in our resizer.


bertrandg:
+1 for the right solution

andreasschacht:
@rande 
I think there are good reasons to provide "extra"-data, can we reopen this PR?
See  https://github.com/sonata-project/SonataMediaBundle/issues/370#issuecomment-328792869

PR #458, opened 2014-01-03
Fix: Merge user form model data into media object copy

xaben:
Upon creating a new user provider that needs an additional form field besides "binaryContent", the bound model data for that field is discarded while the "newMedia" copy contains a null field. This fix, merges the user defined properties into the "newMedia" object.

This error happens only if "new_on_update" is true, which is the default behavior.


core23:
Please rebase @xaben

benglass:
@core23 @rande 

The current implementation of ProviderDataTransformer defaults to behavior that makes it so you cannot add your own properties to Media.

For example we added a url field to Media and the value is lost whenever the binaryContent is changed because the ProviderDataTransformer by-default creates a new media object whenever binaryContent is changed and does not copy over all properties besides the core media bundle ones.

I am not sure why this behavior is the default, but it is also not something that be turned off in any obvious way.

PR #494, opened 2014-02-17
Remove doctrine managers

EmmanuelVella:
It seems there is no reason to keep these managers ?

### DIFF
--- a/Admin/Manager/DoctrineMongoDBManager.php
+++ b/Admin/Manager/DoctrineMongoDBManager.php
@@ -19,6 +19,8 @@ use Symfony\Bridge\Doctrine\ManagerRegistry;
 /**
  * this method overwrite the default AdminModelManager to call
  * the custom methods from the dedicated media manager.
+ *
+ * @deprecated This class is deprecated since version 2.4 and will be removed in 3.0.
  */
 class DoctrineMongoDBManager extends ModelManager
 {
--- a/Admin/Manager/DoctrineORMManager.php
+++ b/Admin/Manager/DoctrineORMManager.php
@@ -21,6 +21,8 @@ use Symfony\Bridge\Doctrine\RegistryInterface;
 /**
  * this method overwrite the default AdminModelManager to call
  * the custom methods from the dedicated media manager.
+ *
+ * @deprecated This class is deprecated since version 2.4 and will be removed in 3.0.
  */
 class DoctrineORMManager extends ModelManager
 {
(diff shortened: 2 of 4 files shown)


core23:
Please rebase @EmmanuelVella

EmmanuelVella:
@core23 Rebase done (but I can't test as I'm not using this bundles anymore).

core23:
Thanks @EmmanuelVella 
 
Maybe we can leave both files and add a deprecation notice instead?

EmmanuelVella:
@core23 That a good idea. I have updated my PR.

Admin/Manager/DoctrineMongoDBManager.php
```diff
@@ -19,6 +19,8 @@
 /**
  * this method overwrite the default AdminModelManager to call
  * the custom methods from the dedicated media manager.
+ *
+ * @deprecated
```
core23: Can you please leave a comment here?

``@deprecated since XXX (to be removed in 3.0)``

EmmanuelVella:
Done !

core23:
Users need to know since when they are deprecated and when they will be removed. @EmmanuelVella

core23:
ping @EmmanuelVella

EmmanuelVella:
I have updated the deprecated message.

Admin/Manager/DoctrineMongoDBManager.php
```diff
@@ -19,6 +19,8 @@
 /**
  * this method overwrite the default AdminModelManager to call
  * the custom methods from the dedicated media manager.
+ *
+ * @deprecated This class is deprecated since version 2.4 and will be removed in 3.0.
```
greg0ire: 2.4 does not exist.

Soullivaneuh:
@EmmanuelVella According to https://github.com/sonata-project/SonataAdminBundle/issues/3731, sonata media bundle won't be 2.4 but 3.0.

After that, if those manager are really not used, I think we could accept this BC break for 3.0.

I'll ping you when the 3.x will be ready so you will be able to rework this PR. :+1:

EmmanuelVella:
Ho ok I didn't know there won't be a 2.4 version. Waiting for your ping then !

Soullivaneuh:
Thank you @EmmanuelVella, I cherry-picked your commit with deletion instead of deprecation.

https://github.com/sonata-project/SonataMediaBundle/commit/49e628ec497179130988c8aefe83f45d891852a4

EmmanuelVella:
ok :+1:

Issue #480, opened 2014-02-03
Blurred images due to Resizer bug?


xaben:
While trying to resize a 400x224 image to a 96px width constraint, the resizer creates a 95x53 image instead of the 96x54 expected.

And to make matters worse, the twig helper sets the width attribute, making the image 96px (browser scaled) which is very blurry.

What would be the way to fix this? Create a custom resizer? Fix twig helper to return actual thumbnail size? Revert to own <img> tag with path and no media attributes?

xaben:
The problem is still there, but I found that my solution isn't perfect. Especially if the reference image is of a different size then the media (metadata). If anyone has any ideas how to fix this...

core23:
This issue is very old. Can you please check this against the latest stable / master version? 

Feel free to reopen, if the error still exists. 

A PR would be welcome if you reopen this issue.

PR #501, opened 2014-02-26
Fix SimpleResizer image ratios when one of the parameters isn't set.

xaben:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #480
| License       | MIT

When scaling images, the ratio is calculated of both width and height. If one of these isn't set in the context format setting, is is calculated and rounded to an integer value. The ratio is then calculated of this rounded value which gives unexpected results as shown in #480 issue.
Fixes test too, as scaling a 567x200 image to 90px width should return a 90x32 image and not 88x31 (easily checked in any image editor).

### DIFF
--- a/Resizer/SimpleResizer.php
+++ b/Resizer/SimpleResizer.php
@@ -13,9 +13,11 @@ namespace Sonata\MediaBundle\Resizer;
 
 use Imagine\Image\ImagineInterface;
 use Imagine\Image\Box;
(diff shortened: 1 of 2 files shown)


core23:
Please rebase @xaben

PR #582, opened 2014-06-20
Move phpcr mappings out of default location to avoid conflicts with mongdb/allow auto_mapping

### LINKED ISSUE #509: Problem with sonata.media.manager.media service and doctrine_phpcr db_driver
I dont know if this is becouse of changes in #494 but I get 

```
Argument 2 passed to Sonata\\CoreBundle\\Model\\BaseManager::__construct()
must implement interface Doctrine\\Common\\Persistence\\ManagerRegistry,
instance of Sonata\\DoctrinePHPCRAdminBundle\\Model\\ModelManager given
```

when I try to use db_driver: doctrine_phpcr for sonata_media.

If I modify the sonata.media.manager.media like this:

```
<service id="sonata.media.manager.media" class="%sonata.media.manager.media.class%">
            <argument type="service" id="sonata.media.pool" />
            <argument type="service" id="doctrine_phpcr" />
            <argument>%sonata.media.media.class%</argument>
        </service>
```

(using doctrine_phpcr service instead of sonata.admin.manager.doctrine_phpcr)
It get past that problem but then I get

```
The class 'Application\Sonata\MediaBundle\Entity\Gallery' 
was not found in the chain configured namespaces
```

Why is it searching for an entity when I use doctrine_phpcr?

benglass:
Fixes #509

I think its possible users could see the same conflict when they run easy extends because it would create application classes in the Document namespace for mongo but these would not be needed (the user can delete these if they are using PHPCR) and therefore easily circumvent the problem, although whether it would be clear to users that they needed to do this is unlikely (they are going to get a generic error saying the Application/MediaBundle/Document/Gallery has invalid mappings). On the other hand there is no reason to commit Document classes to your repo that you arent even using so if users are paying attention they can avoid the issue. Still seems problematic, perhaps something for the easy extends documentation? Or the media bundle docs?


core23:
Are you still working on this? Please squash and rebase or close this PR @benglass

Soullivaneuh:

Regards.

PR #601, opened 2014-07-24
Fix errors upload_max_filesize, wrong mimetypes for SonataAdmin

antoinemineau:
Hi.

I found a way to fix the issue when you upload a file bigger than the upload_max_filesize, and also on invalid mimetypes.

I added an exception if the clientsize is at  zero bytes, i'm guessing that it's not uploaded because of upload_max_filesize value. I had to set a providerReference because there is a validation on this field: can't be null or blank  (if there is no providerReference set, on top of the form it displays: this field cant be null, this field cant be blank).  I am catching this exception in the BaseProvider. I added in the validate method an error violation for the client size at zero bytes. The wrong mimetypes already had an error violation related, the issue was exception not being caught. I had to modify the last if statement so when the file isn't uploaded there is no check on the non existing file mimetypes.

Might not be the prettiest way to do it, but there is no more 500 errors.

Screenshot of file bigger than the upload_max_filesize:
capture decran 2014-07-24 a 17 21 42 2

Screenshot of wrong mimetypes:
capture decran 2014-07-24 a 17 22 03 2

### DIFF
--- a/Provider/BaseProvider.php
+++ b/Provider/BaseProvider.php
@@ -71,7 +71,10 @@ abstract class BaseProvider implements MediaProviderInterface
             return;
         }
 
-        $this->doTransform($media);
+        try {
+            $this->doTransform($media);
+        } catch (\Exception $e) {
+        }
     }
 
     /**
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -218,6 +218,12 @@ class FileProvider extends BaseProvider
         $this->fixBinaryContent($media);
         $this->fixFilename($media);
 
+        if ($media->getBinaryContent()->getClientSize() == 0) {
+            $media->setProviderReference(sha1($media->getName() . rand(11111, 99999)));
+            $media->setProviderStatus(MediaInterface::STATUS_ERROR);
+            throw new UploadException('The uploaded file is not found');
+        }
+
         // this is the name used to store the file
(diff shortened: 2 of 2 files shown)


Filoz:
This is really useful! :+1:

Provider/BaseProvider.php
```diff
@@ -71,7 +71,10 @@ public function __construct($name, Filesystem $filesystem, CDNInterface $cdn, Ge
             return;
         }
 
-        $this->doTransform($media);
+        try {
+            $this->doTransform($media);
+        } catch (\Exception $e) {
+        }
```
rande: Can you log the exception ?

antoinemineau:
When the file is bigger than the upload_max_filesize, it returns this exception : https://github.com/antoinemineau/SonataMediaBundle/blob/c19bf05c328e1b0326ec6d0726bb74a071e65f02/Provider/FileProvider.php#L224

When the file has a wrong mimetypes it returns :
Notice: Undefined variable: image in /Applications/MAMP/htdocs/ideastim/vendor/sonata-project/media-bundle/Provider/ImageProvider.php line 105

Not sure why it returns this undefined variable. Instead of throwing exception, we could use a return to stop the function call and remove the try/catch from BaseProvider.

borNfreee:
any news?

Soullivaneuh:

Regards.

PR #609, opened 2014-08-07
api_form and serializer require JMSSerializerBundle

Laulibrius:
The api_form require jms_serializer.metadata_factory service defined by JMSSerializerBundle


core23:
Fixed with https://github.com/sonata-project/SonataMediaBundle/pull/753/

PR #617, opened 2014-08-29
Adding a new option to sonata_media_type

davidfernandezperrino:
The idea is to be give the developer the chance to use or not the unlink checkbox of the sonata_media_type without overriding the entire type. For that, we can pass to the admin the option 'unlink_enabled' (true by default). Changing it to false, will not add that checkbox to the form.


core23:
Are you still working on this? Please squash and rebase or close this PR @davidfernandezperrino

PR #666, opened 2014-11-18
Ensure element unlink exist

cbastienbaron:
The form type media_type can be manipulated with events, inherits etc 

Now we can remove element unlink without have 

Child "unlink" does not exist. - 500 Internal Server Error - OutOfBoundsException


sebastienFidesio:
need this one

cbastienbaron:
thks Thomas !

PR #687, opened 2014-12-31
[WIP] Added CloudFront to list of available CDNs

phansys:
Added [CloudFront](https://aws.amazon.com/cloudfront) to list of available CDNs
Any suggestion or improvement proposal are welcome.

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | yes
| Deprecations? | no
| Tests pass?   | yes
| Doc PR        | none

TODO:
- [x] Unit testing
- [x] Update model in order to persist flush identifier
- [x] Create command to update flush status
- [x] Update translation catalogs

### DIFF
--- a/CDN/CDNInterface.php
+++ b/CDN/CDNInterface.php
@@ -34,7 +34,7 @@ interface CDNInterface
      *
      * @param string $string
      *
-     * @return void
+     * @return void|string
      */
     public function flush($string);
 
@@ -43,7 +43,7 @@ interface CDNInterface
      *
      * @param string $string
      *
-     * @return void
+     * @return void|string
      */
     public function flushByString($string);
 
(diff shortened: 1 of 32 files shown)


phansys:
I have too some ideas about CDN usage / integration. IE, consider the following snippet attached to a lifecycle event or something similar:
```php
#...
if ($media->getCdnIsFlushable()) {
    $flushPaths = array();
    foreach ($provider->getFormats() as $format => $settings) {
        $flushPaths[] = $provider->getFilesystem()->get($provider->generatePrivateUrl($media, $format));
    }
    $provider->getCdn()->flushPaths($flushPaths);
    $media->setCdnFlushAt(new \DateTime());
}

```

This also should take care about ```new_on_update``` form setting and only flush the CDN paths when it is set as ```false```, because the "magically generated" new medias are fresh and aren't cached yet while the previous ones has kept immutables.

What do you think about?

phansys:
I've rebased PR, squashed commits and added this logic in [BaseProvider::transform()](https://github.com/sonata-project/SonataMediaBundle/pull/687/files#diff-a2b4c180284030170d499212621ba110R75) method.
Any concern about this feature will be great.

/cc @rande.

PR #704, opened 2015-01-29
Update doctrine_phpcr_admin.xml

icevita:


Soullivaneuh:

Regards.

OskarStark:
@dbu can you verify this? if yes i will create a new PR for this

dbu:
i am not familiar with the media manager. a similar service exists for orm and there we have the category manager. i tried to find if the category manager is defined regardless of the doctrine flavour, but did not find where that service is defined (maybe in a different bundle?). if the service will be available when we only have the phpcr storage, then injecting it seems reasonable.

PR #710, opened 2015-02-10
Prevent "undefined index: context" error for batch delete action

sfavot:
Check if key 'context' exists in filter to prevent "undefined index: context" error from appearing when deleting several media at a time in admin.


xavierlacot:
:+1:

PR #712, opened 2015-02-12
[PHPCR] size nullable

Peekmo:
Hello,

To be able to use Youtube video, size must be nullable (because YT service does not return the size)
The modification seems to me done for ORM but not PHPCR

Regards,
Peekmo


Peekmo:
Hello,
Any news about ?

PR #730, opened 2015-03-11
Updated FileProvider.php for translation

krewetka:
Right now there is now way to translate this message in validators.[lang].xliff

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -404,7 +404,7 @@ class FileProvider extends BaseProvider
         if (!in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
             $errorElement
                 ->with('binaryContent')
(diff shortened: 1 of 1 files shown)


Provider/FileProvider.php
```diff
@@ -404,7 +404,7 @@ public function validate(ErrorElement $errorElement, MediaInterface $media)
         if (!in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
             $errorElement
                 ->with('binaryContent')
-                ->addViolation('Invalid mime type : ' . $media->getBinaryContent()->getMimeType())
+                ->addViolation('Invalid mime type : %type%' ,array("%type%"=>$media->getBinaryContent()->getMimeType()))
```
rande: Can you fix CS ?

krewetka: Any better now?

krewetka: @rande - is it ok for you now?

PR #735, opened 2015-03-20
request the sonata classification root categories mapped by context name...

bardius:
... instead of the default, that is id, as the context for each media item is set by name and this causes incompatibility between the two bundles.

this PR is tightly associated with the Rl of Sonata Classification Bundle https://github.com/sonata-project/SonataClassificationBundle/pull/98 as both bundles have to be amended to solve the issue


bardius:
Apparently the context table id field is a varchar so its values can be the same with the name. Thus the fix is not needed.

PR #739, opened 2015-03-23
Sync thumbnails command optimization

MAXakaWIZARD:
I encountered such problem - when trying to sync large amount of thumbnails (6000+) command gets aborted by memory limit, cause it fetches all medias at once in one batch. This PR introduces medias processing in batches (batch size can be changed via --batchSize option and equals to 100 by default). So now this command can process as many thumbnails, as we have.

### DIFF
--- a/Command/SyncThumbsCommand.php
+++ b/Command/SyncThumbsCommand.php
@@ -38,7 +38,7 @@ class SyncThumbsCommand extends BaseCommand
             ->setDefinition(array(
                 new InputArgument('providerName', InputArgument::OPTIONAL, 'The provider'),
                 new InputArgument('context', InputArgument::OPTIONAL, 'The context'),
-                new InputOption('batchSize', null, InputOption::VALUE_REQUIRED, 'Media batch size', 100)
+                new InputOption('batchSize', null, InputOption::VALUE_REQUIRED, 'Media batch size (100 by default)', 100)
             )
         );
     }
--- a/Command/SyncThumbsCommand.php
+++ b/Command/SyncThumbsCommand.php
@@ -13,6 +13,7 @@ namespace Sonata\MediaBundle\Command;
 
 use Symfony\Component\Console\Input\InputArgument;
 
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Input\InputInterface;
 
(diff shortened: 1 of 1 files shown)


rande:
@MAXakaWIZARD I think you can also clear the Doctrine Unit Of Work. This will save more memory.

MAXakaWIZARD:
@rande Yep, working on this at the moment.

MAXakaWIZARD:
@rande This doesn't work (memory consumption increases with each batch):
```php
$this->getContainer()->get('doctrine')->getManager()->clear();
```
Does sonata media use different entity manager (non-default)?

MAXakaWIZARD:
This returns true for each media, so this is right entity manager:
```php
$this->getContainer()->get('doctrine')->getManager()->contains($media);
```

MAXakaWIZARD:
As I see, memory leak is somewhere in those calls (cause if I comment them out, memory usage is stable and is not growing):
```php
$provider->removeThumbnails($media);
$provider->generateThumbnails($media);
```

rande:
The generateThumbnails does not create state variables, so it either a leak in the gd lib, or a variable reference somewhere.

MAXakaWIZARD:
@rande problem is in `Gaufrette\Filesystem` - it stores all files in `fileRegister` and it never gets cleared, and there is no method for that. Since it holds files with content, it runs out of memory pretty fast.
I see such solutions:
1) extend `Gaufrette\Filesystem` class with `clear` method and call it after each resize
2) try to contribute to `Gaufrette` and introduce that method

rande:
@MAXakaWIZARD option 3 ... use the reflection api to make the property public and clear the array ...

MAXakaWIZARD:
@rande not very good, but is OK as a temporary solution

rande:
you need to do that in the command only, you can add a big notice about the issue.

MAXakaWIZARD:
Implemented those methods here: https://github.com/KnpLabs/Gaufrette/pull/329
Will do temporary workaround since that PR may not be merged for months.

MAXakaWIZARD:
Fixed those issue.
But we have another one)
When one format gets removed from context in config, sync-thumbnails does not delete old thumbnails in old formats. I think we need delete all files matching this mask: `thumb_{mediaId}_*.*`

rande:
I think this require a dedicated command to remove a specific format.

MAXakaWIZARD:
@rande Good idea. I will do that in another PR. Could you please merge this?

MAXakaWIZARD:
I think this PR can be closed now...

rande:
@MAXakaWIZARD the PR is closed as I merged it

PR #740, opened 2015-03-23
Compatible with SonataCoreBundle and SonataAdminBundle

imilyukov:
Compatible with SonataCoreBundle and SonataAdminBundle


rande:
use the 2.4-dev release.

PR #741, opened 2015-03-24
Implement remove-thumbnails command

MAXakaWIZARD:
With this command we can remove thumbnails of specified format or even all thumbnails.
This one will remove all thumbs of `small` format from `default` context:
```bash
app/console sonata:media:remove-thumbnails sonata.media.provider.image default small
```
This one will remove all thumbs from `default` context:
```bash
app/console sonata:media:remove-thumbnails sonata.media.provider.image default all
```
Also, command gives ability to choose provider, context and format, if any of them are not specifed in command line.

### DIFF
--- a/Thumbnail/FormatThumbnail.php
+++ b/Thumbnail/FormatThumbnail.php
@@ -84,10 +84,19 @@ class FormatThumbnail implements ThumbnailInterface
     /**
      * {@inheritdoc}
      */
-    public function delete(MediaProviderInterface $provider, MediaInterface $media)
+    public function delete(MediaProviderInterface $provider, MediaInterface $media, $formats = null)
     {
-        // delete the differents formats
-        foreach ($provider->getFormats() as $format => $definition) {
+        if (is_null($formats)) {
+            $formats = array_keys($provider->getFormats());
+        } elseif (is_string($formats)) {
+            $formats = array($formats);
+        }
+
+        if (!is_array($formats)) {
+            throw new \InvalidArgumentException('"Formats" argument should be string or array');
+        }
+
+        foreach ($formats as $format) {
             $path = $provider->generatePrivateUrl($media, $format);
             if ($path && $provider->getFilesystem()->has($path)) {
                 $provider->getFilesystem()->delete($path);
--- a/Thumbnail/ThumbnailInterface.php
+++ b/Thumbnail/ThumbnailInterface.php
@@ -39,6 +39,7 @@ interface ThumbnailInterface
     /**
      * @param \Sonata\MediaBundle\Provider\MediaProviderInterface $provider
      * @param \Sonata\MediaBundle\Model\MediaInterface            $media
+     * @param string|array                                        $formats
      */
-    public function delete(MediaProviderInterface $provider, MediaInterface $media);
+    public function delete(MediaProviderInterface $provider, MediaInterface $media, $formats = null);
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ A [BC BREAK] means the update will break the project for many reasons:
 * new dependencies
 * class refactoring
 
+### 2015-04-03
+
+* [BC BREAK] Add `formats` param to `removeThumbnails` method of `MediaProviderInterface`
+* [BC BREAK] Add `formats` param to `delete` method of `ThumbnailInterface`
+
 
 ### 2014-09-01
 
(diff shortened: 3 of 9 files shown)


rande:
I think we need to add a remove method in the Thumbnail interface. This
will avoid duplicate code and make sure custom implementation will still
work.

MAXakaWIZARD:
@rande done

MAXakaWIZARD:
@rande Any news when this PR can be merged?

Thumbnail/FormatThumbnail.php
```diff
@@ -84,18 +85,59 @@ public function generate(MediaProviderInterface $provider, MediaInterface $media
...
-                $provider->getFilesystem()->delete($path);
+        $filesystem = $provider->getFilesystem();
+
+        if (is_null($format)) {
+            //delete all thumbnails
+            $format = 'default_whateverformat';
+
+            $relPath = $provider->generatePrivateUrl($media, $format);
+            $relDir = dirname($relPath);
+            $fsRootDir = $this->getFilesystemRootDir($filesystem);
```
rande: You cannot do that. The filesystem can be aws. So you cannot rely on php fs functions

MAXakaWIZARD: You mean `dirname`?

rande: The finder, pathinfo ..

MAXakaWIZARD: So, what should I do? Leave old code and use new code only for `Local` adapter?

rande: I think you just need to keep the current logic, but add an extra check:

```php
    public function delete(MediaProviderInterface $provider, MediaInterface $media, array $formats = null)
    {
        $formats = $formats ?: array_keys($provider->getFormats());

        // delete the differents formats
        foreach ($formats as $format) {
            $path = $provider->generatePrivateUrl($media, $format);
            if ($path && $provider->getFilesystem()->has($path)) {
                $provider->getFilesystem()->delete($path);
            }
        }
    }
```

MAXakaWIZARD: @rande I don't understand, what you mean. The code you posted is original code with no changes.

rande: nop it is not the original, the provided formats are now specified, so you can remove specific formats

MAXakaWIZARD: @rande but in that way we cannot remove all old thumbnails, if we do not remember format names

rande: Of course you can, there is no check on the format validity. The media and the format is enough to generate the private path, and so remove it from the fs driver.

MAXakaWIZARD: I mean if there was some format long time ago, and i do not remember it's name, I cannot remove it, so code should search filesystem by mask.

rande: I think the situation is an edge case and we don't want to support this.

MAXakaWIZARD: @rande OK, I will come again later ;)

MAXakaWIZARD:
@rande Please review

MAXakaWIZARD:
@rande Any news when this PR can be merged?

rande:
Can you add the related documentation about the new command ? also this PR introduce a BC break, can you add it to the CHANGELOG file ?

Thanks

MAXakaWIZARD:
@rande I don't see any BC break here, it's just new param in `remove` method with default value with exactly the same behavior by default.

Thumbnail/ThumbnailInterface.php
```diff
@@ -39,6 +39,7 @@ public function generate(MediaProviderInterface $provider, MediaInterface $media
     /**
      * @param \Sonata\MediaBundle\Provider\MediaProviderInterface $provider
      * @param \Sonata\MediaBundle\Model\MediaInterface            $media
+     * @param string|array                                        $formats
      */
-    public function delete(MediaProviderInterface $provider, MediaInterface $media);
+    public function delete(MediaProviderInterface $provider, MediaInterface $media, $formats = null);
```
rande: You add new arguments in the interface, so it is a BC breaks

MAXakaWIZARD:
@rande Ah, I see, new param in interfaces. OK

MAXakaWIZARD:
@rande Done. Tests fail, but as I see it's not related to this PR.

PR #723, opened 2015-03-05
Removed not callable MediaProviderInterface::transform from test

hanovruslan:
Removed not callable MediaProviderInterface::transform from test

fix for #722 **Bug in testPutMediumBinaryContentAction**

Issue #732, opened 2015-03-20
Critical: blocks log in screen: Duplicate key-sequence


scasei:
An exception has been thrown during the rendering of a template ("[ERROR 1877] Element '{urn:oasis:names:tc:xliff:document:1.2}trans-unit': Duplicate key-sequence ['form.label_category'] in key identity-constraint '{urn:oasis:names:tc:xliff:document:1.2}K_unit_id

SonataMediaBundle.de.xliff
Keys: form.label_category, form.label_provider_reference

mrjingles:
Will be fixed with #738.

PR #742, opened 2015-03-24
fix / #723 duplicate key

scasei:


scasei:
wrong issue: right is #732

rande:
already fixed

PR #749, opened 2015-04-06
Change any use ErrorElement from AdminBundle to CoreBundle

ckmm:


ckmm:
only one use

OskarStark:
this looks wrong to me and should not be merged in 2.3

PR #760, opened 2015-04-30
Updated MediaType

phansys:
Updated ```MediaType``` in order to avoid deprecations introduced in Symfony 2.3 and 2.4

| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

### DIFF
--- a/Form/Type/MediaType.php
+++ b/Form/Type/MediaType.php
@@ -50,7 +50,7 @@ class MediaType extends AbstractType
             'new_on_update' => $options['new_on_update'],
         )));
 
-        $builder->addEventListener(FormEvents::BIND, function (FormEvent $event) {
+        $builder->addEventListener(FormEvents::SUBMIT, function(FormEvent $event) {
             if ($event->getForm()->has('unlink') && $event->getForm()->get('unlink')->getData()) {
                 $event->setData(null);
             }
@@ -77,11 +77,21 @@ class MediaType extends AbstractType
     /**
      * {@inheritdoc}
      *
-     * @deprecated Remove it when bumping requirements to Symfony >=2.7
+     * @deprecated Remove it when bumping requirements to Symfony >=2.7, use MediaType::configureOptions() instead
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
-        $this->configureOptions($resolver);
+        if ($resolver instanceof OptionsResolver) {
+            $this->configureOptions($resolver);
+        } else {
+            $resolver->setDefaults(array(
+                'data_class' => $this->class,
+                'provider' => null,
+                'context' => null,
+                'empty_on_new' => true,
+                'new_on_update' => true,
+            ));
+        }
     }
 
     /**
@@ -90,10 +100,10 @@ class MediaType extends AbstractType
     public function configureOptions(OptionsResolver $resolver)
     {
         $resolver->setDefaults(array(
-            'data_class'    => $this->class,
-            'provider'      => null,
-            'context'       => null,
-            'empty_on_new'  => true,
+            'data_class' => $this->class,
+            'provider' => null,
+            'context' => null,
+            'empty_on_new' => true,
             'new_on_update' => true,
         ));
     }


Form/Type/MediaType.php
```diff
@@ -77,14 +76,34 @@ public function buildView(FormView $view, FormInterface $form, array $options)
 
     /**
      * {@inheritdoc}
+     *
+     * @deprecated To be removed when bumping to Symfony 2.7+, use MediaType::configureOptions() instead
      */
     public function setDefaultOptions(OptionsResolverInterface $resolver)
     {
+        if ($resolver instanceof OptionsResolver) {
```
Soullivaneuh: `OptionsResolver` implements `OptionsResolverInterface`. So this condition make no sense.

See: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/OptionsResolver/OptionsResolver.php#L27

You can directly call `$this->configureOptions($resolver)` instead like it's done on many bundle.

phansys: That isn't the condition being checked. Hint for ```setDefaultOptions()``` is ```OptionsResolverInterface```. So, I could call this method with my own implementation of ```OptionsResolverInterface```. Do you agree?

Soullivaneuh: `setDefaultOptions` or `configureOptions` are called by `ResolvedFormType` so you don't have to care of this.

BTW, the deprecated notice is called on this class: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Form/ResolvedFormType.php#L214-L216

Please follow classic method like is done on many bundle like this one: https://github.com/FriendsOfSymfony/FOSUserBundle/pull/1821

This works perfectly and is even more readable and simpler.

Regards

phansys: I know that, but since it's a public method,  I think that (weird but possible) situation could be prevented.

Soullivaneuh: What should it prevent? I don't understand at all.

In both case (`configureOptions` call or `setDefaults` on `setDefaultOptions`) does **exactly the same thing**.

This approach does not add anything more instead of redundant code AFAIK.

phansys: It prevents a fatal error while let both methods coexist (even with different behaviors if it needed) before bumping Symfony's version.
By insntance, a possible exception could be:
```php
Argument 1 passed to Sonata\MediaBundle\Form\Type\MediaType::configureOptions() must be an instance of Symfony\Component\OptionsResolver\OptionsResolver, MyOwnOptionsResolver given
```

Soullivaneuh: In this case, `MyOwnOptionsResolver` should inherit `OptionsResolver` on your project.

And I don't think this can be apply and this project.

I don't know more. Maybe @stof could give a better review/explanation than me.

cc @rande

phansys: I know that @Soullivaneuh, but this is the issue here: ```Symfony\Component\Form\FormTypeInterface::setDefaultOptions()``` allows you to use any implementation of ```Symfony\Component\OptionsResolver\OptionsResolverInterface```. The constraint is only narrowed from ```configureOptions()```.

stof: Well, reimplementing your own OptionsResolver is not something the Symfony team considers as being supported (which is why the interface is deprecated btw)

Soullivaneuh: Thanks @stof. Well, in that case, there is no question about a `MyOptionResolver`.

@phansys Can you please update your PR about that? Thanks.

Soullivaneuh:
Please remove modification about `OptionResolver` and rename your commit:

```
Fix deprecated FormEvents::SUBMIT usage
```

For options resolver, just see this: https://github.com/nelmio/NelmioApiDocBundle/pull/641/files#r32103862

Thanks.

Form/Type/MediaType.php
```diff
@@ -50,7 +50,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             'new_on_update' => $options['new_on_update'],
         )));
 
-        $builder->addEventListener(FormEvents::BIND, function (FormEvent $event) {
+        $builder->addEventListener(FormEvents::SUBMIT, function(FormEvent $event) {
```
Soullivaneuh: `function (` is correct. A space must be there.

Please run `make cs` to have correct coding style.

Soullivaneuh:
BTW, as a bug fix, this could be done on 2.3 branch.

@phansys could you please do that?

phansys:
Of course @Soullivaneuh, I can close this PR and open other against 2.3.

Soullivaneuh:
@phansys Yes. Closing this one so.

phansys:
#823.

PR #764, opened 2015-05-07
publish job on kernel.finish_request

toooni:
when using rabbitmq, there is a race condition because of the database transaction. In our case the worker was too fast and the database entry didn't exist at the time the consumer was processed.
Our solution was to use the event dispatcher and publish the job just at the end of the request (``kernel.finish_request``). This even works if used with sonata-notification-bundle strategy postpone which uses the kernel.terminate event.

### DIFF
--- a/Resources/config/consumer.xml
+++ b/Resources/config/consumer.xml
@@ -17,6 +17,7 @@
             <argument>sonata.media.thumbnail.format</argument>
             <argument type="service" id="sonata.media.thumbnail.format"/>
             <argument type="service" id="sonata.notification.backend" />
+            <argument type="service" id="event_dispatcher" />
         </service>
     </services>
 </container>
--- a/Thumbnail/ConsumerThumbnail.php
+++ b/Thumbnail/ConsumerThumbnail.php
@@ -14,6 +14,7 @@ namespace Sonata\MediaBundle\Thumbnail;
 use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Sonata\NotificationBundle\Backend\BackendInterface;
+use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
 class ConsumerThumbnail implements ThumbnailInterface
 {
@@ -33,15 +34,22 @@ class ConsumerThumbnail implements ThumbnailInterface
     protected $backend;
 
     /**
-     * @param string             $id
-     * @param ThumbnailInterface $thumbnail
-     * @param BackendInterface   $backend
+     * @var EventDispatcherInterface
      */
-    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend)
+    protected $dispatcher;
+
+    /**
+     * @param string                   $id
+     * @param ThumbnailInterface       $thumbnail
+     * @param BackendInterface         $backend
+     * @param EventDispatcherInterface $dispatcher
+     */
+    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend, EventDispatcherInterface $dispatcher = null)
     {
-        $this->id        = $id;
-        $this->thumbnail = $thumbnail;
-        $this->backend   = $backend;
+        $this->id           = $id;
+        $this->thumbnail    = $thumbnail;
+        $this->backend      = $backend;
+        $this->dispatcher   = $dispatcher;
     }
 
     /**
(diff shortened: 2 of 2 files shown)


Thumbnail/ConsumerThumbnail.php
```diff
@@ -56,14 +61,16 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
      */
     public function generate(MediaProviderInterface $provider, MediaInterface $media)
     {
-        $this->backend->createAndPublish('sonata.media.create_thumbnail', array(
-            'thumbnailId'       => $this->id,
-            'mediaId'           => $media->getId(),
+        $this->dispatcher->addListener('kernel.finish_request', function () use ($media) {
+            $this->backend->createAndPublish('sonata.media.create_thumbnail', array(
```
phansys: You can't use ```$this``` inside an anonymous function with PHP < 5.4

toooni:
@phansys thanks.. i updated the PR

toooni:
@rande this really is an issue

Thumbnail/ConsumerThumbnail.php
```diff
@@ -23,16 +25,19 @@ class ConsumerThumbnail implements ThumbnailInterface
...
 
+    protected $dispatcher;
+
     /**
      * @param string                                              $id
      * @param ThumbnailInterface                                  $thumbnail
      * @param \Sonata\NotificationBundle\Backend\BackendInterface $backend
      */
-    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend)
+    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend, EventDispatcherInterface $dispatcher)
```
Soullivaneuh: Change signature by adding a required argument would be BC break.

Could you please handle it?

toooni: @Soullivaneuh what should I do exactly? just add an entry into CHANGELOG.md?

Soullivaneuh: Nope. Just make the class working both with old and new constructors parameters.

In this case, you can make this argument optional and trigger a deprecation error if it's not specified.

Please refer to Symfony methods.

toooni:
@Soullivaneuh made the EventDispatcher optional and added a deprecation warning

Thumbnail/ConsumerThumbnail.php
```diff
@@ -23,16 +24,20 @@ class ConsumerThumbnail implements ThumbnailInterface
...
-    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend)
+    public function __construct($id, ThumbnailInterface $thumbnail, BackendInterface $backend, EventDispatcherInterface $dispatcher = null)
     {
-        $this->id        = $id;
-        $this->thumbnail = $thumbnail;
-        $this->backend   = $backend;
+        $this->id           = $id;
+        $this->thumbnail    = $thumbnail;
+        $this->backend      = $backend;
+        $this->dispatcher   = $dispatcher;
```
Soullivaneuh: A `trigger_error` should be called if `$dispatcher` is null.

Resources/config/consumer.xml
```diff
@@ -17,6 +17,7 @@
             <argument>sonata.media.thumbnail.format</argument>
             <argument type="service" id="sonata.media.thumbnail.format"/>
             <argument type="service" id="sonata.notification.backend" />
+            <argument type="service" id="event_dispatcher" />
```
Soullivaneuh: This argument should be optional to IMO.

cc @rande.

Soullivaneuh:
Could you please also add some related tests?

Thumbnail/ConsumerThumbnail.php
toooni: @Soullivaneuh it's done, but in the generate() method. I thought that later it is more clear to remove the workaround.

Resources/config/consumer.xml
toooni: @Soullivaneuh do you mean by adding a setEventDispatcher() method?

Soullivaneuh: Not necessary. You can replace it by `<argument />` with noting inside and use a `CompilerPass` to add it.

But I prefer wait for @rande confirmation.

Thumbnail/ConsumerThumbnail.php
```diff
@@ -23,16 +24,20 @@ class ConsumerThumbnail implements ThumbnailInterface
 
     protected $backend;
 
+    protected $dispatcher;
```
Soullivaneuh: Please add a doc comment:

```php
/**
 * @var EventDispatcherInterface
 *
 */
```

Thumbnail/ConsumerThumbnail.php
Soullivaneuh: It would be better to have this trigger on constructor instead.

toooni: @Soullivaneuh I'll wait on that. If @rande thinks that the EventDispatcher argument should be optional, there won't be a need for a deprecation message.

Soullivaneuh: It should one code because this method is public and could be used as an extension.

I'm not sure for service only.

Resources/config/consumer.xml
toooni: @rande rabbitmq support is really not working without this.. can you please give tell me if i should use a `CompilerPass`

toooni: @Soullivaneuh can't you make a decision here?

toooni:
@Soullivaneuh @rande ping

Resources/config/consumer.xml
Soullivaneuh: Not sure. As I said, I prefer to hear @rande opinion.

OskarStark:
@toooni can you please rebase your PR?

toooni:
@OskarStark done. sorry for the delay!

OskarStark:
thank you @toooni 

ping @rande for review

OskarStark:
:+1:

Issue #759, opened 2015-04-29
Can´t add a youtube video


bnd170:
When I try to add a new video from youtube, form returns me two errors:

This value should not be null
This value should not be empty

This is a screenshot of the error: http://i.gyazo.com/776749c3f8a819b8a0a49160ea28323b.png

Im using:

        "sonata-project/admin-bundle": "dev-master",
        "sonata-project/core-bundle": "dev-master",
        "sonata-project/doctrine-orm-admin-bundle": "*",
        "sonata-project/jquery-bundle": "dev-master",
        "sonata-project/formatter-bundle": "dev-master",
        "sonata-project/media-bundle": "dev-master",
        "sonata-project/intl-bundle": "dev-master",
        "sonata-project/datagrid-bundle": "dev-master",
        "sonata-project/user-bundle": "dev-master",
        "sonata-project/classification-bundle": "dev-master",

Guillaume-Rossignol:
For me : 
```
$mediaManager = $this->getMediaManager();
$youtube = $mediaManager->create();
$youtube->setBinaryContent('https://www.youtube.com/watch?v=wwwww');
$youtube->setEnabled(true);
$youtube->setName('presentation');
$mediaManager->save($youtube, 'default', 'sonata.media.provider.youtube');
``` 
And the result : 
```
    [exec]   [Buzz\Exception\RequestException]
     [exec]   file_get_contents(https://i.ytimg.com/vi/EQwX4VodyjM/hqdefault.jpg): failed to open stream: operation failed
     [exec] 
     [exec] 
     [exec] 
     [exec] Exception trace:
     [exec]  () at /home/guillaume/PhpstormProjects/axiona/vendor/kriswallsmith/buzz/lib/Buzz/Client/FileGetContents.php:64
     [exec]  Buzz\Client\FileGetContents->send() at /home/guillaume/PhpstormProjects/axiona/vendor/kriswallsmith/buzz/lib/Buzz/Browser.php:130
     [exec]  Buzz\Browser->send() at /home/guillaume/PhpstormProjects/axiona/vendor/kriswallsmith/buzz/lib/Buzz/Browser.php:82
     [exec]  Buzz\Browser->call() at /home/guillaume/PhpstormProjects/axiona/vendor/kriswallsmith/buzz/lib/Buzz/Browser.php:31
[...]

PR #765, opened 2015-05-12
fix default media connector

amine26tz:
set curl as default buzz connector. see https://github.com/sonata-project/SonataMediaBundle/issues/759


rande:
This does not fix the problem. We need to understand we a file_get_contents does not work anymore.

amine26tz:
my bad, when use curl connector the metadata recuperation is broken :/

PR #766, opened 2015-05-14
fix youtube thubmbnail url witout https

amine26tz:
override the https thubmbnail url rendred by youtube api. 

file_get_contents used by buzz connector don't allow https see https://github.com/kriswallsmith/Buzz/issues/67


amine26tz:
ping @rande

Issue #232, opened 2012-11-21
Download with x-sendfile not working


tiagojsag:
From a fresh install of SonataSandbox, just changed "default" context download mode to "X-Sendfile".

On apache error file I get the following:

No such file or directory: xsendfile: cannot open file: redirect:/default/0001/01/thumb_1_reference.jpeg, referer: http://sandbox/admin/sonata/media/media/1/view?context=default

On the browser (FF/Ubuntu) I get:

OK
The requested URL /app.php was not found on this server.
Apache/2.2.22 (Ubuntu) Server at sandbox Port 80

PR #787, opened 2015-06-13
[Thumbnail] Fix thumbnail URL for reference format

NeuralClone:
Fixes a bug in [Sonata\MediaBundle\Thumbnail\FormatThumbnail](https://github.com/sonata-project/SonataMediaBundle/blob/master/Thumbnail/FormatThumbnail.php) that causes an incorrect file name to be returned when "reference" format is requested. This fix is a follow-up to #232 and #277. #277 fixed the issue in FileProvider but left a similar bug in ImageProvider. Credit should go to @tiagojsag for bringing up the original problem and for suggesting a similar fix.

The source of the problem is in [Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePrivateUrl()](https://github.com/sonata-project/SonataMediaBundle/blob/master/Thumbnail/FormatThumbnail.php#L46), which does not take the format into account when generating a thumbnail URL (see discussion below -- thanks to @phansys for suggesting the move). Because of this, requesting the "reference" format results in an incorrectly named thumbnail. This then causes a `NotFoundHttpException` to be thrown. This bug also prevents an image from being downloaded using X-Sendfile and X-Accel-Redirect as the incorrectly named image cannot be found.

**Note:** The fix I've applied to `Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePrivateUrl()` results in the function being functionally identical to `Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePublicUrl()`. The results of these functions are manipulated differently in ImageProvider but they otherwise behave exactly the same. I implemented it this way to prevent `generatePrivateUrl()` from relying on `generatePublicUrl()`. This can be easily changed and I'm willing to make that change if desired.

(Revamped description and title to match the most recent version of the pull request. Unfortunately, I'm unable to change the branch name due to GitHub restrictions.)

### DIFF
--- a/Tests/Provider/ImageProviderTest.php
+++ b/Tests/Provider/ImageProviderTest.php
@@ -69,6 +69,9 @@ class ImageProviderTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals('default/0011/24', $provider->generatePath($media));
         $this->assertEquals('/uploads/media/default/0011/24/thumb_1023456_big.png', $provider->generatePublicUrl($media, 'big'));
         $this->assertEquals('/uploads/media/default/0011/24/ASDASDAS.png', $provider->generatePublicUrl($media, 'reference'));
+
+        $this->assertEquals('default/0011/24/ASDASDAS.png', $provider->generatePrivateUrl($media, 'reference'));
+        $this->assertEquals('default/0011/24/thumb_1023456_big.png', $provider->generatePrivateUrl($media, 'big'));
     }
 
     public function testHelperProperies()
--- a/Thumbnail/FormatThumbnail.php
+++ b/Thumbnail/FormatThumbnail.php
@@ -45,7 +45,12 @@ class FormatThumbnail implements ThumbnailInterface
      */
     public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterface $media, $format)
     {
-        return sprintf('%s/thumb_%s_%s.%s',
+        if ('reference' === $format) {
+            return $provider->getReferenceImage($media);
+        }
+
+        return sprintf(
+            '%s/thumb_%s_%s.%s',
             $provider->generatePath($media),
             $media->getId(),
             $format,


Provider/ImageProvider.php
```diff
@@ -174,6 +174,10 @@ public function generatePublicUrl(MediaInterface $media, $format)
      */
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        return $this->thumbnail->generatePrivateUrl($this, $media, $format);
+        if ($format == 'reference') {
+            return $this->getReferenceImage($media);
+        } else {
```
Soullivaneuh: The `else` is not necessary here.

NeuralClone: Oops, yes, you're right. I'll update this. :)

Provider/ImageProvider.php
```diff
@@ -174,6 +174,10 @@ public function generatePublicUrl(MediaInterface $media, $format)
      */
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
+        if ($format == 'reference') {
```
phansys: Please use Yoda condition and strict comparison.

phansys:
@NeuralClone, could you add some tests?

Provider/ImageProvider.php
Soullivaneuh: :+1: For strict comparison, not for Yoda. This is not an obligation AFAIK.

phansys: I think Yoda it's a Symfony's convention: https://github.com/FriendsOfPHP/PHP-CS-Fixer/commit/4725fec466e9a796186896334d8c87f3bc46f7ee

NeuralClone: I'm fine with making those changes. However, I do question why a Yoda condition is being asked for with this when the rest of the bundle doesn't seem to use them. It was written this way to maintain consistency with the rest of the bundle.

For example: https://github.com/sonata-project/SonataMediaBundle/blob/master/Provider/FileProvider.php#L364-369

NeuralClone:
Yeah, I can try to add some tests. I'm not sure when I'll have time to get to that, though.

Provider/ImageProvider.php
phansys: @NeuralClone, I asked for Yoda condition because it's a current Symfony's convention as I pointed in response to @Soullivaneuh comment.

Soullivaneuh: > Yoda condition because it's a current Symfony's convention

@phansys I don't think so. Or I missed something. Could you please give your source?

Nothing about Yoda condition here: http://symfony.com/doc/current/contributing/code/standards.html

Soullivaneuh: @phansys Forgot this, didn't see your link.

But, It's not on documentation nor on stable version of the fixer: https://github.com/FriendsOfPHP/PHP-CS-Fixer/find/v1.9

So I think we should ignore this before it's official (or not).

Soullivaneuh: Maybe @rande can give his opinion too.

phansys: Indeed, given the [current documentation of Symfony about conventions](http://symfony.com/doc/current/contributing/code/conventions.html) misses some of these aspects, I'm working now in a PR for that.

NeuralClone: I'm fine with whatever ends up being decided. Ultimately, I just want to see this bug fixed. I'll hold off on updating this further until there's more feedback.

I do appreciate the links and all of the feedback, by the way. I wasn't aware that those changes were being made to the coding conventions.

NeuralClone:
I've updated this pull request to include some additional tests for ImageProvider and I've also updated the other providers to use strict comparisons where appropriate as discussed above. The tests all pass. I didn't switch over to Yoda conditions for the time being but I'll gladly make those changes if that's desired.

phansys:
Ref. for Yoda: symfony/symfony-docs#5402

NeuralClone:
The comments in that PR are good enough for me. :) I've updated this to include Yoda conditions. Since I modified the other providers to include strict comparisons, I also updated them to include Yoda conditions as well. I can easily remove the Yoda condition stuff and put it in a separate pull request if that makes more sense.

More work needs to be done in the rest of the bundle to update it to match the same code standards if we want consistency. Some classes use this style and others do not. That seems to go beyond the scope of this pull request but I'd be willing to volunteer to create another one to work on it.

Soullivaneuh:
@NeuralClone Please update **only** your modification with Yoda comparison.

Since a fixer will be merged on next version of php-cs-fixer, code will be automatically fixed later anyway. :+1:

NeuralClone:
Ok, sounds good! I thought that made more sense. Not sure why I didn't just go with that in the first place. Long weekend I guess? :)

Anyway I'll revert this to just my fix plus the tests I added later today.

NeuralClone:
This is now back to just my fix and the two tests I added. Sorry for the earlier deviation from that!

NeuralClone:
I was just wondering if there was any additional feedback on this pull request. It's been over a month since I originally submitted this and all of the feedback I've received has been about coding styles and standards.

While I agree that that stuff is important, it ultimately isn't relevant to the bug this pull request fixes. Nor is it particularly helpful. X-Sendfile is currently broken in this bundle when dealing with images, which makes the bundle mostly worthless when dealing with large image files. Sending large files through HTTP with PHP is rather slow. It doesn't make sense to use that mode when other, superior modes are available for use.

I'm willing to provide additional tests if more are required to verify the existence of this bug and that this pull request fixes it. But since I haven't received any useful feedback since I submitted this, I have no idea where this stands and what the intentions are regarding this bug. Will this fix eventually be merged? If so, any estimate on when that might be? Or is support for X-Sendfile being removed in a later version of the bundle, which would render this fix pointless?

Thanks!

phansys:
I think this fix doesn't only affects ```X-Sendfile```; moreover, this should be propagated to other methods IMO. By instance, what happen by instance if you call [Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePrivateUrl()](https://github.com/sonata-project/SonataMediaBundle/blob/0c6103251c8e41b0b72f4f885533d8d54a5b3ef2/Thumbnail/FormatThumbnail.php#L46) passing "reference" as 3rd argument?

NeuralClone:
Thanks for the input. :)

As far as I can tell, it impacts both X-Sendfile and X-Accel-Redirect. The `http` mode is always handled directly in [Sonata\MediaBundle\Provider\FileProvider::getDownloadResponse()](https://github.com/sonata-project/SonataMediaBundle/blob/0c6103251c8e41b0b72f4f885533d8d54a5b3ef2/Provider/FileProvider.php#L359). The mode is ignored in all of the providers that extend BaseVideoProvider since a RedirectResponse is used to "download" the media. The title of this pull request should probably be "Fix thumbnail URL for reference format" or something similar.

When it comes to your example, you raise a good point. That method seems to be where the problem starts. If you pass "reference" as the third argument, it will consistently generate an invalid thumbnail name. It makes the assumption that a format other than "reference" will be used, which is rather flawed. I approached my fix in the same way as it was in #277 for consistency. In fact, it's exactly the same fix that @tiagojsag originally proposed (plus some tests). A reason wasn't given for why ImageProvider wasn't also fixed along with FileProvider when pull request was closed.

However, I agree with you. `FormatThumbnail::generatePrivateUrl()` is currently making some bad assumptions and any code that relies on it will be affected by it. I'll give this another look and see if it makes more sense to move the fix to a different method as well as clarify what it is this is fixing.

NeuralClone:
I've played around with this a bit more. Moving the fix to `FormatThumbnail::generatePrivateUrl()` would actually make that function identical to [Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePublicUrl()](https://github.com/sonata-project/SonataMediaBundle/blob/0c6103251c8e41b0b72f4f885533d8d54a5b3ef2/Thumbnail/FormatThumbnail.php#L32). So the fix could be reduced to the following:

```php
// Sonata\MediaBundle\Thumbnail\FormatThumbnail::generatePrivateUrl()

/**
 * {@inheritdoc}
 */
public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterface $media, $format)
{
    return $this->generatePublicUrl($provider, $media, $format);
}
```

It's certainly a more elegant fix that also addresses the core problem with `FormatThumbnail::generatePrivateUrl()` ignoring the format. The tests all pass with this change but it seems like these functions were originally written differently for a reason. Or maybe they weren't. There's a separate class that implements ThumbnailInterface and provides integration with LiipImagineBundle. So it shouldn't impact integration with that bundle (as far as I can tell).

Overall, it *seems* to be ok. Thoughts?

NeuralClone:
I've updated this pull request to simplify the fix and address @phansys's very valid points. The fix itself has been moved into FormatThumbnail::generatePrivateUrl() as that is the source of the problem. Since fixing the issue with the reference format in FormatThumbnail::generatePrivateUrl() results in the exact same function as FormatThumbnail::generatePublicUrl() (as I discussed above), the incorrect URL being generated can be fixed simply by calling that function.

If this is a problem and a different approach is preferred (for whatever reason), I'm more than willing to change it. However, I'm kind of working in the dark when it comes to what is desired. It feels like I'm going in circles with trying to get this fixed. Whenever I make a change, there's always a problem with it and the bug remains unfixed. That kind of moving target is very frustrating. I doubt I'm the only person that needs X-Sendfile and X-Accel-Redirect to work with ImageProvider.

I'm willing to do what it takes to get this bug fixed. It has existed for over 2 years (see #232 and #277) and an identical fix was integrated into FileProvider a year and a half ago (#277). What I'm proposing is very minor and solves a major problem. Just tell me what needs to be done to get this fix (or a similar one) merged and I'll do it. :)

NeuralClone:
Any updates at all on this? @rande, @Soullivaneuh? Anyone? :)

NeuralClone:
@rande 

Nothing? It'd be really nice to get this bug fixed. If my approach to fixing it isn't the desired approach, it would be great to know that. I'll gladly make whatever changes are necessary to get this merged. But without much feedback or any update on the status of this in 3 months, I'm starting to feel that fixing bugs isn't on the list of priorities for this project. That's both baffling and disappointing if that's the case.

I really do not understand why this continues to be ignored. This is a bug that prevents a major feature from working and it has a trivial fix that was suggested over 2 years ago.

Soullivaneuh:
@NeuralClone I'm sorry but I'm not the best to review this PR on this bundle.

I'm sure @rande is not ignoring your PR but have a lot of work to do, just be patient. :+1: 

Just a question: Is this bug exists also on 2.3? If yes, could you reopen a PR on this branch instead of master? Thanks.

NeuralClone:
@Soullivaneuh Thanks for the response! I appreciate it. :) I wasn't quite sure who was best to bring this to the attention of.

And yes, I'm pretty sure this exists in 2.3 (the bug has existed in some form for quite a while) but I'll double check and create a new PR on that branch if appropriate.

Soullivaneuh:
> I'll double check and create a new PR on that branch if appropriate.

Thanks. :+1:

PR #788, opened 2015-06-13
All symfony/symfony package is not needed

Soullivaneuh:


phansys:
I'm :+1:  with this change, but I think maybe it should add at least a "suggest" section for ```doctrine/orm```. It's just an example, I don't know if my assumption is valid or which other packages will be left out with this update.

Soullivaneuh:
@phansys we are already suggesting `"sonata-project/doctrine-orm-admin-bundle": "~2.3"` where doctrine is required.

phansys:
But requirement for ```sonata-project/doctrine-orm-admin-bundle``` is only in ```require-dev``` section AFAIK.

phansys:
I didn't notice the [suggest](https://github.com/sonata-project/SonataMediaBundle/pull/788/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780L47). Forget about this concern then.

phansys:
BTW, it should be considered as a BC break IMO.

Soullivaneuh:
> it should be considered as a BC break IMO.

What thing? This PR?

phansys:
Yes, because setups relying in implicit requirements of ```symfony/symfony``` could be broken after this change.
By other hand, I checked ```symfony/symfony``` and the requirement for ```doctrine/orm``` is declared in ```require-dev```, so not an issue at all (at least with that dependency).

Soullivaneuh:
> Yes, because setups relying in implicit requirements of symfony/symfony could be broken after this change.

Well, I think user that want unused libs of this bundle have added it manually on composer.json file.

Don't think is a real BC break, BTW it's already done on other bundles.

> By other hand, I checked symfony/symfony and the requirement for doctrine/orm is declared in require-dev, so not an issue at all (at least with that dependency).

What do you mean? Don't understand the relation with this subject.

phansys:
The relation is I gave ```doctrine/orm``` as example about my concern in my first comment (thinking that ```symfony/symfony``` had a hard dependency for it, which is not true).

Soullivaneuh:
Yes, but I don't see why it should be suggested here as doctrine-orm-admin is already suggested too.

phansys:
Recap:
I exposed a concern about breaking requirements that was inherited in an implicit way from ```symfony/symfony```. I used ```doctrine/orm``` as an example, which later I figured out that wasn't a requirement in ```symfony/symfony```, so  my example is not valid.
If you doesn't consider this a BC break, just forget about my comments. As I said before, it's a :+1: for me.

Soullivaneuh:
Thanks for the recap, it's better comprehensive now. :-)

Issue #783, opened 2015-06-12
Sonata Classification manager problem


milosmoto:
Hi I have problem with adding sonata media bundle, after I add and configure , there is problem with some service.

Any help on this?

[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]  
  The service "sonata.media.admin.media" has a dependency on a non-existent service "sonata.classification.manager.category".

NeuralClone:
I had a similar issue when I first started using the bundle. If you're using the master branch like I am, it has a dependency on SonataClassificationBundle, which isn't mentioned in the documentation:

https://github.com/sonata-project/SonataMediaBundle/blob/master/composer.json

```json
"require": {
        "symfony/symfony": "~2.3",
        "sonata-project/core-bundle": "~2.2",
        "sonata-project/notification-bundle": "~2.2",
        "sonata-project/easy-extends-bundle": "~2.1",
        "sonata-project/doctrine-extensions": "~1.0",
        "sonata-project/classification-bundle": "~2.3",
```

You should be able to get it to work if you follow the instructions for installing the SonataClassificationBundle:

https://sonata-project.org/bundles/classification/master/doc/reference/installation.html

I'm not sure if the documentation for that bundle is 100% up to date but I don't recall having any problems getting it to work. It's pretty straightforward.

Trip09:
other option is to remove the dependency, since this should be decouple from the classification bundle. I will work on this.

NeuralClone:
Well, yes. That would be ideal. I was simply suggesting a way to get it working as is. :)

x500:
@NeuralClone Thanks!

tolstoyleo:
is there an easier way to remove the classification bundle? use an older version? i'm also facing this issue but classification bundle is not an option for us.

Trip09:
@tolstoyleo is a PR to solve this issue... you will need to wait for https://github.com/sonata-project/SonataMediaBundle/pull/794

Okimbellum:
Try to add
`new Sonata\ClassificationBundle\SonataClassificationBundle(),`
into AppKernel.php 
It works for me.

k-alex:
@Okimbellum Many thanks 

this worked

Try to add
new Sonata\ClassificationBundle\SonataClassificationBundle(),
into AppKernel.php
It works for me.

jordisala1991:
[...]

PR #794, opened 2015-06-19
[ISSUE-783]: Remove dependency of classification-bundle

Trip09:
BC: break

This PR will decouple the media-bundle from classification-bundle.

https://github.com/sonata-project/SonataMediaBundle/issues/783


NeuralClone:
I haven't looked at this in full detail but conceptually :+1: from me.

Soullivaneuh:
> BC: break

No change to keep BC? If not, this PR will be review later for the next major release.

Also tests need to be fixed.

Thanks.

Model/CategoryManager.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
```
phansys: Missing license header.

Model/CategoryManagerInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
```
phansys: Missing license header.

Soullivaneuh:
Also tests are failing.

Please rebase your branch and fix tests if necessary.

Model/CategoryManagerInterface.php
```diff
@@ -0,0 +1,24 @@
...
+     * @param integer $categoryId
+     * @return array
+     */
+    public function find($categoryId);
+
+    /**
+     * @param array $criteria
+     * @return array
+     */
+    public function findBy($criteria);
```
phansys: Should ```$criteria``` be type hinted with ```array```?

Trip09:
Hi @Soullivaneuh, in fact I double check it with the tags and the classification bundle is not required in any tag so There is **NO BC break**.

The thing is in the moment SonataMediaBundle is relying on if Sonata\ClassificationBundle\Model\CategoryInterface exists so Media should have a relation with Category and should use it.

But in fact as developers we should have the option to use it or not.
So I need to know what should be the default behaviour:
1>> If CategoryInterface existe use it by default.
2>> Only use it if the user defined it.

IMO 2nd option should be adopted.

Let me know what you think. And I will fix the PR.

For research purpose classification was introduced only in commit #0dbc19f0 .
https://github.com/sonata-project/SonataMediaBundle/compare/2.3.2...master

Soullivaneuh:
@Trip09 

> in fact I double check it with the tags and the classification bundle is not required in any tag so There is NO BC break.

In this case, we could remove it without BC indeed.

> 1>> If CategoryInterface existe use it by default.
> 2>> Only use it if the user defined it.

I prefer option 1. And classification package should be one `suggest` instead of `require`.

ping @rande for opinion.

rande:
I will prefer to have an explicit dependency to classification bundle, and have an option to disable the integration.

Soullivaneuh:
> I will prefer to have an explicit dependency to classification bundle, and have an option to disable the integration.

Hmm well but why? May bundle with optional deps use class detection. This way will reduce configuration time.

Soullivaneuh:
BTW, any optional used packages should be suggested, not required. Am I wrong?

rande:
@Soullivaneuh class detection is not enough, we need an option to disable the feature. We might have a situation where the Blog Bundle is used and still don't want categorisation on media.

Soullivaneuh:
> we need an option to disable the feature.

This is the subject of my why question. :-)

If we disable this feature via configuration, we will have an unused installed package, isn't it?

Maybe both (configuration + detection/suggest) would be a nicer deal.

rande:
unused installed package is not an issue. there are other unused packages already installed with sonata or symfony's bundles.

Soullivaneuh:
I understand even if we can avoid to add more unused package. ;-)

I'm still curious: Why class detection is not enough for you?

Trip09:
ok, I will fix it according option 2:

* auto detect if classification bundle exist and use it by default
* option to disable the use of classification
* allow user to use custom Classification if implements CategoryManagerInterface && CategoryInterface

also @rande we shouldn't force classification bundle to be installed because:
1. this is optional so we should never force optional stuff to be installed
2. if we use other bundle that detects interface_exists('Sonata\\ClassificationBundle\\Model\\CategoryInterface') and force it to use we are creating a dependency problem without reason.

[as example SonataCommentBundle + SonataMediaBundle with classification bundle in composer required === SonataCommentBundle using classification bundle even if we don't want it...]

I will be working on this PR during the weekend...

Trip09:
@Soullivaneuh 

auto detect if classification bundle exist and use it by default
option to disable the use of classification
allow user to use custom Classification if implements CategoryManagerInterface

BC: true, 
Tests: Pass

Scrutinizer is failing but I don't think is related to the code, anyway I need more information to solve that.

Trip09:
@rande  & @Soullivaneuh :

Q                         | A
---------------------- | ---
Bug fix?               | yes
New feature?      | yes
BC breaks?         | no
Deprecations?    | no
Tests pass?        | yes
Fixed tickets       | 
License               | MIT
Doc PR               | yes

Model/Media.php
```diff
@@ -618,7 +617,7 @@ public function getCategory()
     /**
      * @param mixed $category
      */
-    public function setCategory(CategoryInterface $category = null)
+    public function setCategory($category = null)
```
rande: why are you removing the type hinting ?

composer.json
```diff
@@ -37,13 +36,17 @@
         "sonata-project/formatter-bundle": "~2.2",
         "sonata-project/datagrid-bundle": "~2.2",
         "sonata-project/seo-bundle": "^1.1.5",
+        "amazonwebservices/aws-sdk-for-php": "~1.6",
```
rande: This is not related to the PR

rande:
The objective is to disable classification bundle in the media bundle, not the other way around. So the classification is mandatory

composer.json
Trip09: I need to add this to require-dev since it's required to pass the tests. other way they will fail or being skipped.

rande: OK

Model/Media.php
Trip09: the idea is to decouple the bundle. So that's why is required to remove the "Sonata\ClassificationBundle\Model\CategoryInterface"

Trip09:
I didn't get your last comment, because this is what I did "disable classification bundle in the media bundle". 
Why you want to force a MediaBundle to use a Classification system?? 
Media Bundle should be concern about taking care of Media stuff.
Make a classification in the media should be an extra not a requirement.

Joooseph:
no updates?

OskarStark:
ping @rande i would like to merge this PR

could you please give us a final review?

wadjeroudi:
ping @rande
:+1:

Nightbr:
Hi, when would you merge this PR? All checks have passed and it's a really great feature (I need it ^^).
Thanks in advance!

Resources/doc/reference/advanced_configuration.rst
```diff
@@ -13,6 +13,10 @@ Full configuration options:
             gallery_has_media:  Application\Sonata\MediaBundle\Entity\GalleryHasMedia
             category:           Application\Sonata\ClassificationBundle\Entity\Category
 
+        force_disable_category:   false # true IF you really want to disable the relation with category
+        category_manager:   null # default set sonata.classification.manager.category if sonata-project/classification exists
```
OskarStark: please add these comments as `->info()` to the configuration value, too

OskarStark:
This is a big code change, and there are no tests... :(

NeuralClone:
This is definitely a must have feature as far as I'm concerned. The ClassificationBundle is extremely useful (it's saved me a ton of time) but there are use cases where it may be overkill for what needs to be done or a custom system may need to be used instead. If it isn't needed, it doesn't make sense to make it a required dependency, IMO.

OskarStark:
@NeuralClone , I'm definitly :+1: for this change, but i would like to have a test, that the code will work as expected.

OskarStark:
and a rebase is needed, too

Nightbr:
Actually I use the fork of @Trip09 https://github.com/IgniteYourProject/SonataMediaBundle and it works great. But I would prefere to use the main sonata Media bundle from the official repo. @Trip09 can you rebase your repo and made some tests in order to validate this PR ? Thanks in advance and great job for the sonata community !

NeuralClone:
@OskarStark, oh absolutely. I agree that tests and a rebase are needed. :)

And, yes. Thank you for your work on this, @Trip09!

Trip09:
Hi guys.... I thought this PR was dead since 20.08.2015... anyway... 

1. @OskarStark I add the missing test and Info as requested [please change PR label]
2. @Nightbr I would also like to use SonataMediaBundle... ;)

All checks are green [Travis is failing but not related with this PR]... hope this will be merged this time...

Tests/Command/FixMediaContextCommandTest.php
```diff
@@ -47,20 +45,29 @@ class FixMediaContextCommandTest extends CommandTest
     private $pool;
 
     /**
-     * @var \PHPUnit_Framework_MockObject_MockObject|ContextManagerInterface
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Sonata\ClassificationBundle\Model\ContextManagerInterface
```
core23: Can you please leave the shorthand and add the full classpath to the use section

Tests/Command/FixMediaContextCommandTest.php
```diff
@@ -47,20 +45,29 @@ class FixMediaContextCommandTest extends CommandTest
...
      */
     private $contextManger;
 
     /**
-     * @var \PHPUnit_Framework_MockObject_MockObject|CategoryManagerInterface
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Sonata\ClassificationBundle\Model\CategoryManagerInterface
```
core23: Same here

Tests/Command/FixMediaContextCommandTest.php
NeuralClone: Doesn't adding a use statement like that defeat the purpose of this pull request (i.e., decouple MediaBundle and ClassificationBundle)?

Or am I missing something? My apologies if I am. Looking at this on my phone and it isn't the most convenient way to look at the code.

Trip09: @core23 @NeuralClone is completely right, plus there is no need to include a dependence if is not used in the code. 

Right now with the full qualified name as comment php can ran without error. If you add a dependence of a class that doesn't exist php will give you a error...

Tests/Command/FixMediaContextCommandTest.php
Trip09: Same as before

OskarStark:
Great work @Trip09 :+1:

OskarStark:
we will merge this in the next major

Nightbr:
Hi all, since your last commit @Trip09 I have this error with your fork:
ServiceNotFoundException in CheckExceptionOnInvalidReferenceBehaviorPass.php line 58:
The service "sonata.media.admin.media" has a dependency on a non-existent service "sonata.classification.manager.category".

It seems like your last commit has some issue. I will investigate more but I think there is a call to the service sonata.classification.manager.category from the service sonata.media.admin.medi which it can't resolve.

Nightbr:
Ok I found something:
I had to add in config.yml in sonata_media force_disable_category: true like this:
```
sonata_media:
    force_disable_category: true
    default_context: default
    ...
```

It is due to this:
In SonataMediaExtension.php ll. 259, 452, 467
You should remove the condition on $config['force_disable_category']

mpoiriert:
Is this pull request dead ?

Nightbr:
I hope not.... It should be merged in next major I think

mpoiriert:
Sorry I saw that this should be merge in next major...

I am a bit disapointed since this pull request is not accepted because of a BC  break but the all point of it is to fix the dependency on the classification bundle that I consider was a BC break by itself...

Rmannn:
It would be very interesting to remove this bundle. This creates some unwanted managements over the media bundle. Please do something about it.

greg0ire:
@OskarStark why merge this in the next major if it is BC
@Trip09 is it BC or not ? Can you rebase ? I think the build should be green now.

Trip09:
@greg0ire it's Backward compatible! 

I'm tired of this PR (almost 1year to merge it...). This was ready to merge at least 3 times! Everything was green... So I only rebase it if there is a commitment from a project owner to merge it! Not just because.

Also the BC BREAK that I talk was cause in this commit  https://github.com/sonata-project/SonataMediaBundle/commit/0dbc19f0 , because was added a dependency of something without a major release.

greg0ire:
If @OskarStark is ok to merge this before the next major, I can totally commit to merging this. I can understand your feeling, but rest assured that things will change (and are already changing, just look at [the admin pulse](https://github.com/sonata-project/SonataAdminBundle/pulse) if you want to be convinced)

OskarStark:
if there is no BC then i am fine to merge this!

To handle it more clear we can merge this in the next major and make a notice in the UPGRADE.md to point that out to the user!

wadjeroudi:
@Trip09 Could you rebase please to unblock this situation ?
@OskarStark Will this be merged on next release 3.0 or right now ?
It's really annoying to install something that won't be used.

OskarStark:
if we could stay BC it will be merge in 3.x, otherwise in the next major

OskarStark:
i reviewed it again and it looks BC to me.

@core23 @greg0ire @Soullivaneuh can you please give it a final review?

@Trip09 please rebase now or after the review is done by the other sonata guys, thank you

Command/FixMediaContextCommand.php
```diff
@@ -11,7 +11,6 @@
 
 namespace Sonata\MediaBundle\Command;
 
-use Sonata\ClassificationBundle\Model\ContextInterface;
```
core23: I think we can leave this import and use non FQNs in the code.

Controller/Api/GalleryController.php
```diff
@@ -343,7 +343,7 @@ protected function handleWriteGalleryhasmedia(GalleryInterface $gallery, MediaIn
             'csrf_protection' => false,
         ));
 
-        $form->bind($request);
+        $form->handleRequest($request);
```
core23: Not related to this PR. Please create a separate one.

DependencyInjection/Configuration.php
```diff
@@ -34,6 +34,14 @@ public function getConfigTreeBuilder()
             ->children()
                 ->scalarNode('db_driver')->isRequired()->end()
                 ->scalarNode('default_context')->isRequired()->end()
+                ->scalarNode('category_manager')
+                    ->info('if sonata-project/classification exists will set "sonata.classification.manager.category" | if you want to define your own category manager you need to implement \Sonata\MediaBundle\Model\CategoryManagerInterface and set the service name here')
```
core23: Please add line breaks, so the line doesn't exceed 80 chars.

DependencyInjection/Configuration.php
```diff
@@ -407,7 +415,10 @@ private function addModelSection(ArrayNodeDefinition $node)
                         ->scalarNode('media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Media')->end()
                         ->scalarNode('gallery')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Gallery')->end()
                         ->scalarNode('gallery_has_media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\GalleryHasMedia')->end()
-                        ->scalarNode('category')->defaultValue('Application\\Sonata\\ClassificationBundle\\Entity\\Category')->end()
+                        ->scalarNode('category')
+                            ->info('OR Application\Sonata\ClassificationBundle\Entity\Category if exists')
```
core23: What ``OR ...``? Please change this to a meaningful description.

DependencyInjection/SonataMediaExtension.php
```diff
@@ -252,7 +256,7 @@ public function registerDoctrineMapping(array $config)
             ),
         ));
 
-        if (interface_exists('Sonata\ClassificationBundle\Model\CategoryInterface')) {
+        if (false === $config['force_disable_category'] && null !== $config['class']['category'] && false !== $config['class']['category']) {
```
core23: Why are you checking against `false`?

DependencyInjection/SonataMediaExtension.php
```diff
@@ -435,6 +439,42 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
+     * @param array            $config
+     * @param XmlFileLoader    $loader
+     */
+    public function configureCategoryInMedia(ContainerBuilder $container, array $config, XmlFileLoader $loader)
+    {
+        // Use default sonata-project classification if present and nothing against
+        if (
+            false === $config['force_disable_category']
+            && interface_exists('Sonata\ClassificationBundle\Model\CategoryInterface')
+            && (false === $config['category_manager'] || null === $config['category_manager'])
```
core23: Same here. Why `false`?

DependencyInjection/SonataMediaExtension.php
```diff
@@ -435,6 +439,42 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
+            $container->getDefinition('sonata.media.admin.media')
+                ->replaceArgument(4, $container->getDefinition('sonata.media.manager.category'));
+
+            // Set the default sonata classification class
+            if (null === $config['class']['category'] || false === $config['class']['category']) {
+                $config['class']['category'] = 'Application\\Sonata\\ClassificationBundle\\Entity\\Category';
+            }
+        } elseif (
+            false === $config['force_disable_category']
+            && false !== $config['category_manager']
```
core23: Same here. Why `false`?

Model/Media.php
```diff
@@ -630,7 +629,7 @@ public function getCategory()
     /**
      * @param CategoryInterface $category|null
      */
-    public function setCategory(CategoryInterface $category = null)
+    public function setCategory($category = null)
```
core23: BC break!

Tests/Command/FixMediaContextCommandTest.php
```diff
@@ -11,8 +11,6 @@
 
 namespace Sonata\MediaBundle\Tests\Command;
 
-use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
```
core23: We can keep the imports.

Tests/Mock/CategoryInterface.php
```diff
@@ -0,0 +1,16 @@
...
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\ClassificationBundle\Model;
```
core23: Sure this is the right namespace? The file is located under `Tests/Mock/...`.

core23: Same for the other `Mock` files.

composer.json
```diff
@@ -35,15 +37,19 @@
...
         "symfony/phpunit-bridge": "~2.7|~3.0",
         "friendsofsymfony/rest-bundle": "~1.1",
-        "nelmio/api-doc-bundle": "~2.4"
+        "nelmio/api-doc-bundle": "~2.4",
+        "doctrine/phpcr-odm": "~1.2.4",
+        "fabpot/php-cs-fixer": "~0.1|~1.0"
     },
     "suggest": {
         "sonata-project/doctrine-orm-admin-bundle": "~2.3",
+        "sonata-project/classification-bundle": "~2.3",
```
core23: Should be added to `require-dev` for testing.

Controller/Api/GalleryController.php
Trip09: This is "boy scout rule" yes is not related to PR but is improving the code not damage it!

Command/FixMediaContextCommand.php
Trip09: No because other wise your are couplint this to ClassificationBundle

DependencyInjection/SonataMediaExtension.php
Trip09: allowing to define it as false|null

DependencyInjection/SonataMediaExtension.php
Trip09: same as before

Controller/Api/GalleryController.php
core23: That's the reason for a new PR. If we rollback this PR, the *improving the code* will also be reverted too.

DependencyInjection/SonataMediaExtension.php
core23: Can you show me the line, where we could set `$config['class']['category']` to false?

Model/Media.php
Trip09: It's NOT a BC Break! The BC break was caused in https://github.com/sonata-project/SonataMediaBundle/commit/0dbc19f0 . This is removing the couple of SonataMedia to SonataClassification... This accept a category and not *only* the CategoryInterface from categoryclassification...

Tests/Command/FixMediaContextCommandTest.php
Trip09: No if you really want to be decouple from other bundles...

Tests/Mock/CategoryInterface.php
Trip09: yes, this is a test to Mock if the file CategoryInterface from ClassificationBundle exists

composer.json
Trip09: there is a test mocking the object. classification-bundle is optional.

Trip09:
@OskarStark  @greg0ire @wadjeroudi , I can work on this bundle during the next days. But I will not add the coupling back into the bundle [branch] as been pointed by @core23 comments. 

So Tell me what is your idea about this PR.

Model/Media.php
core23: This commit is 2 years old and there are serveral tagged version between this and now, so it is a break.

https://3v4l.org/OvDJP

DependencyInjection/SonataMediaExtension.php
```diff
@@ -435,6 +439,42 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
+     * @param ContainerBuilder $container
+     * @param array            $config
+     * @param XmlFileLoader    $loader
+     */
+    public function configureCategoryInMedia(ContainerBuilder $container, array $config, XmlFileLoader $loader)
+    {
+        // Use default sonata-project classification if present and nothing against
+        if (
+            false === $config['force_disable_category']
+            && interface_exists('Sonata\ClassificationBundle\Model\CategoryInterface')
```
core23: This code will only be executed if the `classifcation-bundle` is loaded, so we must `require-dev` the `classifcation-bundle` to test your code.

Trip09: Nope, this is relying on "if Class exists" so if you have the class in composer bootstrap this will be true.

Model/Media.php
Trip09: if I add CategoryInterface I will be doing a BC since it's removing it's not adding a BC this is the difference.

NeuralClone:
I understand the need for code consistency and discussion when adding features, fixing bugs, etc., but this is really getting ridiculous. This pull request was submitted almost a year ago. It is not rewriting the way the bundle works or doing anything all that advanced. It disables another bundle that is unneeded. That is all it does. Can we please get this merged in the near future? It's clearly a highly sought after change and it doesn't hurt those that still want to use ClassificationBundle. I don't understand the resistance to these changes at all.

Code style changes can also be fixed with PHP-CS-Fixer and it doesn't make sense to delay this pull request even more by asking for changes that are unrelated to it that can be automatically fixed in a later, minor commit. It makes even less sense to ask to revert the changes central to this pull request.

@core23 I really do not understand the repeated requests to recouple this bundle with ClassificationBundle in a pull request intended to **decouple** it. The whole point of this pull request is to remove a BC break added in a previous commit (i.e., forcing people to use ClassificationBundle even if they don't need it) and give people the option to disable a bundle that is not needed for basic MediaBundle functionality. If you keep use statements intact for bundles that have been intentionally disabled, PHP will throw an exception and the code _will not work_. It also undoes the main thing this pull request accomplishes.

Removing ClassificationBundle as a required dependency adds flexibility for developers and removes unnecessary overhead/code for those that do not need ClassificationBundle for their project. Conceptually it also makes sense to be able to disable optional bundles that are not required for MediaBundle to work. ClassificationBundle is a very useful bundle and adds some great functionality (it's saved me tons of time in multiple projects) but not everyone needs it or wants it to be used. :)

mpoiriert:
Is this the reason why the bundle wasn't upgraded to 3.0 ?

Since the 2.4 have disappear a lot of thing broke in the SonataAdmin and I cannot upgrade to 3.0 since the media bundle is not compatible (core 2.3)...

Controller/Api/GalleryController.php
greg0ire: @core23 : a separate commit will do.

Model/Media.php
greg0ire: @Trip09 : it is a BC-break : a class that extends this class and overrides setCategory will have to change its signature if you remove the type hinting.

Tests/Command/FixMediaContextCommandTest.php
```diff
@@ -47,20 +45,29 @@ class FixMediaContextCommandTest extends CommandTest
...
-     * @var \PHPUnit_Framework_MockObject_MockObject|CategoryManagerInterface
+     * @var \PHPUnit_Framework_MockObject_MockObject|\Sonata\ClassificationBundle\Model\CategoryManagerInterface
      */
     private $categoryManger;
 
     /**
      * {@inheritdoc}
      */
-    protected function setUp()
+    public function setUp()
```
greg0ire: Why would you need to change that?

Soullivaneuh:

Regards.

OskarStark:
This is a valid major PR

core23:
Can you please rebase your PR @Trip09

debach:
Is there even a consensus among the Sonata team members about decoupling MediaBundle and ClassificationBundle? Reading through this discussion, I’m not so sure about it.

Seeing that @Trip09 has put a lot of effort in this long-running PR already, the most respectful thing would be to first decide whether or not the PR will be merged (putting code style and such aside), and when.

greg0ire:
@sonata-project/contributors please vote :+1: or :-1: for decoupling both bundles.

core23:
Closing this in favor of #1086

PR #537, opened 2014-04-28
Fixed andFlush parameter in MediaManager never being used.

Uriziel:

PR #796, opened 2015-06-22
Fixed andFlush parameter in MediaManager never being used. Based on #537

prashantpalikhe:


Soullivaneuh:
Having unit test that reveal the concerned would be useful. Can you please consider it?

prashantpalikhe:
Don't have much experience there. May be somebody else could pitch in?

Soullivaneuh:
Maybe @Uriziel?

Uriziel:
Give me an hour.

Soullivaneuh:
> Give me an hour.

You have all the time you want @Uriziel! ;-)

Uriziel:
@Soullivaneuh #797

PR #797, opened 2015-06-23
Fixed andFlush parameter in MediaManager never being used

Uriziel:
Rebased changes of  PR #537 
Added tests compared to PR #796

### DIFF
--- a/Document/MediaManager.php
+++ b/Document/MediaManager.php
@@ -30,7 +30,7 @@ class MediaManager extends BaseDocumentManager
             $entity->setProviderName(func_get_arg(2));
         }
 
-        if ($andFlush && is_bool($andFlush)) {
+        if (is_bool($andFlush)) {
             parent::save($entity, $andFlush);
         } else {
             // BC compatibility with previous signature
--- a/Entity/MediaManager.php
+++ b/Entity/MediaManager.php
@@ -37,7 +37,7 @@ class MediaManager extends BaseEntityManager implements MediaManagerInterface
             $media->setProviderName(func_get_arg(2));
         }
 
-        if ($andFlush && is_bool($andFlush)) {
+        if (is_bool($andFlush)) {
             parent::save($media, $andFlush);
         } else {
             // BC compatibility with previous signature
--- a/PHPCR/MediaManager.php
+++ b/PHPCR/MediaManager.php
@@ -30,7 +30,7 @@ class MediaManager extends BasePHPCRManager
             $entity->setProviderName(func_get_arg(2));
         }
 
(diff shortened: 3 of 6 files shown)


Soullivaneuh:
You could add a PR on @prashantpalikhe repo to complete it instead! ;-)

Anyway, will take a look later. Thanks.

Uriziel:
So?

Document/MediaManager.php
```diff
@@ -34,7 +34,7 @@ public function save($entity, $andFlush = true)
             $entity->setProviderName(func_get_arg(2));
         }
 
-        if ($andFlush && is_bool($andFlush)) {
+        if (is_bool($andFlush)) {
```
Soullivaneuh: Just asking: With this condition, `false` and `true` would pass as `$andFlush` is a boolean.

Would be more logic to have:

```php
is_bool($andFlush) && $andFlush
```

Ore more simple:

```php
true === $andFlush
```

Strict comparison will check both type and value.

What do you think?

Uriziel: The whole point of this commit is to get rid of comparing this to true as you might want to pass false value to the save method.

Soullivaneuh: I don't understand.

Your condition just check if it's a boolean, never-mind if it's `true` or `false`.

I don't think this is the wanted behavior.

Ping @rande

Uriziel: How about you checkout this branch, run tests, checkout HEAD^, run tests again.
I've separated test and code commits so you can easily find out how much it actually helps.
Also read test code to understand why one might want to pass false there

dvapelnik: I want convert one thousand articles with images. But I don't wont flush changes after each articles with media object. I want create media-object, save it w/o flushing. So I'm writing:
```php
// ... creating new media with image
$mediaManager->save($media, false);
$em->persist(); // may by it's excessive
// ... again
// ... again
// ... again
$em->flush();
```
But my `false` flag was ignored and all entities flushed after each saved media with article, authors, author pseudonyms etc. 
I think that this way is incorrect. The developer should be able to choose and this choice should be transparent

OskarStark:
ping @rande

OskarStark:
@Uriziel any news on this PR?

Uriziel:
@OskarStark it was all green when I made it :neutral_face: 

Fun stuff is that it probably just needs rebasing, as new test engines were introduced later on.
Given that this change is already waiting since early 2014 I don't seem to care anymore, if you decide to merge this I can rebase it so new test engines pass.

Soullivaneuh:

Regards.

PR #805, opened 2015-07-01
[MediaType] Made form options "provider" and "context" required

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets |
| License       | MIT
| Doc PR        |

Given the following example:
```php
$builder->add('media', 'sonata_media_type');
```
**Before:**
```
unable to retrieve the provider named : ``
```
**After:**
```
The required option "provider" is missing.
```

### DIFF
--- a/Form/Type/MediaType.php
+++ b/Form/Type/MediaType.php
@@ -89,13 +89,18 @@ class MediaType extends AbstractType
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults(array(
-            'data_class'    => $this->class,
-            'provider'      => null,
-            'context'       => null,
-            'empty_on_new'  => true,
-            'new_on_update' => true,
-        ));
+        $resolver
+            ->setDefaults(array(
+                'data_class'    => $this->class,
+                'empty_on_new'  => true,
+                'new_on_update' => true,
+            ))
+            ->setRequired(array('provider', 'context'))
+            ->setAllowedTypes(array(
+                'provider' => 'string',
+                'context' => 'string',
+            ))
+        ;
     }
 
     /**


Soullivaneuh:
This is BC break as this option was not required before, isn't it?

phansys:
@Soullivaneuh, they was required AFAIK: check [this doc entry](https://sonata-project.org/bundles/media/2-0/doc/reference/form.html).

Soullivaneuh:
Indeed, good catch. :+1: 

Some things to do:

* Fix CS: https://travis-ci.org/sonata-project/SonataMediaBundle/jobs/69181200#L537
* Make `setAllowedTypes` compatible with new SF `>=2.6` signature: https://github.com/symfony/symfony/blob/f59a0baa95ec608bca88a7d6a6e6148e8f298934/src/Symfony/Component/OptionsResolver/OptionsResolver.php#L592
* Could you please take profit to add some tests on this class?

Thanks.

Soullivaneuh:
After reflexion, I think this PR should be targeted on `2.3` branch, as a fix.

Could you do that?

phansys:
Sure @Soullivaneuh, indeed I first made this changes compatible just for Symfony 2.6+, but as I consider this an option that should work with earlier versions of this bundle, I decided to move that part to keep BC.
I'd like to know which branching strategy you're using about new features/bugfixes, etc; because I though it was following the Symfony way; but after [your comment few days ago](https://github.com/sonata-project/SonataUserBundle/pull/552#issuecomment-110800225) in other bundle of this project, I think it isn't true.
Thank you in advance.

phansys:
BTW, I can't change the target branch, so I have to closing this after make another PR against 2.3.
By other hand, about CS rules:
IMO, it should be better to follow all rules inherited from Symfony trying to not disable any of them (such as ```unalign_double_arrow``` and ```unalign_equals```). Adding new ones (like ```ordered_use```) without contradicting previous is fine to me.
At last, I want to mention that IMO the rule ```newline_after_open_tag``` is already covered by ```blankline_after_open_tag```, which is part of Symfony level.

Soullivaneuh:
> I'd like to know which branching strategy you're using about new features/bugfixes, etc; because I though it was following the Symfony way; but after your comment few days ago in other bundle of this project, I think it isn't true.

We failed about that and branch are diverged on some project, please see this WIP: https://github.com/sonata-project/SonataAdminBundle/issues/3053

For this one, branch are not diverged so we can continue to maintain this one. :+1: 

> IMO, it should be better to follow all rules inherited from Symfony trying to not disable any of them

This was already discussed with @rande and we have custom rules we want to keep on sonata. See: https://github.com/sonata-project/SonataAdminBundle/pull/3038

Soullivaneuh:
> indeed I first made this changes compatible just for Symfony 2.6+, but as I consider this an option that should work with earlier versions of this bundle, I decided to move that part to keep BC.

This is not the best practices because Symfony >=2.6 will throw deprecation errors and this bundle will not be compatible with `3.0`.

Use `method_exists` system to switch between old and new implementation.

Soullivaneuh:
I close this one for #806. Branch will be merged.

phansys:
:+1:

PR #807, opened 2015-07-02
Fix namespace in Media.phpcr.xml.skeleton

kea:
The namespace is ...\Document\... instead of ..\PHPCR\...


OskarStark:
good catch :+1: 

@rande please merge

OskarStark:
Thank you @kea

PR #809, opened 2015-07-05
Improved exception message for invalid provider name

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| License       | MIT

Example:
```
Unable to retrieve the provider named "some_provider". Available providers are "provider_a", "provider_b", "provider_c"
```

Maybe the change of the exception type could be considered as a small BC break: previously it was ```RuntimeException``` and now it's changed for ```InvalidArgumentException``` (since it's more accurate for this case).

### DIFF
--- a/Provider/Pool.php
+++ b/Provider/Pool.php
@@ -45,8 +45,14 @@ class Pool
      */
     public function getProvider($name)
     {
+        if (!$name) {
+            throw new \InvalidArgumentException('Provider name cannot be empty, did you forget to call setProviderName() in your Media object?');
+        }
+        if (empty($this->providers)) {
+            throw new \RuntimeException(sprintf('Unable to retrieve provider named "%s" since there are no providers configured yet.', $name));
+        }
         if (!isset($this->providers[$name])) {
-            throw new \RuntimeException(sprintf('unable to retrieve the provider named : `%s`', $name));
+            throw new \InvalidArgumentException(sprintf('Unable to retrieve the provider named "%s". Available providers are %s.', $name, '"'.implode('", "', $this->getProviderList()).'"'));
         }
 
         return $this->providers[$name];
--- a/Tests/Provider/PoolTest.php
+++ b/Tests/Provider/PoolTest.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+ * This file is part of the Sonata package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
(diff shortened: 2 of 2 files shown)


rande:
@phansys tests are failing

phansys:
Yes @rande, I'm checking how the "unkown" provider is intended to work.

phansys:
My bad, fixed.

Provider/Pool.php
```diff
@@ -45,8 +45,14 @@ public function __construct($context)
...
     {
+        if (!$name) {
+            throw new \InvalidArgumentException('Provider name cannot be empty, did you forget to call setProviderName() in your Media object?');
+        }
+        if (empty($this->providers)) {
+            throw new \RuntimeException('There are no providers configured yet');
+        }
         if (!isset($this->providers[$name])) {
-            throw new \RuntimeException(sprintf('unable to retrieve the provider named : `%s`', $name));
+            throw new \InvalidArgumentException(sprintf('Unable to retrieve the provider named "%s". Available providers are %s.', $name, '"'.implode('", "', $this->getProviderList()).'"'));
```
Soullivaneuh: Please set `Available providers are "%s"` on sprintf and remove quotes around the `implode` function.

Soullivaneuh:
@phansys Could you please add some test for this new Exception message? With expected message.

Provider/Pool.php
phansys: @Soullivaneuh, these quotes are there to get a list in this way: 
```"provider_a", "provider_b", "provider_c"```

Comment (Soullivaneuh) [inline Provider/Pool.php]: Yes. This is exactly why I said to put `"` on sprintf. Something like this:

```php
throw new \InvalidArgumentException(sprintf('Unable to retrieve the provider named "%s". Available providers are "%s".', $name, implode('", "', $this->getProviderList())));
```

Much proper IMO.

phansys:
Tests added.

Provider/Pool.php
phansys: I think the current syntax is better since these quotes you're suggesting to put arround ```%s``` are actually the opening and the closing quotes for the first and last array element in the list respectively.

Soullivaneuh:
Thank you @phansys

phansys:
Thank you too @Soullivaneuh.

PR #811, opened 2015-07-11
Update amazon_s3.rst

finchen:
latest aws-sdk-php are not working


Soullivaneuh:
No need to specify the version. Composer will get the latest stable itself by following semantic versions.

BTW, should be `~2.8` and not `2.8.*`.

Issue #818, opened 2015-07-22
Check whether the model classes exists


pyrech:
Hi,

In several sonata bundles (at least SonataMedia and SonataClassification), the config need to specify model classes with the `class` parameter.

For example (in config.yml):
```yaml

sonata_media:
    # if you don't use default namespace configuration
    class:
        media:             AppBundle\Entity\Media\Media
        gallery:           AppBundle\Entity\Media\Gallery
        gallery_has_media: AppBundle\Entity\Media\GalleryHasMedia
        category:          AppBundle\Entity\Classification\Category
```

However, if a specified class doesn't exist (f.e. typo or wrong copy/paste), some wtf problems can happen anywhere in the application (because some metadata are wrong and then some relations / tables are not created) without obvious reasons.

I propose to add a check on `DependencyInjection\Configuration` to ensure the specified classes exist. Do you agree with this? If yes I could eventually make a PR for all the bundles using the `class` configuration

sfavot:
:+1:

Soullivaneuh:
Looks like a good idea. Could you propose a PR about that?

pyrech:
Yep, I will have a look at it this evening ;)

Soullivaneuh:
Also could you please paste some sample errors when specified class does not exist?

With it, I could evaluate the wtf level of these. ;-)

pyrech:
Sure :) Using sonata media 0c6103251c8e41b0b72f4f885533d8d54a5b3ef2, I found these errors:

When clicking on the btn_add button from an entity edit form
> Impossible to invoke a method ("id") on a null variable in "SonataDoctrineORMAdminBundle:Form:form_admin_fields.html.twig" at line 59

When clicking on the btn_list from an entity edit form or trying to display the admin media listing
> Notice: Undefined index: category in /path/vendor/sonata-project/doctrine-orm-admin-bundle/Guesser/FilterTypeGuesser.php on line 64

Soullivaneuh:
And those errors are thrown because specified class simply not exists?

Indeed, this is not very clear. :-)

pyrech:
Yes :-)

When trying to find the problem, I noticed that the doctrine metadata name use className of Base{Model} (not even my wrong class).

So in `Sonata\EasyExtendsBundle\Mapper\DoctrineORMMapper`, all these checks obviously fail:
```php
if (!array_key_exists($metadata->name, $this->associations)) {
    return;
}
```

[...]

PR #820, opened 2015-07-22
Ensure model classes exists in configuration

pyrech:
Ref #818

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -388,9 +388,33 @@ class Configuration implements ConfigurationInterface
                 ->arrayNode('class')
                     ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Media')->end()
-                        ->scalarNode('gallery')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Gallery')->end()
-                        ->scalarNode('gallery_has_media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\GalleryHasMedia')->end()
+                        ->scalarNode('media')
+                            ->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Media')
+                            ->validate()
+                                ->ifTrue(function ($v) {
+                                    return !(class_exists($v) && in_array('Sonata\MediaBundle\Model\MediaInterface', class_implements($v)));
+                                })
+                                ->thenInvalid('Invalid model class (not found or not implementing \Sonata\MediaBundle\Model\MediaInterface) - %s')
+                            ->end()
+                        ->end()
+                        ->scalarNode('gallery')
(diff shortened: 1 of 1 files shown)


DependencyInjection/Configuration.php
```diff
@@ -388,9 +388,33 @@ private function addModelSection(ArrayNodeDefinition $node)
                 ->arrayNode('class')
                     ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Media')->end()
-                        ->scalarNode('gallery')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Gallery')->end()
-                        ->scalarNode('gallery_has_media')->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\GalleryHasMedia')->end()
+                        ->scalarNode('media')
+                            ->defaultValue('Application\\Sonata\\MediaBundle\\Entity\\Media')
+                            ->validate()
+                                ->ifTrue(function ($v) {
```
Soullivaneuh: Why not just `ifFalse` instead of a reversed boolean result?

Soullivaneuh:
Not sure if configuration is the proper place for testing those cases.

Would not be better on the configuration processing?

Also, could you add tests about it?

Feel free to use [matthiasnoback/symfony-dependency-injection-test](https://packagist.org/packages/matthiasnoback/symfony-dependency-injection-test) package if this can help in your case.

DependencyInjection/Configuration.php
pyrech: Are you sure the `ifFalse()` method exists? Can't see it in [ExprBuilder](https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Config/Definition/Builder/ExprBuilder.php).

Soullivaneuh: Indeed. Making sense to me having a `ifFalse`.

Ignore my comment so. :+1:

pyrech:
IMO, the validation should be done inside the Configuration class, not the extension (which is more related to merge and build a clean config).

I will have a look to add tests for this ;)

Soullivaneuh:
Ok thanks, waiting for it.

@rande what do you think of this PR?

rande:
:-1: class might not exist yet as they need to be generated by easy extends.

Soullivaneuh:
@rande What do you mean?

In this case, maybe the check could be done elsewhere in order to get a proper error message.

rande:
The issue is not specific to the media bundle, but for all sonata's bundle and other bundles too. If the user fails to create the correct class nothing will work. I think we can check if doctrine generates a proper error message for missing class.

rande:
closing for now. See related issue.

pyrech:
If the user generates the bundle with easy extends, he doesn't need to specify the `class` parameter. Then, unless I'm wrong, if no class parameter is added to the conf, the default value will be given and the validation will not be run thus no error for the user, is that right?

rande:
easy extends generated only the missing entities required to make the bundle works and so you can extends the model if you want. Now your code will throw an exception because the class will not exist.

PR #831, opened 2015-08-20
Correctly handle $andFlush parameter without breaking BC

julien-maitan:
We only need to check for the type of $andFlush because it will be set to true if not provided.
If we use the condition ($andFlush && is_bool($andFlush)) a call to save($media, false) will never match and will fallback to save($media, true)


Soullivaneuh:
Seems you did something wrong with your PR.

julien-maitan:
You're right ! Sorry, I'm gonna create a correct one.

Issue #747, opened 2015-04-02
API postProviderMediumAction validation errors


ruscon:
```
"sonata-project/media-bundle": "~2.3",
"nelmio/api-doc-bundle": "dev-master",
"jms/serializer-bundle": "@dev",
```

![json respnse](https://api.monosnap.com/rpc/file/download?id=6YOPWlkH1TqQWASqHQeucwFGWJTD6l)

![form profile](https://api.monosnap.com/rpc/file/download?id=AJlPfkkfaCVgkMxDUoapiJgVKtt3MO)

ruscon:
ping ?

ruscon:
ping ?

ruscon:
ping...

bodnar1212:
Hello! Help me please. How you resolve this issues??

PR #832, opened 2015-08-20
Correctly handle $andFlush parameter

julien-maitan:
We only need to check for the type of ``$andFlush`` because it will be set to ``true`` if not provided.

If we use the condition ``($andFlush && is_bool($and_flush))`` a call to ``save($media, false)`` will never match and will fallback to ``save($media, true)``


Uriziel:
Duplicate of #797

OskarStark:
I close this one, as the PR #747 of @Uriziel has tests

Thank you for your contribution @julien-maitan 

Oskar

PR #841, opened 2015-09-02
now it is possible to extend the simple resizer

OskarStark:
and overwrite just this method


OskarStark:
ping @rande @Soullivaneuh

OskarStark:
thank you

Issue #838, opened 2015-08-28
some finetuning for the media widget


OskarStark:
shall i provide a PR?

look at "Typ" and "Kontext"

old:
![bildschirmfoto 2015-08-28 um 14 54 05](https://cloud.githubusercontent.com/assets/995707/9546630/b2706702-4d94-11e5-9b0e-7591f0910118.png)

new:
![bildschirmfoto 2015-08-28 um 14 54 07](https://cloud.githubusercontent.com/assets/995707/9546633/b60347fe-4d94-11e5-86ab-fbd04fceaaed.png)

ping @Soullivaneuh @rande

rande:
look better for me, just the delete action with look weird

OskarStark:
yeah i'm working on it and the delete checkbox is not translated by default, but lets do this in another PR

one more idea:
![bildschirmfoto 2015-08-28 um 15 39 31](https://cloud.githubusercontent.com/assets/995707/9547477/0351f388-4d9b-11e5-810a-82f505cc6a75.png)

@rande

OskarStark:
@rande i'm wondering why the binary content is required all the time...

it's also required, when there is an linked binary already.
any idea?

i fixxed this by implementing this method in the IMageProvider:
```php
    /**
     * {@inheritdoc}
     */
    public function buildMediaType(FormBuilder $formBuilder)
    {
        $formBuilder->add('binaryContent', 'file', array(
            'required' => false
        ));

        if ($formBuilder->getOption('context') == 'api') {
            $formBuilder->add('contentType');
        }
    }
```

and i added the required option to this field.

if you say this is correct, i can provide a PR

stiffroy:
@OskarStark 
Thanks for pointing out the solution with 'required' false. I was playing around a bit with it. And my conclusion is, you don't need to touch the media bundle. You can just update your form builder in the admin code something like
```php
->add('o2oRelation', 'sonata_media_type', array(
     'provider' => 'sonata.media.provider.image',
     'context'  => 'default',
     'required' => false, // using it here
))
```
What do you think?

OskarStark:
no @stiffroy in my use case the media IS required!

But when i edit my entity, and it already has a media, then the form to upload a new one IS NOT required!!

if i use your suggestion, then nothing is required... :(

stiffroy:
@OskarStark 
[...]

PR #848, opened 2015-09-18
the binary content is not required in the media widget

OskarStark:
Refs #838

you can finde more information about this PR here: 
https://github.com/sonata-project/SonataMediaBundle/issues/838#issuecomment-137134101

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -117,10 +117,13 @@ class FileProvider extends BaseProvider
      */
     public function buildMediaType(FormBuilder $formBuilder)
     {
-        $formBuilder->add('binaryContent', 'file');
-
         if ($formBuilder->getOption('context') == 'api') {
+            $formBuilder->add('binaryContent', 'file');
             $formBuilder->add('contentType');
+        } else {
+            $formBuilder->add('binaryContent', 'file', array(
+                'required' => false,
+            ));
         }
     }
 


rande:
@OskarStark there is an issue with CS: 

```
Legend: ?-unknown, I-invalid file syntax, file ignored, .-no changes, F-fixed, E-error
   1) Provider/FileProvider.php (multiline_array_trailing_comma)
```

OskarStark:
done @rande 

but i'm not sure if this is correct, is this method used to create a new media file in den media overview?

because there it has to be mandatory...

rande:
Well, I think you can rely on the context information from the form ``$formBuilder->getOption('context') == 'api'`` to fine tune when the binaryContent is mandatory or not. At least with the API, it is required.

OskarStark:
what does api in this case mean?

would you introduce a widget context?

rande:
The api context is used with the API controller. maybe this can be applied to other use case, but I am not a big fan of such solution. Maybe we need to review the form documentation for an alternative solution.

OskarStark:
yes or we can profide a seperate Form Builder like ```buildWidgetForm```

OskarStark:
my fix worked, for creating a new one we use another formbuilder

new:
bildschirmfoto 2015-09-18 um 21 23 27

widget:
bildschirmfoto 2015-09-18 um 21 24 52

OskarStark:
@rande please merge

OskarStark:
@rande can we please merge this?

Provider/FileProvider.php
```diff
@@ -117,7 +117,9 @@ public function buildCreateForm(FormMapper $formMapper)
      */
     public function buildMediaType(FormBuilder $formBuilder)
     {
-        $formBuilder->add('binaryContent', 'file');
+        $formBuilder->add('binaryContent', 'file', array(
+            'required' => false,
```
rande: It is mandatory for the api context, so maybe just do ``$formBuilder->getOption('context') == 'api'`` for now

OskarStark: done, what do you think?

OskarStark:
ping @Soullivaneuh @rande

PR #850, opened 2015-09-18
updated translation file for media widget

OskarStark:
Refs #849

### DIFF
--- a/Resources/translations/SonataMediaBundle.bg.xliff
+++ b/Resources/translations/SonataMediaBundle.bg.xliff
@@ -405,6 +405,18 @@
                 <source>no_linked_media</source>
                 <target>Няма медиа към този обект</target>
             </trans-unit>
+            <trans-unit id="widget_label_type">
+                <source>widget_label_type</source>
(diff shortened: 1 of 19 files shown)


rande:
Can you add the missing keys for other translations files ?

OskarStark:
yes, what is the translation then?

the key itself?

rande:
yes, the key itself

OskarStark:
done @rande 

btw. i would like to change ```no_linked_media``` to ```widget.no_linked_media``` 

is that allowed?

rande:
it is important to keep translation keep concistency. So just compare with the rest of the file. the dot notation is used by the admin translation key generation.

OskarStark:
ok, shall i rename to ```widget_context```instead of ```widget.labeLcontext```?

rande:
@OskarStark yes it will be better as the widget is not *only* used for the Admin

OskarStark:
i'll do

OskarStark:
done @rande

OskarStark:
Thank you @rande

PR #859, opened 2015-09-30
Proper config for gallery manager

renooo:
Change config for the service **sonata.media.manager.gallery** according to the constructor here :
https://github.com/sonata-project/SonataCoreBundle/blob/master/Model/BaseManager.php#L40

### DIFF
--- a/Resources/config/doctrine_mongodb.xml
+++ b/Resources/config/doctrine_mongodb.xml
@@ -18,7 +18,7 @@
         </service>
 
         <service id="sonata.media.manager.gallery" class="%sonata.media.manager.gallery.class%">
-            <argument>%sonata.media.media.class%</argument>
+            <argument>%sonata.media.gallery.class%</argument>
             <argument type="service" id="doctrine_mongodb" />
         </service>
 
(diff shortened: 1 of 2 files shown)


Soullivaneuh:
It would be BC break, wouldn't be?

renooo:
I'm not sure, @Soullivaneuh

If you look the config for **doctrine_orm** (https://github.com/sonata-project/SonataMediaBundle/blob/master/Resources/config/doctrine_orm.xml#L19), it works fine.

But this constructor 
https://github.com/sonata-project/SonataCoreBundle/blob/master/Model/BaseManager.php#L40
is not happy when you're using **doctrine_phpcr** and **doctrine_mongodb**, until you change these values.

OskarStark:
i think not @Soullivaneuh because before it didn't worked, as the placeholder was wrong

IMO its a bug and should be fixxed with this PR

@rande what do you mean?

Soullivaneuh:
If it's a bugfix, it should be proposed agains last stable branch (2.3).

renooo:
OK, I'll make the PR in the right branch.

PR #864, opened 2015-10-17
streamlined documentation

OskarStark:


OskarStark:
ping @javiereguiluz

Resources/doc/reference/installation.rst
```diff
@@ -23,10 +23,9 @@ Installation
 
 Retrieve the bundle with composer:
 
-.. code-block:: sh
-
-    php composer.phar require sonata-project/media-bundle --no-update
+.. code-block:: bash
 
+    $ php composer.phar require sonata-project/media-bundle --no-update
```
javiereguiluz: In Symfony we already use `composer require ...` instead of `php composer.phar require ...` But I don't know if the latter is the common Sonata practice.

Soullivaneuh: I would prefer `composer require ...` and without the `$`.

OskarStark: you are right, we should change this, but we should do this across all the Sonata bundles

Soullivaneuh: If you have time... feel free! ;-)

OskarStark: will change this later in a separate PR

PR #870, opened 2015-10-31
Removed unused properties

core23:
This properties were never used.

### DIFF
--- a/Command/AddMediaCommand.php
+++ b/Command/AddMediaCommand.php
@@ -19,6 +19,10 @@ use Symfony\Component\Console\Output\OutputInterface;
 class AddMediaCommand extends BaseCommand
 {
     protected $quiet = false;
+
+    /**
+     * @deprecated This property is deprecated since version 2.4 and will be removed in 3.0.
+     */
     protected $output;
 
     /**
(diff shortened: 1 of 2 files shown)


Command/AddMediaCommand.php
```diff
@@ -19,7 +19,6 @@
 class AddMediaCommand extends BaseCommand
 {
     protected $quiet = false;
-    protected $output;
```
Soullivaneuh: Should be deprecated instead.

core23:
Added deprecation notice

Command/AddMediaCommand.php
```diff
@@ -19,6 +19,9 @@
 class AddMediaCommand extends BaseCommand
 {
     protected $quiet = false;
+    /**
```
Soullivaneuh: New line is required.

Command/AddMediaCommand.php
```diff
@@ -19,6 +19,9 @@
...
+     * @deprecated this property is never used and will be removed
```
Soullivaneuh: This property is deprecated since version 2.4 and will be removed in 3.0.

PR #871, opened 2015-11-04
Removed unused parameter

core23:
This property is never used and should be removed in the next major version.


OskarStark:
Should wie create a PR Directly for 3.0 ?

PR #878, opened 2015-11-10
added missing use statements

gnorby88:


OskarStark:
already fixxed in #879

PR #898, opened 2015-11-20
Created new command for cleaning upload directory

core23:
This command scans the upload directory and removes unlinked files.

### DIFF
--- a/Command/CleanMediaCommand.php
+++ b/Command/CleanMediaCommand.php
@@ -1,0 +1,134 @@
+<?php
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Command;
+
+use Sonata\MediaBundle\Provider\FileProvider;
+use Sonata\MediaBundle\Provider\MediaProviderInterface;
+use Sonata\MediaBundle\Provider\Pool;
+use Symfony\Bundle\FrameworkBundle\Command\ContainerAwareCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Filesystem\Exception\IOException;
+use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Component\Finder\Finder;
+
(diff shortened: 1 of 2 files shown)


OskarStark:
:+1: 

a ``dry-run`` option would be great!

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -0,0 +1,268 @@
...
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|MediaManagerInterface
+     */
+    private $mediaManager;
+
+    /**
+     * @var \PHPUnit_Framework_MockObject_MockObject|Local
+     */
+    private $fileSystemLocal;
```
OskarStark: Can you create a ``BaseCommandTest`` or something like this, i think this would reduce duplicate code. what do you think?

OskarStark:
@core23 can you please check travis?

Thank you

core23:
The problem is that the test relies on the symfony 2.7 ``FilesystemTestCase``. 

Any good idea how to solve this without copy&paste the 2.7 stuff in this bundle? @OskarStark

OskarStark:
ok for now we could use this in this way, can you please check travis, afterwards i will merge this

core23:
There are two travis problems at the moment: mongo orm got broken and the test for this command relies on the ``FilesystemTestCase`` which exists since symfony 2.4. In 2.3 it should rely on ``FilesystemTest`` but still no idea how to solve it :/

OskarStark:
not sure if this is fixed after a rebase, but can you give it a tray?

core23:
Ping @OskarStark

OskarStark:
Thank you

Command/CleanMediaCommand.php
```diff
@@ -0,0 +1,134 @@
...
+        if (count($fileParts) > 1 && $fileParts[0] == 'thumb') {
+            return $mediaManager->findOneBy(array(
+                    'id'      => $fileParts[1],
+                    'context' => $context,
+                )) != null;
+        }
+
+        return count($mediaManager->findBy(array(
+                'providerReference' => $filename,
+                'providers'         => $this->getProviders(),
```
rschumacher: Should be `                'providerName'         => $this->getProviders(),` instead?
In my case (a project using Doctrine ORM), the command works only with the latter.

core23: Can you please create a PR or an issue for that?

rschumacher: created issue #954
No PR yet because I am not sure whether above-mentioned patch will really implement the expected behavior of the command.

PR #905, opened 2015-11-24
fixed sync command when using consumer as thumbnail service

toooni:
# Issue with the sync-command
When using `sonata.media.thumbnail.consumer.format` as thumbnail service I recently fixed an issue where the job was sent too early resulting in a race condition. The solution was to postpone firing of the event with the `event_dispatcher` until the event `kernel.finish_request` is thrown.
This resulted in an other issue where the sync command was not working anymore because there is no `kernel.finish_request` event being fired. First I thought that the solution might be to add another listener on the `console.terminate` command. But this would be resulting in a strange behavior of the sync command (The jobs would be published until the command was run completely).
An other issue would have been that the deletion was not made through a notification, but the generation might. This would have been a problem on production environments because of the duration between deleting and creating.

This PR might not be very "sexy" but I really don't know what else to do. I am open to discuss alternatives.


OskarStark:
ping @rande

toooni:
@OskarStark @rande reminder: the "sonata:media:sync-thumbnails" commend does not work in environments using "sonata.media.thumbnail.consumer.format" until this PR is merged.

OskarStark:
@core23 am i right, that you created the sync command?

core23:
No, the file was created by @pulse00. I fixed the cs on this file.

Have a look at https://github.com/sonata-project/SonataMediaBundle/commits/master/Command/SyncThumbsCommand.php

Soullivaneuh:

Regards.

toooni:
@Soullivaneuh I'll do. I really appreciate your work (very much!!!), but next time I hope you guys aren't implementing something like StyleCI at the same time ;). In my case it's not much work because the code changes are very small. But I would understand if some contributors might be frustrated and won't create a new PR.

OskarStark:
@toooni thank you for resubmitting and reworking your PR!

We hope that we will not have such massive changes in the future ;-)

PR #909, opened 2015-11-26
Refactor thumbnail Twig helper

elvetemedve:
Delegate the responsibility of populating the options to the provider object.
That way the provider can add new items to the options array, making the media extension
more extendable.

### DIFF
--- a/Tests/Twig/Extension/MediaExtensionTest.php
+++ b/Tests/Twig/Extension/MediaExtensionTest.php
@@ -1,0 +1,126 @@
+<?php
+/*
+ * This file is part of the Sonata package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Sonata\MediaBundle\Model\MediaInterface;
+use Sonata\MediaBundle\Twig\Extension\MediaExtension;
+
+/**
+ * Class MediaExtensionTest.
+ *
+ * Unit test of MediaExtension class.
+ *
+ * @author Geza Buza <bghome@gmail.com>
+ */
+class MediaExtensionTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @var Sonata\MediaBundle\Provider\MediaProviderInterface
+     */
+    private $provider;
+
+    /**
+     * @var Twig_TemplateInterface
+     */
+    private $template;
+
+    /**
+     * @var Twig_Environment
+     */
+    private $environment;
+
+    /**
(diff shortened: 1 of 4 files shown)


core23:
Can you please add some tests.

elvetemedve:
@core23 I've added a unit test for this use case.

I think Scrutinizer has a configuration problem, because it hang up halfway. Can you please check that?

OskarStark:
Thank you @elvetemedve

Twig/Extension/MediaExtension.php
```diff
@@ -169,7 +169,7 @@ public function thumbnail($media = null, $format, $options = array())
 
         $options = array_merge($defaultOptions, $options);
 
-        $options['src'] = $provider->generatePublicUrl($media, $format);
+        $options = $provider->getHelperProperties($media, $format, $options);
```
greg0ire: Why was this merged? It should have raised some questions IMO @core23 @OskarStark . It's already in 3 releases now, so we can't revert it can we? How can we fix this? Please have a look a #1065

OskarStark: @greg0ire it was in my beginning and tests were added which were all green

greg0ire: > @greg0ire it was in my beginning and tests were added which were all green

Yeah plus the tests look good and so does the commit message, so I can understand why you merged, I might have done the same had I seen that. We must not hesitate to ask clarifications when we see strange things like that.

@elvetemedve answered about this on #1065 , he told us "Since it's FOSS software, feel free to contribute. ;)". So basically, we're on our own. The problem is, I have no clue how to undo this without creating other BC-breaks, or what the proper fix.

core23: Maybe we can call both methods and merge the `$options` to be BC with old and newer versions?

greg0ire: That would be BC but would not fix #1065 : @navalex would still get the array to string conversion error, b/c the youtube providers adds an array `player_parameters` option, that is not compatible with the thumbnail template.

greg0ire: See #1113

Tests/Twig/Extension/MediaExtensionTest.php
```diff
@@ -0,0 +1,126 @@
+<?php
+/*
+ * This file is part of the Sonata package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
```
core23: The namespace is missing

PR #913, opened 2015-12-03
Symfony 3 compatibility

mbabker:
This PR should fix the remaining Symfony 3 compatibility issues based on the test suite and fixes one bug.

- Changes Travis-CI to test on stable versions of 2.8 and 3.0
- Changes `Form::bind()` calls to `Form::submit()`
- Handles deprecation of the `SecurityContextInterface` and associated `security.context` service
- Handles deprecation of the pre-2.5 validator API
- Corrects a call to a non-existing method in the `FormatValidator`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ cache:
 env:
   global:
     - PATH="$HOME/.composer/vendor/bin:$PATH"
-    - SYMFONY_DEPRECATIONS_HELPER=weak
 
 matrix:
   fast_finish: true
@@ -31,16 +30,16 @@ matrix:
       env: SYMFONY_VERSION=2.3.*
     - php: 5.6
       env: SYMFONY_VERSION=2.7.*
-    - php: 5.6
-      env: SYMFONY_VERSION=2.8.*
     - php: 5.5
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION=3.0.* ACL_VERSION="dev-master"
     - php: 5.6
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION=3.0.* ACL_VERSION="dev-master"
     - php: 7.0
-      env: SYMFONY_VERSION="3.0.*" ACL_VERSION="dev-master"
+      env: SYMFONY_VERSION=3.0.* ACL_VERSION="dev-master"
 
   allow_failures:
+    - php: 7.0
+      env: SYMFONY_VERSION=3.0.* ACL_VERSION="dev-master"
     - php: hhvm
 
 before_script:
--- a/CDN/CloudFront.php
+++ b/CDN/CloudFront.php
@@ -171,7 +171,7 @@ class CloudFront implements CDNInterface
     public function setClient($client)
     {
         if (!$client instanceof CloudFrontClient) {
-            trigger_error('The '.__METHOD__.' expects a CloudFrontClient as parameter.', E_USER_DEPRECATED);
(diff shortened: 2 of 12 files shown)


mbabker:
Added bonus, fixed a failure for the `COMPOSER_FLAGS="--prefer-lowest"` job by setting `sonata-project/block-bundle` minimum dev dependency to `^2.3.7`.

Ener-Getick:
You must allow to use ``FOSRestBundle`` 2.0 to install symfony 3.0 (``FOSRestBundle`` 1.7 doesn't support it).

Ener-Getick:
:+1:

mbabker:
Pending https://github.com/nelmio/NelmioApiDocBundle/pull/752 being merged NelmioApiDocBundle should stop causing build errors.  At a glance not entirely sure on the conflicts with the security-acl package yet.

Ener-Getick:
@mbabker https://github.com/nelmio/NelmioApiDocBundle/pull/752 has been merged ;)

mbabker:
Back into dependency circles it seems.  The last build is at a point where it needs https://github.com/KnpLabs/KnpMarkdownBundle/pull/80 to be merged so the dependency on it from SonataFormatterBundle resolves correctly in the Symfony 3 environment.

rande:
@mbabker the main blocker is the SonataUserBundle. It locks all the chain...

Ener-Getick:
@rande you can use [this](https://github.com/nelmio/NelmioApiDocBundle/blob/master/.travis.yml#L27) to test symfony 3.0 when all the dependencies are not stable.

.travis.yml
```diff
@@ -24,28 +24,26 @@ matrix:
     - php: 5.4
       env: SYMFONY_VERSION=2.8.*
     - php: 5.6
-      env: SYMFONY_VERSION=2.8.* CS_FIXER=run
+      env: SYMFONY_VERSION=2.8.* SYMFONY_DEPRECATIONS_HELPER=strict CS_FIXER=run
```
Soullivaneuh: Please just remove this line instead: https://github.com/sonata-project/SonataMediaBundle/pull/913/files#diff-354f30a63fb0907d4ad57269548329e3R16

mbabker: The `SYMFONY_DEPRECATIONS_HELPER=weak` line or the strict setting for this build?  IMO you should have at least one build using the strict setting on 2.8 as that will force the tests to fail if any deprecated methods are in use.

Soullivaneuh: The `SYMFONY_DEPRECATIONS_HELPER=weak` line.

This was added before because of a lot of deprecated stuff.

Now, if all deprecations are corrected, we should keep that way. ;-)

mbabker: Done.  Also renamed the two tests testing legacy behavior in the bundle to silence the uncaught deprecation warnings.

core23:
Please rebase, so we can merge this

mbabker:
Rebased.

core23:
... and squash please

mbabker:
Done

OskarStark:
ping @core23

core23:
Looks okay to me

OskarStark:
looks like there are only two deprecation warnings or am i missing sth @core23 ?

core23:
Can you have a look at travis @mbabker 

Would like to merge this to become stable

mbabker:
Should be good now.  PHP 7 is moved to allowed failures because the `doctrine/mongo` library does not yet support their new driver.  The Travis failures that are left are pre-existing (the 3 failures on the `--prefer-lowest` run and one PHPCS issue).

core23:
Thank you @mbabker

PR #923, opened 2016-01-19
add region to S3Client config

juliensantos87:
pass the region for aws sdk to generate signature

https://github.com/aws/aws-sdk-php/blob/2.7.27/src/Aws/S3/S3Client.php#L272


dao:
+1 - this is also required for S3Client doesBucketExist to work. If region is not provided in the client constructor, amazon returns a 302 redirect requesting you to use an URL including region, causing gaufrette to conclude that the bucket does not exist and try to create the bucket in AwsS3::ensureBucketExists. Since the bucket does exist, you end up getting a BucketAlreadyOwnedByYouException with the message "Your previous request to create the named bucket succeeded and you already own it.". Simply passing region to the constructor as above resolves this error.

Soullivaneuh:

Regards.

PR #925, opened 2016-01-25
Make download id more restrictive

slu125:


rande:
id can be a string, uuid and not only a digit

PR #932, opened 2016-01-30
Fix broken build

### LINKED ISSUE #930: Documentation is broken
Build is not passing, because something is wrong with sphinx

see sphinx-doc/sphinx#2264

greg0ire:
Fixes #930

### DIFF
--- a/Resources/doc/reference/advanced_configuration.rst
+++ b/Resources/doc/reference/advanced_configuration.rst
@@ -78,7 +78,7 @@ Full configuration options:
 
         filesystem:
             local:
-                directory:  %kernel.root_dir%/../web/uploads/media
+                directory:  "%kernel.root_dir%/../web/uploads/media"
                 create:     false
 
             ftp:
--- a/Resources/doc/reference/api.rst
+++ b/Resources/doc/reference/api.rst
@@ -70,7 +70,7 @@ This would look like this for the cURL call:
 
 And like this for the request body:
 
-.. code-block:: http
+::
 
     ------WebKitFormBoundaryFhX9k2FPT3sQos00
     Content-Disposition: form-data; name="name"
--- a/Resources/doc/reference/creating_a_provider_class.rst
+++ b/Resources/doc/reference/creating_a_provider_class.rst
@@ -310,7 +310,7 @@ The thumbnail template is common to all media and it is quite simple:
 
 .. code-block:: html+jinja
 
-    <img {% for name, value in options %}{{name}}="{{value}}" {% endfor %} />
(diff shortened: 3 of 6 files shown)


greg0ire:
Look this one closely, there are some controversial things in it.

greg0ire:
Looks like there are non-doc issues too…

OskarStark:
but i think we can merge this anyway, what do you think?
Can you please squash your commits?

Thank you!!

greg0ire:
I don't know why you want me to squash my beautiful, independent commits into a big one. I grouped the related commits together so there are now 2 commits instead of 4. Is it about disk space?

core23:
Because every fix or feature should seen as an atomic change in the code. This will keep the history clean.

See: https://github.com/sonata-project/SonataAdminBundle/blob/master/CONTRIBUTING.md#sending-a-pull-request

greg0ire:
I'm fixing two different things here : should I make 2 PRs ?

core23:
Which two things are you fixing? IMO you are fixing https://github.com/sonata-project/SonataMediaBundle/issues/930 which is one concrete bug.

greg0ire:
I'm fixing #930 , and I'm adding a missing newline at the end of a file, but #930 would be fixed without that cs commit. It's just that I saw that issue on the way, and ship it in this PR.

core23:
IMO in this case you can combine these things into one commit.

greg0ire:
I beg to disagree. Imagine you want to make a fix on a file that uses tabs instead of spaces. First thing you do, you change tabs to spaces, and commit that, and then work on your fix. The diff of your fix stays clean, while the diff of a squash of both commits would be quite hard to read. Plus, one can revert one commit and not the other, just in case your fix turns out to be not so good (it happens). Of course you could make two PRs, but they would be conflicting with each other, and the change from tabs to space does not really warrant a PR, does it?

greg0ire:
I think we should fix this first : https://github.com/sonata-project/SonataAdminBundle/pull/3563

rande:
@greg0ire just squash your commit please ;) one PR = one commit, easier to revert ;)

greg0ire:
The PR is about my first commit, the second one is CS, it is something else (and could be in another PRà If my first commit turns out to be bad, you can simply revert the first commit. And just let the cs commit stay. But your repo, your rules… I guess I'll just **remove** my commit from the PR if no one wants to humor me on this.

greg0ire:
BTW @rande : what about the ACL problem? I summed it up [here](https://github.com/sonata-project/SonataDoctrinePhpcrAdminBundle/pull/357). It is one of the reasons why the build still fails.

PR #934, opened 2016-02-02
Type checking to avoid fatal error

morille:
Some of my users had fatal errors while using the API because they made mistake in the Media Form.
This check allow to return a clear description of the problem.

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -192,6 +192,11 @@ class FileProvider extends BaseProvider
             return;
         }
 
+        // is_file will cause a fatal error if binary content is not a string
+        if (!is_string($media->getBinaryContent())) {
+            throw new \RuntimeException(sprintf('Invalid data provided for binary content, choose among: string, %s, %s', File::class, Request::class));
+        }
+
         // if the binary content is a filename => convert to a valid File
         if (!is_file($media->getBinaryContent())) {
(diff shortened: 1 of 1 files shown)


OskarStark:
looks good to me, but can you please check travis? thank you

chlabelek:
I fixed the PHP compatibility issues.
Now there are 2 remaining errors but they do not depend on me I think.

morille:
Rebased.

Provider/FileProvider.php
```diff
@@ -192,6 +192,15 @@ protected function fixBinaryContent(MediaInterface $media)
             return;
         }
 
+        // is_file will cause a fatal error if binary content is not a string
+        if (!is_string($media->getBinaryContent())) {
```
OskarStark: what if `$media->getBinaryContent() == Symfony\Component\HttpFoundation\File\File`
does `is_string()` return `true`?

ping @greg0ire

greg0ire: Do you literally mean the class name or an object of this class?

OskarStark: I mean, what if this returns an object?

greg0ire: I don't think so.

Soullivaneuh:

Regards.

PR #938, opened 2016-02-03
Replaced deprecated bind with handleRequest method

core23:


OskarStark:
looks good to me :+1:

PR #952, opened 2016-02-18
Make the documentation parseable by sphinx 1.3.5

greg0ire:
- quote meta characters in YAML
- remove highlighting for http block
  Http does not seem to be recognized at all. If it is, there is no color
  in the output, so…
- use another syntax for twig snippet

Conflicts:
	Resources/doc/reference/installation.rst


greg0ire:
Goddamns cs problems!

OskarStark:
Thank you

PR #960, opened 2016-02-28
Fix media access times increment in session variable

enekochan:
Pre-increment should be used otherwise the $times session variable never increases.

### DIFF
--- a/Resources/doc/reference/security.rst
+++ b/Resources/doc/reference/security.rst
@@ -112,7 +112,7 @@ Let's create the following strategy : a media can be downloaded only once per se
                 return false;
             }
 
-            $this->getSession()->set($this->sessionKey, $times++);
+            $this->getSession()->set($this->sessionKey, $times + 1);
 
             return true;
         }


OskarStark:
ping @core23 @greg0ire

Resources/doc/reference/security.rst
```diff
@@ -112,7 +112,7 @@ Let's create the following strategy : a media can be downloaded only once per se
                 return false;
             }
 
-            $this->getSession()->set($this->sessionKey, $times++);
+            $this->getSession()->set($this->sessionKey, ++$times);
```
greg0ire: There is no need to increment `$times`, is there? So maybr write `$times + 1`. Good catch anyway!

enekochan: Yes you're right, I just used the ++ to use the same approach but it makes no sense. Changed it to $times + 1

greg0ire:
Thanks! Please fixup your commits together, keeping only the first commit's message.

enekochan:
Sorry for the mess with the commits :(

greg0ire:
No problem. Do you know how to fixup?

enekochan:
Not very sure. I've managed to change the commit message in the last one but there are still 2 commits.

greg0ire:
You can follow there instructions : https://github.com/sonata-project/SonataAdminBundle/blob/master/CONTRIBUTING.md#squash-your-commits (just replace squash with fixup, it works the same way)

enekochan:
Done! Thanks for the help with the fixup :)

greg0ire:
Good to merge IMO @OskarStark : we should have a look at the 2.0 build

core23:
I was wondering, why you merged this against the 2.0 branch? It's really old and not supported anymore 
ping @OskarStark @greg0ire

greg0ire:
Oh wow, didn't catch that sorry, I don't know this bundle that well.

enekochan:
My bad. Should I open another PR then? In what branch?

core23:
AFAIK only the latest branch + master are active for all projects @greg0ire 
Can you create this fix for the 2.3 branch @enekochan

greg0ire:
Maybe we should create one last tag on 2.0 and remove it.

enekochan:
Just to be sure @core23 Should I do the PR on the master branch or in 2.3?

core23:
2.3

enekochan:
Fixed in #966

PR #966, opened 2016-03-04
Fix media access times increment in session variable

enekochan:
Post-increment operator replaced by + 1 so the session variable increases every time the media is downloaded


core23:
Thank you @enekochan

Issue #991, opened 2016-04-24
Usage without sonata admin ?


Padam87:
php app\console cache:warmup

**Fatal error: Class 'Sonata\AdminBundle\Controller\CRUDController' not found in .....\vendor\sonata-project\media-bundle\Controller\GalleryAdminController.php on line 19**

To fix it I have to add the admin bundle to the dependencies, which is far from ideal. Any suggestions?

greg0ire:
Wow. Does anyone know why the admin bundle isn't required ? It is referenced in `GalleryAdminController`, apparently.

core23:
It is an optional dependency and should work without the ``SonataAdminBundle``, see: https://github.com/sonata-project/SonataMediaBundle/blob/master/DependencyInjection/SonataMediaExtension.php#L86

greg0ire:
Indeed, I just saw that in the doc.

greg0ire:
@Padam87 , can you use `app/console -vvv cache:warmup` and give us a full stack trace ?

Padam87:
It is only an issue when warming the cache after a clear. Unfortunately the -vvv option wasn't really helpful.

Fatal error: Class 'Sonata\AdminBundle\Controller\CRUDController' not found in .....\vendor\sonata-project\media-bundle\Controller\GalleryAdminController.php on line 19

                                                                                            
  [Symfony\Component\Debug\Exception\ClassNotFoundException]                                
  Attempted to load class "CRUDController" from namespace "Sonata\AdminBundle\Controller".  
  Did you forget a "use" statement for another namespace?                                   
                                                                                            

Exception trace:
 () at ....\vendor\sonata-project\media-bundle\Controller\GalleryAdminController.php:19

Padam87:
Maybe I'm wrong, but this seems like a class loader problem?

greg0ire:
Well if you don't have the admin bundle, it will not load, so yeah…

greg0ire:
How does warmup find this class ? Is there any route pointing to it?

greg0ire:
@Padam87 : have you had a look at the routes (`debug:router` ? )

Padam87:
Yep, routing is not the problem. It is caused by the DiExtra bundle.

https://github.com/schmittjoh/JMSDiExtraBundle/blob/master/HttpKernel/ControllerInjectorsWarmer.php#L46

Can be fixed by adding the file to the blacklist.

```yml
jms_di_extra:
    cache_warmer:
[...]

PR #992, opened 2016-04-24
Added DiExtra warmup problem fix documentation

Padam87:
#991

### DIFF
--- a/Resources/doc/reference/troubleshooting.rst
+++ b/Resources/doc/reference/troubleshooting.rst
@@ -106,3 +106,27 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
             local:
                 directory:  "%kernel.root_dir%/../web/uploads/media"
                 create:     false
+
+ 
+Error on cache:warmup
+---------------------
+
+You may encounter the following error when warming up the cache if you don't have the admin bundle installed:
+
+.. code-block:: text
+
+    Fatal error: Class 'Sonata\AdminBundle\Controller\CRUDController'
+    not found in .....\vendor\sonata-project\media-bundle\Controller\GalleryAdminController.php on line 19
+    
+This problem is most likely caused by the JMSDiExtraBundle's cache warmer (https://github.com/schmittjoh/JMSDiExtraBundle/blob/master/HttpKernel/ControllerInjectorsWarmer.php).
+
+To fix the problem, you will have to add the Sonata Media controllers to the blacklist.
+
+.. code-block:: yaml
+
+    jms_di_extra:
+        cache_warmer:
+            controller_file_blacklist:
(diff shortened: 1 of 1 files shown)


Resources/doc/reference/troubleshooting.rst
```diff
@@ -106,3 +106,27 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
...
+---------------------
+
+You may encounter the following error when warming up the cache if you don't have the admin bundle installed:
+
+.. code-block:: text
+
+    Fatal error: Class 'Sonata\AdminBundle\Controller\CRUDController'
+    not found in .....\vendor\sonata-project\media-bundle\Controller\GalleryAdminController.php on line 19
+    
+This problem is most likely caused by the JMSDiExtraBundle's cache warmer (https://github.com/schmittjoh/JMSDiExtraBundle/blob/master/HttpKernel/ControllerInjectorsWarmer.php).
```
greg0ire: Either `JMSDiExtraBundle's cache warmer` or `the JMSDiExtraBundle cache warmer`, but not both (I prefer the former). Also, try to keep lines under 120 characters.

Resources/doc/reference/troubleshooting.rst
```diff
@@ -106,3 +106,27 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
             local:
                 directory:  "%kernel.root_dir%/../web/uploads/media"
                 create:     false
+
+ 
+Error on cache:warmup
```
greg0ire: I would name this section "Usage without the admin bundle" (this might not be the last issue).

greg0ire:
Now that I think of it, it might be possible to configure the JMS bundle with [this trick](http://symfony.com/doc/current/cookbook/bundles/prepend_extension.html)

If you're motivated, could you give it a look ? It would go that way : detect if the admin bundle is present, and if it isn't prepend the blacklist configuration. If you don't have the time, we'll probably merge this instead.

Padam87:
Sorry, but this is it for me on this issue.

core23:
This looks like a hack to solve this issue. Any possibility to fix this the proper way? Maybe we should fix ``JMSDiExtraBundle``.

core23:
There is an issue https://github.com/schmittjoh/JMSDiExtraBundle/issues/126

greg0ire:
> This looks like a hack to solve this issue. Any possibility to fix this the proper way? Maybe we should fix JMSDiExtraBundle.

Yeah and not every body uses The di extra bundle, (it's an extra bundle, after all), so maybe we should just merge this in the documentation, and that's all.

Soullivaneuh:

Regards.

PR #1001, opened 2016-05-20
Support binary content which have in memory path

ksom:
Hi,

This PR provide support for stream wrapped paths (including Gaufrette) as pre-processed binaryContent.

In `FileProvider#setFileContents()` (**L 310**), the "realpath" to the pre-processed file is determined using `$media->getBinaryContent()->getRealPath()`. However, this returns `false` for stream wrapped paths (at least Gaufrette does). This causes a `ContextErrorException: Warning: file_get_contents(): Filename cannot be empty` error in Symfony.

With this PR, we now rely on  `$media->getBinaryContent()->getPathName()` if (and only if) `getRealPath()` fails, allowing the use of stream wrapped files.

This allow us, for example, to use Gaufrette's InMemoryAdapter to store pre-processed files and stop dealing with temporary files.

Since there were no tests for `setFileContents()`, I didn't know what kind of tests to add. The new behavior is relevant only in cases where the previous was failing anyway.

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -318,8 +318,10 @@ class FileProvider extends BaseProvider
             return;
         }
 
-        if ($media->getBinaryContent() instanceof File) {
-            $file->setContent(file_get_contents($media->getBinaryContent()->getRealPath()), $metadata);
+        $binaryContent = $media->getBinaryContent();
+        if ($binaryContent instanceof File) {
+            $path = $binaryContent->getRealPath() ?: $binaryContent->getPathname();
+            $file->setContent(file_get_contents($path), $metadata);
 
             return;
         }
--- a/Tests/Provider/FileProviderTest.php
+++ b/Tests/Provider/FileProviderTest.php
@@ -224,4 +224,85 @@ class FileProviderTest extends \PHPUnit_Framework_TestCase
             array(null, $media3),
         );
     }
+
+    public function testBinaryContentWithRealPath()
+    {
+        $media = $this->getMock('Sonata\MediaBundle\Model\MediaInterface');
+
+        $media->expects($this->any())
+            ->method('getProviderReference')
+            ->willReturn('provider');
+
+        $media->expects($this->any())
+            ->method('getId')
+            ->willReturn(10000);
+
+        $media->expects($this->any())
+            ->method('getContext')
+            ->willReturn('context');
+
+        $binaryContent = $this->getMockBuilder('Symfony\Component\HttpFoundation\File\File')
+            ->setMethods(array('getRealPath', 'getPathname'))
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $binaryContent->expects($this->atLeastOnce())
+            ->method('getRealPath')
+            ->willReturn(__DIR__.'/../fixtures/file.txt');
+
+        $binaryContent->expects($this->never())
+            ->method('getPathname');
+
+        $media->expects($this->any())
+            ->method('getBinaryContent')
+            ->willReturn($binaryContent);
+
+        $provider = $this->getProvider();
+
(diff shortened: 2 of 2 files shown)


ksom:
@OskarStark @rande

core23:
Please be patient. Someone will review your changes

ksom:
@core23 Sorry I poke because I did PR in Sonata Admin and take a while (more than a month to have a response). I'm patient I just did that to ensure you have seen it. Sorry again

core23:
Can you please add some test.

We know, that there was some trouble with PRs and issues in the past. We try our bests to improve the process, but we are all part time developers who are working in our free time for this project.

ksom:
@core23 
Tests added.

Soullivaneuh:
Do you have some concrete example of your issue? Some piece of code maybe?

Maybe add some documentation if necessary?

Thanks.

ksom:
```php
$adapter = new \Gaufrette\Adapter\InMemory();
$filesystem = new \Gaufrette\Filesystem($adapter);
$map =  \Gaufrette\StreamWrapper::getFilesystemMap();
$map->set('image', $filesystem);
 \Gaufrette\StreamWrapper::register();

$bin = new SplFileInfo('gaufrette://image/'.$name);
file_put_contents($bin->getPathname(), $content);

$media = new Media;
$media->setProviderName($this->mediaProvider->getName());
$media->setContext($this->context);
$media->setName($name);
$media->setBinaryContent($bin);
$this->mediaProvider->updateMetadata($media);

$em->persist($media);
$em->flush($media);
```
PostPersist will be catch and call to `FileProvider#setFileContents()` and throwing exception, $bin->getRealPath() returns `false`

Provider/FileProvider.php
```diff
@@ -318,8 +318,10 @@ protected function setFileContents(MediaInterface $media, $contents = null)
             return;
         }
 
-        if ($media->getBinaryContent() instanceof File) {
-            $file->setContent(file_get_contents($media->getBinaryContent()->getRealPath()), $metadata);
+        $bin = $media->getBinaryContent();
```
OskarStark: `$bin` -> `$binary` please, or this this variable already used?

OskarStark:
I currently don't have time to reproduce this, but the tests look ok to me.

Can you please check travis, or is it related to anything else?

ksom:
@OskarStark 
I replaced $bin with $binaryContent.
For travis, not related to my PR IMO

Have nice week end

ksom:
@OskarStark 
Does it miss something ?

greg0ire:
Please make the build green(er) by fixing the coding style.

greg0ire:
You can do that by running `make cs`.

Tests/Provider/FileProviderTest.php
```diff
@@ -224,4 +224,85 @@ public function mediaProvider()
...
+        $media->expects($this->any())
+            ->method('getBinaryContent')
+            ->willReturn($binaryContent);
+
+        $provider = $this->getProvider();
+
+        $setFileContents = new \ReflectionMethod('Sonata\MediaBundle\Provider\FileProvider', 'setFileContents');
+        $setFileContents->setAccessible(true);
+
+        $setFileContents->invoke($provider, $media);
```
greg0ire: Why are you testing a protected method? We don't care how the object method works, all we care about is the public method of the object works. Would it be too cumbersome to test `setFileContents` through a public method?

ksom: If there were pre-existing tests to check the behavior of setFileContents (via the public postPersist/postUpdate), I would have gladly added to them. However, there aren't. These tests seemed the most robust way to check the behavior introduced by my PR given the current state of the repo. Although creating unit tests for postPersist and postUpdate would be the perfect solution, it is clearly outside the scope of this PR.

greg0ire: Fair enough

greg0ire:
Thanks @ksom

PR #1002, opened 2016-05-22
Replace the global symfony dependency by only necessary components

Th3Mouk:


Soullivaneuh:
Well, it seems you added all the framework dependencies instead. :trollface: 

Which dependencies required by framework-bundle is not required by this one?

Th3Mouk:
@Soullivaneuh :trollface: 
- asset
- cache
- class-loader
- polyfill
- doctrine/cache
- doctrine/annotations

Soullivaneuh:
Thanks @Th3Mouk. I cherry-picked your commit: https://github.com/sonata-project/SonataMediaBundle/commit/953ecd619b994dc0b155c61357b01a30b10afe01

PR #996, opened 2016-05-03
datagrid category value must be object, it was being set to integer

ahmetakbn:
| Question       | Answer
|----------------|-------------------------------------------------------------
| Bundle version | sonata-project/media-bundle              dev-master b676113 
| Symfony version| symfony/symfony                          v2.8.4
| php version    | PHP 7.0.4-7ubuntu2

# Error
Deleting media object from list view was not being successful although getting the success message. 

# Steps to reproduce
On Media list view press filter once and then try to delete and object. Object is not being deleted. If you press the filter again and again, then you can delete the object.

# Change
Changed the value of category in datagrid to category object. It was being set only category id. If the category id is set by request from tree list, then we find one object by id and context and set the datagrid category to this object.

PR #1016, opened 2016-05-27
variable name should be category regarding to return

ahmetakbn:
Replaces #997
Refs #996 

### Subject

Variable name should be category regarding to return


OskarStark:
LGTM

PR #1017, opened 2016-05-27
Ensure ClassificationBundle related entities (Categories & Cie) are refreshed after a ObjectManager#clear

lemoinem:
Closes sonata-project/SonataClassificationBundle#60
Replaces #780 

### Changelog

```markdown
### Added
- Added `Sonata\MediaBundle\Listener\ORM\MediaEventSubscriber::onClear` to clear the `rootCategories` cache when the EntityManager is cleared.

### Changed
- `Sonata\MediaBundle\Listener\ORM\MediaEventSubscriber::onClear` now subscribes to `onClear` too.

### Fixed
- Reuse of root categories instances after the entity manager has been cleared.
```

### Subject

The issue wasn't reported in MediaBundle because the error is reported on ClassificationBundle Entities. Please refer to sonata-project/SonataClassificationBundle#60 for more details

### To do

- [x] Rebase on 3.x
- [x] Update the tests

PR #1019, opened 2016-05-27
Fix CleanMediaCommandTest

lemoinem:
Closes N/A (Not passing tests)

### Changelog

```markdown
### Added
- Added `Sonata\MediaBundle\Tests\Command\CleanMediaCommandTest::assertOutputFoundInContext` to check for multiple lines of the output of the command within a context

### Fixed
- Failing tests `testExecuteFilesExistsVerbose`, `testExecute` and `testExecuteDryRun`
```

### Subject

While rebasing #1017, I noticed only a few kind of tests were failing. I took the occasion to fix them.

### To do

- [x] Make the tests green

### DIFF
--- a/Tests/Command/CleanMediaCommandTest.php
+++ b/Tests/Command/CleanMediaCommandTest.php
@@ -212,8 +212,14 @@ class CleanMediaCommandTest extends TestCase
 
         $output = $this->tester->execute(array('command' => $this->command->getName(), '--verbose' => true));
 
-        $this->assertRegExp('@Context: foo\s+\'qwertz.ext\' found\s+\'thumb_1_bar.ext\' found\s+done!@', $this->tester->getDisplay());
-
+        $this->assertOutputFoundInContext(
+            '/Context: foo\s+(.+)\s+done!/ms',
+            array(
+                '\'qwertz.ext\' found',
+                '\'thumb_1_bar.ext\' found',
+            ),
+            $this->tester->getDisplay()
+        );
         $this->assertSame(0, $output);
     }
 
@@ -244,8 +250,14 @@ class CleanMediaCommandTest extends TestCase
 
         $output = $this->tester->execute(array('command' => $this->command->getName(), '--dry-run' => true));
 
-        $this->assertRegExp('@Context: foo\s+\'qwertz.ext\' is orphanend\s+\'thumb_1_bar.ext\' is orphanend\s+done!@', $this->tester->getDisplay());
-
+        $this->assertOutputFoundInContext(
+            '/Context: foo\s+(.+)\s+done!/ms',
+            array(
+                '\'qwertz.ext\' is orphanend',
+                '\'thumb_1_bar.ext\' is orphanend',
+            ),
+            $this->tester->getDisplay()
+        );
         $this->assertSame(0, $output);
     }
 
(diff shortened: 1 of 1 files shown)


lemoinem:
I tried to fix the StyleCI, but the patch seems weird... I'm not sure what should be done.

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -74,6 +74,31 @@ class CleanMediaCommandTest extends TestCase
...
+     * @param array  $expected Excerpts of text expected to be found in the output
+     * @param string $output   Output to find
+     */
+    private function assertOutputFoundInContext($extractor, $expected, $output)
+    {
+        preg_match_all($extractor, $output, $matches);
+
+        $found = false;
+        foreach ($matches[1] as $match) {
+            foreach ($expected as $e) {
```
OskarStark: could we change `$e` to something more speakable?

OskarStark:
@core23 can you have a look at this please?

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -212,8 +237,7 @@ public function testExecuteFilesExistsVerbose()
 
         $output = $this->tester->execute(array('command' => $this->command->getName(), '--verbose' => true));
 
-        $this->assertRegExp('@Context: foo\s+\'qwertz.ext\' found\s+\'thumb_1_bar.ext\' found\s+done!@', $this->tester->getDisplay());
-
+        $this->assertOutputFoundInContext('/Context: foo\s+(.+)\s+done!/ms', array('\'qwertz.ext\' found', '\'thumb_1_bar.ext\' found'), $this->tester->getDisplay());
```
greg0ire: Please linebreak this as well as the lines below.

greg0ire:
> I tried to fix the StyleCI, but the patch seems weird... I'm not sure what should be done.

It's about moving private methods at the bottom of the class ;)

lemoinem:
@greg0ire @OskarStark Thanks for your feedback. I fixed the details you commented upon.

The remaining StyleCI patches are not related to my PR. I'll check to submit another one with these patches.

greg0ire:
I think @Soullivaneuh  can fix that if he hasn't already

lemoinem:
@greg0ire Yup, indeed it was already patched in 3.x, I rebased so StyleCI would be :green_heart: :)

greg0ire:
Can you briefly explain how the tests were failing and how you fixed them in your commit message ? It would help a lot when reviewing / viewing history. In general try to stick to these rules : https://github.com/sonata-project/SonataAdminBundle/blob/3.x/CONTRIBUTING.md#the-commit-message

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -276,8 +288,39 @@ public function testExecute()
...
+     */
+    private function assertOutputFoundInContext($extractor, $expected, $output)
+    {
+        preg_match_all($extractor, $output, $matches);
+
+        $found = false;
+        foreach ($matches[1] as $match) {
+            foreach ($expected as $excerpt) {
+                if (false === strpos($match, $excerpt)) {
+                    continue 2;
```
OskarStark: never saw this `continue 2;` oO

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -276,8 +288,39 @@ public function testExecute()
...
+                }
+                $found = true;
+                break;
+            }
+        }
+
+        $this->assertTrue($found, sprintf('Unable to find "%s" in "%s" with extractor "%s"', implode('", "', $expected), $output, $extractor));
```
OskarStark: please linebreak this

OskarStark:
LGTM 👍

Tests/Command/CleanMediaCommandTest.php
lemoinem: It's breaking the innermost `foreach` (`foreach ($expected`) and continue the enclosing one (`foreach ($matches[1]`).

If you feel it's a hard-to-read or esoteric syntax, I could probably replace it by something better, although I'm not sure what (additional private method maybe)...

greg0ire: It's indeed a bit hard to get… replace with something else if you can.

lemoinem:
@greg0ire @OskarStark Fixed the commit message (is this better?) and the long line.

greg0ire:
> is this better?

lol… yes… yes it is! :clap:

Tests/Command/CleanMediaCommandTest.php
OskarStark: is there any documentation about this, i only know `continue;`

greg0ire: http://php.net/manual/fr/control-structures.continue.php

OskarStark: but in this case `continue 2;` is the same like `break;` isn't it?

greg0ire: It's completely valid so if you can't find anything else, leave it like this. Not even sure you should add a comment.

OskarStark: yes we can keep it 👍

greg0ire: > but in this case continue 2; is the same like break; isn't it?

Hmmm yes it might be, since there is nothing after the inner foreach. @lemoinem , can you check?

lemoinem: It's not equivalent to a `break` because there is the `$found = true;` after the inner loop which I wish to skip too.

lemoinem:
@OskarStark @greg0ire I removed the `continue 2;` and replaced it by a secondary private method.

OskarStark:
LGTM 👍 

/cc @greg0ire

Tests/Command/CleanMediaCommandTest.php
```diff
@@ -276,8 +288,58 @@ public function testExecute()
...
+        ));
+    }
+
+    /**
+     * Returns whether every needle can be found as a substring of the haystack.
+     *
+     * @param string $haystack
+     * @param array  $needles  Array of (potential) substrings of the haystack
+     */
+    private function doesContainAll($haystack, $needles)
```
greg0ire: "doesContainAll" => "containsAll", or do you really want to insist on this fact, like "Yes, it **does** contain all parts" (not very likely, but…) ? Otherwise it easier to understand like this, thanks!

greg0ire: I'll add `RTM` back when this is fixed (or when you give me a good reason for this name).

OskarStark: we should change the name 👍

lemoinem: No, I'll change it. I was a bit split on how to name it and tried to go with something explicit.

greg0ire:
> would you prefer a static one?

No, no a private is really better.

greg0ire:
Thanks a lot @lemoinem !

PR #806, opened 2015-07-01
[2.3] [MediaType] Made form options "provider" and "context" required

phansys:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #805 
| License       | MIT
| Doc PR        |

Closes #805.

Given the following example:
```php
$builder->add('media', 'sonata_media_type');
```
**Before:**
```
unable to retrieve the provider named : ``
```
**After:**
```
The required option "provider" is missing.
```

**TODO**:
- [x] Check given ```provider``` value against ```Pool::getProviderList()```

PR #1026, opened 2016-05-29
Made form options "provider" and "context" required in MediaType

core23:
Rebase of #806

### Changelog

```markdown
### Fixed
- The options `provider` and `context` are now required
```

### Subject

The options must be set in order to show the media_type widget.

### To do

- [x] Update the tests

### DIFF
--- a/Form/Type/MediaType.php
+++ b/Form/Type/MediaType.php
@@ -104,14 +104,37 @@ class MediaType extends AbstractType
      */
     public function configureOptions(OptionsResolver $resolver)
     {
-        $resolver->setDefaults(array(
-            'data_class' => $this->class,
-            'provider' => null,
-            'context' => null,
-            'empty_on_new' => true,
-            'new_on_update' => true,
-            'translation_domain' => 'SonataMediaBundle',
-        ));
+        $resolver
+            ->setDefaults(array(
+                'data_class' => $this->class,
+                'empty_on_new' => true,
+                'new_on_update' => true,
+                'translation_domain' => 'SonataMediaBundle',
+            ))
+            ->setRequired(array(
+                'provider',
+                'context',
+            ));
+
+        // NEXT_MAJOR: Remove this hack when dropping support for symfony 2.3
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+            $resolver
+                ->setAllowedTypes('provider', 'string')
+                ->setAllowedTypes('context', 'string')
+                ->setAllowedValues('provider', $this->pool->getProviderList())
+                ->setAllowedValues('context', array_keys($this->pool->getContexts()))
+            ;
+        } else {
+            $resolver
+                ->setAllowedTypes(array(
+                    'provider' => 'string',
+                    'context' => 'string',
+                ))
+                ->setAllowedValues(array(
+                    'provider' => $this->pool->getProviderList(),
+                    'context' => array_keys($this->pool->getContexts()),
+                ));
+        }
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Form/Type/MediaType.php
```diff
@@ -96,14 +96,37 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
+                'empty_on_new' => true,
+                'new_on_update' => true,
+                'translation_domain' => 'SonataMediaBundle',
+            ))
+            ->setRequired(array(
+                'provider',
+                'context',
+            ));
+
+        // Use proper syntax based on Symfony version
```
OskarStark: can you write down for WHICH Symfony version and WHEN we could drop this?

Tests/Form/Type/MediaTypeTest.php
```diff
@@ -0,0 +1,116 @@
...
+use Symfony\Component\Form\Test\TypeTestCase;
+
+/**
+ * @author Christian Gripp <mail@core23.de>
+ */
+class MediaTypeTest extends TypeTestCase
+{
+    /**
+     * @expectedException        \Symfony\Component\OptionsResolver\Exception\MissingOptionsException
+     * @expectedExceptionMessage The required options "context", "provider" are missing.
```
OskarStark: @greg0ire @Soullivaneuh do we had a decision about this `expectedException` annotation?

If i'm right, @greg0ire like `$this->setExpectedException` more ;-)

OskarStark:
This looks good to me, but is this a BC break when we make properties required?

Soullivaneuh:
@core23, currently, what happen exactly if those options are not provided? Does the bundle work?

Tests/Form/Type/MediaTypeTest.php
greg0ire: You're right, and @Soullivaneuh likes it better too since I showed him [this article](https://thephp.cc/news/2016/02/questioning-phpunit-best-practices)

greg0ire: Hmmm, I re-read the article and when dropping old phpunit support, we will have to move to `expecteException` (as opposed to `setExpectedException`). But I guess won't be very hard…

core23:
When you don't pass the `provider` option you get a `\InvalidArgumentException ` from the `Pool::getProvider` when rendering the form.

When you don't pass the `context` option you get a `\RuntimeException` from `MediaEventSubscriber ::getRootCategory` after submetting the form.

OskarStark:
In this case i think we can merge this in 3.x, because an Exception is raised anyway, isn't it?

core23:
I need some help on the tests for symfony 3.

The API has changed on `FormFactoryInterface::create` which expects a string now. We can't pass an instance of `AbstractType` to this anymore, instead we should pass the classname to it. This is a problem, because I need to mock the constructor parameters.

OskarStark:
jenkins failed @core23

OskarStark:
any news on this @core23 ?

Form/Type/MediaType.php
```diff
@@ -104,14 +104,37 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
+                'empty_on_new' => true,
+                'new_on_update' => true,
+                'translation_domain' => 'SonataMediaBundle',
+            ))
+            ->setRequired(array(
+                'provider',
+                'context',
+            ));
+
+        // BC handle for symfony >=2.6
```
OskarStark: @NEXT_MAJOR missing

core23:
Finally I made the tests work for all symfony versions.

greg0ire:
LGTM

OskarStark:
Thank you @core23 👍

PR #622, opened 2014-09-09
Add BaseVideoProvider::getReferenceUrl() method

EmmanuelVella:
This PR add a ``BaseVideoProvider::getReferenceUrl()`` public abstract method, just to easily get the original URL.

PR #1027, opened 2016-05-29
Add BaseVideoProvider::getReferenceUrl() method

core23:
Rebase of https://github.com/sonata-project/SonataMediaBundle/pull/622

### Changelog

```markdown
### Added
- Created `getReferenceUrl` method for all video providers
```

### Subject

Created a new method which returns the reference url of a video.

### To do

- [X] Update the tests

### DIFF
--- a/Provider/BaseVideoProvider.php
+++ b/Provider/BaseVideoProvider.php
@@ -50,6 +50,11 @@ abstract class BaseVideoProvider extends BaseProvider
     {
         parent::__construct($name, $filesystem, $cdn, $pathGenerator, $thumbnail);
 
+        // NEXT_MAJOR: remove this check!
+        if (!method_exists($this, 'getReferenceUrl')) {
+            @trigger_error('The method "getReferenceUrl" is required with the next major release.', E_USER_DEPRECATED);
+        }
+
         $this->browser = $browser;
         $this->metadata = $metadata;
     }
@@ -201,6 +206,16 @@ abstract class BaseVideoProvider extends BaseProvider
     public function postRemove(MediaInterface $media)
     {
     }
+
+    // NEXT_MAJOR: Uncomment this method
+    /*
+     * Get provider reference url.
+     *
+     * @param MediaInterface $media
+     *
+     * @return string
+     */
+    // abstract public function getReferenceUrl(MediaInterface $media);
 
     /**
      * @throws \RuntimeException
--- a/Provider/DailyMotionProvider.php
+++ b/Provider/DailyMotionProvider.php
@@ -98,7 +98,7 @@ class DailyMotionProvider extends BaseVideoProvider
      */
     public function updateMetadata(MediaInterface $media, $force = false)
     {
-        $url = sprintf('http://www.dailymotion.com/services/oembed?url=http://www.dailymotion.com/video/%s&format=json', $media->getProviderReference());
+        $url = sprintf('http://www.dailymotion.com/services/oembed?url=%s&format=json', $this->getReferenceUrl($media));
 
         try {
             $metadata = $this->getMetadata($media, $url);
@@ -125,7 +125,19 @@ class DailyMotionProvider extends BaseVideoProvider
      */
     public function getDownloadResponse(MediaInterface $media, $format, $mode, array $headers = array())
     {
-        return new RedirectResponse(sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference()), 302, $headers);
+        return new RedirectResponse($this->getReferenceUrl($media), 302, $headers);
+    }
+
+    /**
+     * Get provider reference url.
+     *
+     * @param MediaInterface $media
+     *
+     * @return string
+     */
+    public function getReferenceUrl(MediaInterface $media)
+    {
+        return sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference());
     }
 
     /**
--- a/Tests/Provider/YouTubeProviderTest.php
+++ b/Tests/Provider/YouTubeProviderTest.php
@@ -206,4 +206,11 @@ class YouTubeProviderTest extends AbstractProviderTest
         $this->assertSame(100, $properties['player_parameters']['height']);
         $this->assertSame(100, $properties['player_parameters']['width']);
     }
+
+    public function testGetReferenceUrl()
+    {
+        $media = new Media();
+        $media->setProviderReference('123456');
+        $this->assertEquals('http://www.youtube.com/watch?v=123456', $this->getProvider()->getReferenceUrl($media));
+    }
 }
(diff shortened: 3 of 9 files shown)


Provider/BaseVideoProvider.php
```diff
@@ -179,6 +179,16 @@ public function postRemove(MediaInterface $media)
...
 
+    // NEXT_MAJOR: Uncomment this line
+    /*
+     * Get provider reference url.
+     *
+     * @param MediaInterface $media
+     *
+     * @return string
+     */
+    // public function getReferenceUrl(MediaInterface $media);
```
greg0ire: Should be abstract ?

Provider/DailyMotionProvider.php
```diff
@@ -125,7 +125,22 @@ public function updateMetadata(MediaInterface $media, $force = false)
      */
     public function getDownloadResponse(MediaInterface $media, $format, $mode, array $headers = array())
     {
-        return new RedirectResponse(sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference()), 302, $headers);
+        return new RedirectResponse($this->getReferenceUrl($media), 302, $headers);
+    }
+
+    /**
+     * {@inheritdoc}
```
OskarStark: not true, this can be done in the master PR afterwards

Provider/DailyMotionProvider.php
```diff
@@ -125,7 +125,22 @@ public function updateMetadata(MediaInterface $media, $force = false)
...
+     */
+    public function getReferenceUrl(MediaInterface $media)
+    {
+        return sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference());
+    }
+
+    public function testGetReferenceUrl()
```
OskarStark: why a test method here? oO

OskarStark:
Test für DailymotionProvider is missing

Provider/DailyMotionProvider.php
Soullivaneuh: :wine_glass: 

Or maybe :beers:?

:wink:

Soullivaneuh:
@core23 Could you please add some documentation.

Also, even if it's not a deprecation, I think it would be great to have an upgrade one because this method would be mandatory on next major.

core23:
@Soullivaneuh I don't agree with that. Every change in the interface / abstract class was held in the major change where we finally introduce this method. I would do this in the major PR.

core23:
Ping @Soullivaneuh

Tests/Provider/YouTubeProviderTest.php
```diff
@@ -206,4 +206,11 @@ public function testHelperProperties()
         $this->assertSame(100, $properties['player_parameters']['height']);
         $this->assertSame(100, $properties['player_parameters']['width']);
     }
+
+    public function testGetReferenceUrl()
+    {
+        $media = new Media();
+        $media->setProviderReference('123456');
+        $this->assertEquals('http://www.youtube.com/watch?v=123456', $this->getProvider()->getReferenceUrl($media));
+    }
```
greg0ire: This is getting repetitive… maybe you could introduce an abstract test case with a `getExpectedUrl($providerReference)` method ?

Soullivaneuh:
> @Soullivaneuh I don't agree with that. Every change in the interface / abstract class was held in the major change where we finally introduce this method. I would do this in the major PR

Don't know. Ping @sonata-project/contributors for opinions.

greg0ire:
Well one could see not implementing the new method as "deprecated". It would make the transition smoother, that's for sure…

Tests/Provider/YouTubeProviderTest.php
core23: Note the scope of this PR. We could move more methods to an abstract TestCase.

core23:
> Well one could see not implementing the new method as "deprecated". It would make the transition smoother, that's for sure… this could be done by adding a check in the constructor

Don't thinkt that's a good idea. This would mess up the constructor and slow down the performance or is it the symfony way?

Tests/Provider/YouTubeProviderTest.php
greg0ire: Ok

greg0ire:
> This would mess up the constructor and slow down the performance or is it the symfony way?

Regarding the performance, if you're thinking about reflection, I think this can be achieved without. Just `method_exists` would suffice IMO

With method exists : https://3v4l.org/dsEdb/perf#tabs
With nothing : https://3v4l.org/bOFuq/perf#tabs

core23:
Added the `deprecation` warning. Don't know what kind of docs are missing.

Provider/BaseVideoProvider.php
```diff
@@ -179,6 +183,16 @@ public function postRemove(MediaInterface $media)
     {
     }
 
+    // NEXT_MAJOR: Uncomment this line
```
OskarStark: `Uncomment this line` -> `Uncomment this method`

Provider/DailyMotionProvider.php
OskarStark: or could this inherit from an uncommented method?

Provider/BaseVideoProvider.php
core23: Fixed

core23:
@Soullivaneuh Can you explain what documentation are you expecting?

core23:
Ping @Soullivaneuh

core23:
@Soullivaneuh ?

Soullivaneuh:
> @Soullivaneuh Can you explain what documentation are you expecting?

You introduced a new method, right? This one should be documented. :-)

core23:
> You introduced a new method, right? This one should be documented. :-)

Have a look at the phpDoc ;)

Soullivaneuh:
I'm talking about the RST documentation, with some samples. This is the only goal of this label BTW.

greg0ire:
Here is probably where the doc should be added @core23 : https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Resources/doc/reference/creating_a_provider_class.rst

core23:
Added the missing docs.

greg0ire:
LGTM

Provider/BaseVideoProvider.php
```diff
@@ -50,6 +50,10 @@ public function __construct($name, Filesystem $filesystem, CDNInterface $cdn, Ge
     {
         parent::__construct($name, $filesystem, $cdn, $pathGenerator, $thumbnail);
 
+        if (!method_exists($this, 'getReferenceUrl')) {
```
OskarStark: `NEXT_MAJOR: remove this check!`

OskarStark:
After my comment, this PR is `RTM`

core23:
Fixed @OskarStark

greg0ire:
Thanks @core23!

PR #999, opened 2016-05-11
Adding missing security test

ahmetakbn:

PR #1032, opened 2016-05-30
added missing use statement

ahmetakbn:
Refs #999

```markdown
### Fixed
- Added missing use statement to `Security/SessionDownloadStrategy.php`
```

### Subject

```php
use Symfony\Component\HttpFoundation\Session\Session;
```
was missing.

### DIFF
--- a/Security/SessionDownloadStrategy.php
+++ b/Security/SessionDownloadStrategy.php
@@ -14,6 +14,7 @@ namespace Sonata\MediaBundle\Security;
 use Sonata\MediaBundle\Model\MediaInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class SessionDownloadStrategy implements DownloadStrategyInterface


OskarStark:
Thank you @ahmetakbn 👍

OskarStark:
@greg0ire can you please merge it after travis is green?

greg0ire:
sure

greg0ire:
Wait, we add use statements when they are reference in the phpdoc ?

greg0ire:
I would remove the `getSession` method, instead, and use a local variable.

OskarStark:
i think it is used in the other strategies, too!

greg0ire:
It's private, so no.

greg0ire:
Plus injecting the whole container just for that is probably not the right solution.

greg0ire:
@ahmetakbn, can you please remove the private method? And if you feel like it, fix the huge Law of Demeter violation by injecting `@session` in the class.

greg0ire:
> And if you feel like it, fix the huge Law of Demeter violation by injecting @session in the class.

Actually this should be done in another PR, because of the BC-break.

Soullivaneuh:
> I would remove the getSession method instead, and use a local variable.

Change nothing, you will have to add a PHPDoc for the property.

ahmetakbn:
@greg0ire Changing the injection of container will also change the test. In this case to make it in another PR is better.

greg0ire:
> Change nothing, you will have to add a PHPDoc for the property.

The variable will be local to the only method where it is called, no need for a property.

greg0ire:
@ahmetakbn what about this PR? Should be close it or are you going to remove the private method anyway? Make sure you make the other PR against the master branch, because it will most certainly contain a BC-break.

ahmetakbn:
> @ahmetakbn what about this PR? Should be close it or are you going to remove the private method anyway? Make sure you make the other PR against the master branch, because it will most certainly contain a BC-break.

@greg0ire Yes I will remove the method, will inject the `@session` service and change the test. I am not sure about the BC-break but ok, I will create the PR to master branch. In this case this PR is against 3.x, if you want you can merge this or close.

greg0ire:
I'll close, since you will probably add the type hinting when injecting the dependency.

PR #1015, opened 2016-05-27
added test for SessionDownloadStrategy

ahmetakbn:
Replaces #999

```markdown
### Added
- Added tests for `SonataMediaBundle/Security/SessionDownloadStrategy`
```

### Subject

Test for `SonataMediaBundle/Security/SessionDownloadStrategy`

PR #1033, opened 2016-05-30
Changed injection of container in SessionDownloadStrategy and added test

ahmetakbn:
Replaces #1032 and #1015

### Changelog

```markdown
### Changed
- Injection of `session` instead of the whole `container` in `Security/SessionDownloadStrategy`
```

### DIFF
--- a/Security/SessionDownloadStrategy.php
+++ b/Security/SessionDownloadStrategy.php
@@ -12,26 +12,15 @@
 namespace Sonata\MediaBundle\Security;
 
 use Sonata\MediaBundle\Model\MediaInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
 /**
- * Class SessionDownloadStrategy.
- *
  * @author Ahmet Akbana <ahmetakbana@gmail.com>
  */
 class SessionDownloadStrategy implements DownloadStrategyInterface
 {
-    /**
-     * @var ContainerInterface
-     *
-     * @deprecated Since version 3.x, will be removed in 4.0.
-     * NEXT_MAJOR : remove this property
-     */
-    protected $container;
-
     /**
      * @var TranslatorInterface
      */
@@ -53,31 +42,14 @@ class SessionDownloadStrategy implements DownloadStrategyInterface
     private $session;
 
     /**
-     * @param TranslatorInterface                 $translator
-     * @param ContainerInterface|SessionInterface $session
-     * @param int                                 $times
+     * @param TranslatorInterface $translator
+     * @param SessionInterface    $session
+     * @param int                 $times
      */
-    public function __construct(TranslatorInterface $translator, $session, $times)
+    public function __construct(TranslatorInterface $translator, SessionInterface $session, $times)
     {
-        // NEXT_MAJOR : remove this block and set session from parameter.
-        if ($session instanceof ContainerInterface) {
-            @trigger_error(
-                'Using an instance of Symfony\Component\DependencyInjection\ContainerInterface is deprecated since 
-                version 3.x and will be removed in 4.0. 
-                Use Symfony\Component\HttpFoundation\Session\SessionInterface instead.',
-                E_USER_DEPRECATED
-            );
-
-            $this->session = $session->get('session');
-        } elseif ($session instanceof SessionInterface) {
-            $this->session = $session;
-        } else {
-            throw new \InvalidArgumentException(
-                '$session should be an instance of Symfony\Component\HttpFoundation\Session\SessionInterface'
-            );
-        }
-
         $this->times = $times;
+        $this->session = $session;
         $this->translator = $translator;
     }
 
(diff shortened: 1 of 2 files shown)


Security/SessionDownloadStrategy.php
```diff
@@ -12,16 +12,16 @@
...
+use Symfony\Component\HttpFoundation\Session\Session;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class SessionDownloadStrategy implements DownloadStrategyInterface
 {
     /**
-     * @var ContainerInterface
+     * @var Session
      */
-    protected $container;
```
OskarStark: this is a BC break

OskarStark:
> Coverage increased (+0.4%) to 31.854%

👍

Resources/doc/reference/security.rst
```diff
@@ -74,77 +74,73 @@ Let's create the following strategy : a media can be downloaded only once per se
...
+         * @var string
+         */
         protected $sessionKey = 'sonata/media/session/times';
 
         /**
-         * @param \Symfony\Component\Translation\TranslatorInterface $translator
-         * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
-         * @param int $times
+         * @param TranslatorInterface $translator
+         * @param Session             $container
```
greg0ire: Container?

Soullivaneuh:
@ahmetakbn Thanks for your contribution. :+1: 

But I would like a BC way on the stable branch, could you try it?

greg0ire:
@Soullivaneuh : how would you make that BC? By removing type hinting on the second argument, or by adding an optional argument, and then remove it in a non-BC PR?

ahmetakbn:
@Soullivaneuh I will create an another PR againist 3.x with BC way.

Resources/doc/reference/security.rst
ahmetakbn: changed

Soullivaneuh:
> @Soullivaneuh : how would you make that BC? By removing type hinting on the second argument, or by adding an optional argument, and then remove it in a non-BC PR?

You got it! :tada:

ahmetakbn:
Rebased and fixed merge conflicts

OskarStark:
I updated the CHANGELOG

@ahmetakbn can you please check the diff, because there is some renaming of a test method, which is not necessary IMO, but I could be wrong

Security/SessionDownloadStrategy.php
```diff
@@ -12,27 +12,18 @@
...
 
 use Sonata\MediaBundle\Model\MediaInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
 /**
- * Class SessionDownloadStrategy.
+ * Download strategy with time per Session.
```
OskarStark: should be removed imo

Tests/Security/SessionDownloadStrategyTest.php
```diff
@@ -49,40 +47,4 @@ public function testIsGrantedTrue()
...
-
-        $strategy = new SessionDownloadStrategy($translator, $container, 1);
-
-        $this->assertTrue($strategy->isGranted($media, $request));
-    }
-
-    /**
-     * @expectedException \InvalidArgumentException
-     */
-    public function testInvalidArgumentException()
```
OskarStark: why is this test removed?

ahmetakbn: this was a test to check if the parameter is either `session` or `container` instance. In the previous commit `NEXT_MAJOR : delete this method` was missing.

Security/SessionDownloadStrategy.php
ahmetakbn: ok

OskarStark: @greg0ire @Soullivaneuh what do you think?

Keep: 👍 
Remove: 👎

ahmetakbn:
> @ahmetakbn can you please check the diff, because there is some renaming of a test method, which is not necessary IMO, but I could be wrong

The method which is being tested returns `true` or `false` in just two cases. The naming was unnecessarily long. before the previous PR, there was another case to test, so the name was pointing one specific case. But now there is no need.

OskarStark:
> The method which is being tested returns true or false in just two cases. The naming was unnecessarily long. before the previous PR, there was another case to test, so the name was pointing one specific case. But now there is no need.

I just wanted to be sure, that this wasn't a mistake

Security/SessionDownloadStrategy.php
greg0ire: Remove because I don't understand what is meant by "with time per session" , I still have to look at the code to understand.

ahmetakbn: removed

OskarStark:
LGTM 👍 

@greg0ire why is SonataCI Quick Check failing?

greg0ire:
Just click on the link, and then on "Console output" : http://91.121.3.21/job/sonata-ci-bundle-media/413/console


> [xUnit] [ERROR] - The result file '/var/jenkins_home/jobs/sonata-ci-bundle-media/workspace/build/junit.xml' for the metric 'JUnit' is not valid. The result file has been skipped.

OskarStark:
Thank you 😄  but why 💃 

I would merge this, i don't understand the problem of coveralls

greg0ire:
No idea :P … and I don't know what is the actual phpunit command behind that, so… @rande ? Can you help with that?

PR #1037, opened 2016-06-01
Use AssetsHelper so we can generate relative urls when it's in sub-di…

Warxcell:
### Changelog

``` markdown
### Added
- Now using AssetsHelper to generate urls

### Added
- Added `Sonata\MediaBundle\CDN::setAssetsHelper` so we can set AssetsHelper to be used when generating urls.

### Changed
- Changed `Sonata\MediaBundle\CDN::getPath` so it will use AssetsHelper to generate correctly urls.
```
### Subject

Server CDN is not generating correct url when project is inside directory of domain. (You had to describe the whole url instead).

### DIFF
--- a/CDN/Server.php
+++ b/CDN/Server.php
@@ -8,7 +8,7 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
- 
+
 namespace Sonata\MediaBundle\CDN;
 
 use Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper;
--- a/CDN/Server.php
+++ b/CDN/Server.php
@@ -8,6 +8,7 @@
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
+ 
 namespace Sonata\MediaBundle\CDN;
 
 use Symfony\Bundle\FrameworkBundle\Templating\Helper\AssetsHelper;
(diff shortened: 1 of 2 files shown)


CDN/Server.php
```diff
@@ -11,12 +11,16 @@
...
+
 class Server implements CDNInterface
 {
+    protected $path;
+
     /**
-     * @var string
+     * @var AssetsHelper
      */
-    protected $path;
```
OskarStark: this is a BC break

Resources/config/media.xml
```diff
@@ -29,6 +29,9 @@
         <!-- CDN abstraction service -->
         <service id="sonata.media.cdn.server" class="Sonata\MediaBundle\CDN\Server">
             <argument />
+			<call method="setAssetsHelper">
```
OskarStark: wrong indention

OskarStark:
This needs to be submitted against master if you can't handle BC

core23:
We don't use author tags in the changelog.

Soullivaneuh:
> This needs to be submitted against master if you can't handle BC

This can and should be BC.

Warxcell:
Ok, what if I do not break BC and make it as new feature and option to be switched on?

core23:
Please rebase @Warxcell

core23:
Ping @Warxcell

core23:
Any news @Warxcell ?

core23:

PR #1042, opened 2016-06-05
Sync fix cli

toooni:
### Changelog

ConsumerThumbnail does not use backend when running on cli anymore.

```markdown
### Fixed
- syncing thumbnails with `sonata:media:sync-thumbnails` command when using `sonata.media.thumbnail.consumer.format` thumbnail consumer.
```

### Subject

When using `sonata.media.thumbnail.consumer.format` as thumbnail service I recently fixed an issue where the job was sent too early resulting in a race condition. The solution was to postpone firing of the event with the `event_dispatcher` until the event `kernel.finish_request` is thrown.
This resulted in an other issue where the sync command was not working anymore because there is no `kernel.finish_request` event being fired. First I thought that the solution might be to add another listener on the `console.terminate` command. But this would be resulting in a strange behavior of the sync command (The jobs would be published until the command was run completely).
An other issue would have been that the deletion was not made through a notification, but the generation might. This would have been a problem on production environments because of the duration between deleting and creating.

**IMPORTANT**: The diff might look complicated which is caused by indentation. The change itself is very simple:
```php
if (php_sapi_name() === 'cli') {
    $this->thumbnail->generate($provider, $media);
} else {
    /* UNCHANGED */
}
```

### DIFF
--- a/Thumbnail/ConsumerThumbnail.php
+++ b/Thumbnail/ConsumerThumbnail.php
@@ -73,26 +73,30 @@ class ConsumerThumbnail implements ThumbnailInterface
      */
     public function generate(MediaProviderInterface $provider, MediaInterface $media)
     {
-        $backend = $this->backend;
-        $id = $this->id;
-
-        $publish = function () use ($backend, $media, $id) {
-            $backend->createAndPublish('sonata.media.create_thumbnail', array(
-                'thumbnailId' => $id,
-                'mediaId' => $media->getId(),
-
-                // force this value as the message is sent inside a transaction,
-                // so we have a race condition
-                'providerReference' => $media->getProviderReference(),
-            ));
-        };
-
-        // BC compatibility for missing EventDispatcher
-        if (null === $this->dispatcher) {
-            trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
-            $publish();
+        if (php_sapi_name() === 'cli') {
+            $this->thumbnail->generate($provider, $media);
         } else {
-            $this->dispatcher->addListener('kernel.finish_request', $publish);
+            $backend = $this->backend;
+            $id = $this->id;
+
+            $publish = function () use ($backend, $media, $id) {
+                $backend->createAndPublish('sonata.media.create_thumbnail', array(
+                    'thumbnailId' => $id,
+                    'mediaId' => $media->getId(),
+
+                    // force this value as the message is sent inside a transaction,
+                    // so we have a race condition
+                    'providerReference' => $media->getProviderReference(),
+                ));
+            };
+
+            // BC compatibility for missing EventDispatcher
+            if (null === $this->dispatcher) {
+                trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
+                $publish();
+            } else {
+                $this->dispatcher->addListener('kernel.finish_request', $publish);
+            }
         }
     }
 


Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,26 +73,36 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
...
 
-        // BC compatibility for missing EventDispatcher
-        if (null === $this->dispatcher) {
-            trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
-            $publish();
-        } else {
-            $this->dispatcher->addListener('kernel.finish_request', $publish);
+            // BC compatibility for missing EventDispatcher
+            if (null === $this->dispatcher) {
+                trigger_error(
```
Soullivaneuh: This change is not related. Please revert.

Soullivaneuh:
Could you please:

* Fix StyleCI?
* Rebase?
* Squash your commits?

Concerning the change, except my comment I have nothing to say but I'm not sure this is the right thing to do as I don't use this bundle.

I let someone else manage this one. cc @sonata-project/contributors

Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,27 +73,27 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
      */
     public function generate(MediaProviderInterface $provider, MediaInterface $media)
     {
```
greg0ire: "identation" => "indention"

Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,27 +73,27 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
...
-        } else {
-            $this->dispatcher->addListener('kernel.finish_request', $publish);
-        }
+            // BC compatibility for missing EventDispatcher
+            if (null === $this->dispatcher) {
+                trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
+                $publish();
+            } else {
+                $this->dispatcher->addListener('kernel.finish_request', $publish);
+            }
```
greg0ire: You're not fixing the indention, you're messing it up, aren't you?

Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,27 +73,37 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
      */
     public function generate(MediaProviderInterface $provider, MediaInterface $media)
     {
+        if (php_sapi_name() === 'cli') {
+            $this->thumbnail->generate($provider, $media);
+        } else {
             $backend = $this->backend;
-            $id = $this->id;
+            $id      = $this->id;
```
greg0ire: Don't align things please. And if you do, do it in a separate commit.

Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,27 +73,37 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
...
 
-                    // force this value as the message is sent inside a transaction,
-                    // so we have a race condition
-                    'providerReference' => $media->getProviderReference(),
-                ));
+                        // force this value as the message is sent inside a transaction,
+                        // so we have a race condition
+                        'providerReference' => $media->getProviderReference(),
+                    )
+                );
```
greg0ire: Why are you touching this? What does it have to do with this commit?

Thumbnail/ConsumerThumbnail.php
toooni: I did neither. I just made an if statement to check if it is run on cli:
```
if (php_sapi_name() === 'cli') {
    $this->thumbnail->generate($provider, $media);
} else {
    /* UNCHANGED */
}
```

Thumbnail/ConsumerThumbnail.php
```diff
@@ -73,27 +73,37 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
...
             };
 
             // BC compatibility for missing EventDispatcher
             if (null === $this->dispatcher) {
-                trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
+                trigger_error(
+                    'Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.',
+                    E_USER_DEPRECATED
+                );
```
greg0ire: This should be in a separate commit

Thumbnail/ConsumerThumbnail.php
```diff
@@ -77,7 +77,7 @@ public function generate(MediaProviderInterface $provider, MediaInterface $media
             $this->thumbnail->generate($provider, $media);
         } else {
             $backend = $this->backend;
-            $id      = $this->id;
+            $id = $this->id;
```
greg0ire: Please fixup this commit with the previous one.

Thumbnail/ConsumerThumbnail.php
toooni: @greg0ire because StyleCI want's it that way

Thumbnail/ConsumerThumbnail.php
```diff
@@ -95,10 +95,7 @@ public function generate(MediaProviderInterface $provider, MediaInterface $media
 
             // BC compatibility for missing EventDispatcher
             if (null === $this->dispatcher) {
-                trigger_error(
-                    'Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.',
-                    E_USER_DEPRECATED
-                );
+                trigger_error('Since version 2.3.3, passing an empty parameter in argument 4 for __construct() in '.__CLASS__.' is deprecated and the workaround for it will be removed in 3.0.', E_USER_DEPRECATED);
```
greg0ire: Why do the unrelated change if it's for reverting afterwards? Please fixup this commit too.

toooni: @greg0ire because @Soullivaneuh wrote to revert that change..

Thumbnail/ConsumerThumbnail.php
greg0ire: Try to squash all your commits together, I'm reading PRs commit by commit and I just understood what you want to do, but it was hard.

Thumbnail/ConsumerThumbnail.php
greg0ire: Yeah but he did not actually mean use `git revert`, he meant "just make it disappear from the commits".

Thumbnail/ConsumerThumbnail.php
toooni: @greg0ire Sorry. I didn't think about that because GitHub has a nice feature to squash the commits on merge.

greg0ire:
https://github.com/sonata-project/SonataMediaBundle/pull/1042/files?w=1 to see the diff and ignoring whitespace. Try to make this diff as simple as you say it is.

Thumbnail/ConsumerThumbnail.php
toooni: @greg0ire I was going to squash all the commits (like requested in the comments from Soullivaneuh) but I wasn't fast enough to do this. It's not usual that somebody looks at the PR a minute after a new push. I am sorry I pushed it too early.

Thumbnail/ConsumerThumbnail.php
greg0ire: No problem, we do use it sometimes but we prefer when contributor provide something as clean as possible from the start, it's easier to read.

Thumbnail/ConsumerThumbnail.php
greg0ire: No worries, I probably should have read the PR body **before** reading the commits my bad.

toooni:
@greg0ire done

core23:
Some travis erros should be fixed. Please rebase @toooni

toooni:
@core23 done

core23:
The issue only occurs when using the cli and the `sonata:media:sync-thumbnails` command? 

Shouldn't we fix the issue there instead of the fixing the consumer? The consumers should now nothing about the outer world (the application, the console or anything else).

toooni:
@core23 yes. And I agree that the fix should happen somewhere else. This is just a quick hotfix. I dug a bit into the source and came to the conclusion that we'd have to make big architectural changes to solve this.
Conclusion: There is no easy way out. But an easy way to fix this (huge) problem for the moment. The current state is that the `sonata:media:sync-thumbnails` command does NOT work in a project where `sonata.media.thumbnail.consumer.format` is used.

rande:
@core23 agreed with you. I think this should be documented and can be resolved by using a specific environment with a dedicated configuration.

toooni:
@core23 @rande ok, you can close this PR. Just be assure that currently the `sonata:media:sync-thumbnails` command does **NOT** work in a project where `sonata.media.thumbnail.consumer.format` is used. There is **no error, no hint, nothing**. And I do not have the time now to work on a better solution at the moment.
Btw. this might be exactly the configuration which is used in bigger projects where image processing will be handled by a worker.

core23:
You said, that the `kernel.finish_request` event isn't fired when using the console command. Maybe we can manuel trigger it in the command?

toooni:
@core23 this sounds like a working solution. But this event will trigger every other listener on kernel.finish_request. We do not have control over what other applications are doing there.

core23:
Or maybe bind the `$publish` action to the `console.terminate` event too?

```php
                  $this->dispatcher->addListener('kernel.finish_request', $publish);           
                  $this->dispatcher->addListener('console.terminate', $publish);      
```

toooni:
@core23 Working solution too. I preferred the one in the PR because of the console output. You will not be able to show step by step which thumbnail is currently syncing.

core23:
> Working solution too. I preferred the one in the PR because of the console output. 

That no option at all, as mentioned before.

We should fix the command instead.

toooni:
@core23 this isn't going to be easy and won't happen soon. Please don't forget about this issue. People can easily lose hours with this one.
We are currently replacing SonataMediaBundle, so it doesn't matter for us anymore.

PR #1012, opened 2016-05-27
composer require should not have the flag --no-update

OskarStark:
Replaces #957 

### Subject

> composer require should not have the flag --no-update when only one package is required.
> 
> Would introduce some confusion

PR #1050, opened 2016-06-10
simplify composer command in docs

OskarStark:
Refs #1012

### Subject

See comment: https://github.com/sonata-project/SonataMediaBundle/pull/1012/files#r66574910

### DIFF
--- a/Resources/doc/reference/installation.rst
+++ b/Resources/doc/reference/installation.rst
@@ -25,7 +25,7 @@ Retrieve the bundle with composer:
 
 .. code-block:: bash
 
-    $ php composer.phar require sonata-project/media-bundle
+    $ composer require sonata-project/media-bundle
 
 Register these bundles in your AppKernel:
 


Resources/doc/reference/installation.rst
```diff
@@ -25,7 +25,7 @@ Retrieve the bundle with composer:
 
 .. code-block:: bash
 
-    $ php composer.phar require sonata-project/media-bundle
+    $ composer require sonata-project/media-bundle
```
Soullivaneuh: IMHO, the `$` sign is not necessary.

The code-block is set up for bash, this is enough.

cc @sonata-project/contributors

OskarStark: thats right, but for now this is consistent through all our bundles, because i refatored them all, a few months ago...

maybe we could do something like this with dev-kit?
Doing it manually would be a lot of work with nearly no benefit 😞

greg0ire: I'd say leave it as is, that's not that bad…

Soullivaneuh: > maybe we could do something like this with dev-kit?
> Doing it manually would be a lot of work with nearly no benefit :disappointed:

It would be great, but how?

> I'd say leave it as is, that's not that bad…

It is not a big deal indeed, but the `$` means a prompt. A prompt is not always a `$`.

For easy copy/paste, only the command should be here IMHO.

OskarStark: > It would be great, but how?

Maybe something like adding search and replace on specific files (*.rst), which result in a PR on this repo IF there is a match?

In this case we could search for: `$ composer.phar` and replace it with `composer` in `*.rst` files?

> It is not a big deal indeed, but the $ means a prompt. A prompt is not always a $.

True, but this should be executed in userland and not in `#`

> For easy copy/paste, only the command should be here IMHO.

I agree with you, but the work is not worth to do it

greg0ire: So you'd like dev-kit to have a refactoring command, right ? With two arguments, like "shell-command" and "file-pattern". Could be handy.

OskarStark: any news on this ?

greg0ire: The `$` sign was already there, so let's say it is not in the scope of this PR, shall we?

PR #1058, opened 2016-06-11
Remove Buzz dependency and replace it with guzzle

### LINKED ISSUE #951: Replace Buzz with PSR http-client
Refs https://github.com/sonata-project/SonataMediaBundle/issues/910

Today we are using both bundles for http connections. IMO we should drop Buzz and Guzzle and use the PSR http-client.

core23:
Closes #951

### Changelog

```markdown
### Changed
- Removed `buzz` dependency and replaced it with `GuzzleHttp`
```

### Subject

Today we use both dependencies for http connection. This change removed the buzz dependency completly. This shouldn't be done with a BC, because we had to load both to stay BC.

### To do

- [X] Update the tests
- [X] Add an upgrade note

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -54,7 +54,7 @@ class Configuration implements ConfigurationInterface
         $this->addProvidersSection($node);
         $this->addExtraSection($node);
         $this->addModelSection($node);
-        $this->addBuzzSection($node);
+        $this->addGuzzleSection($node);
         $this->addResizerSection($node);
 
         return $treeBuilder;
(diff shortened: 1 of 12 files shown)


core23:
We need php >=5.5 for this

greg0ire:
We should really consider using http://httplug.io/ cc @dbu

core23:
Should we use `HTTPlug` and suggest a `guzzle6-adapter` @sonata-project/contributors

dbu:
glad to help with httplug if you need, i think it would make sense to use it.

core23:
Closed in favor of https://github.com/sonata-project/SonataMediaBundle/pull/1060

PR #1060, opened 2016-06-13
Remove Guzzle and Buzz dependency and replace it with HTTPlug

### LINKED ISSUE #951: Replace Buzz with PSR http-client

core23:
Closes #951
### Changelog

``` markdown
### Changed
- Removed `Guzzle` and `Buzz` dependency and replaced it with `HTTPlug`
```
### Subject

Today we use both dependencies for http connection. This change removed both dependencies completly and replaces them with some abstract library. 

We can't handle this with BC, because HTTPlug requires at least PHP 5.4.
### To do

- [x] Update the tests
- [x] Update the documentation
- [X] Add an upgrade note

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -54,7 +54,7 @@ class Configuration implements ConfigurationInterface
         $this->addProvidersSection($node);
         $this->addExtraSection($node);
         $this->addModelSection($node);
-        $this->addBuzzSection($node);
+        $this->addHttpClientSection($node);
         $this->addResizerSection($node);
 
         return $treeBuilder;
@@ -428,23 +428,15 @@ class Configuration implements ConfigurationInterface
     /**
      * @param ArrayNodeDefinition $node
      */
-    private function addBuzzSection(ArrayNodeDefinition $node)
+    private function addHttpClientSection(ArrayNodeDefinition $node)
     {
         $node
             ->children()
-                ->arrayNode('buzz')
+                ->arrayNode('http')
                     ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('connector')->defaultValue('sonata.media.buzz.connector.curl')->end()
-                        ->arrayNode('client')
-                        ->addDefaultsIfNotSet()
-                        ->children()
-                            ->booleanNode('ignore_errors')->defaultValue(true)->end()
-                            ->scalarNode('max_redirects')->defaultValue(5)->end()
-                            ->scalarNode('timeout')->defaultValue(5)->end()
-                            ->booleanNode('verify_peer')->defaultValue(true)->end()
-                            ->scalarNode('proxy')->defaultNull()->end()
-                        ->end()
+                        ->scalarNode('client')->defaultValue('httplug.client.default')->end()
+                        ->scalarNode('message_factory')->defaultValue('httplug.message_factory.default')->end()
                     ->end()
                 ->end()
             ->end()
--- a/Metadata/AmazonMetadataBuilder.php
+++ b/Metadata/AmazonMetadataBuilder.php
@@ -13,7 +13,7 @@ namespace Sonata\MediaBundle\Metadata;
 
 use Aws\S3\Enum\CannedAcl;
 use Aws\S3\Enum\Storage;
-use Guzzle\Http\Mimetypes;
+use Core23\MimeUtil\MimeUtil;
 use Sonata\MediaBundle\Model\MediaInterface;
 
 class AmazonMetadataBuilder implements MetadataBuilderInterface
@@ -105,9 +105,8 @@ class AmazonMetadataBuilder implements MetadataBuilderInterface
      */
     protected function getContentType($filename)
     {
-        $extension = pathinfo($filename, PATHINFO_EXTENSION);
-        $contentType = Mimetypes::getInstance()->fromExtension($extension);
-
-        return array('contentType' => $contentType);
+        return array(
+            'contentType' => MimeUtil::getInstance()->getTypeFromFilename($filename),
+        );
     }
 }
--- a/Provider/BaseVideoProvider.php
+++ b/Provider/BaseVideoProvider.php
@@ -11,9 +11,11 @@
 
 namespace Sonata\MediaBundle\Provider;
 
-use Buzz\Browser;
 use Gaufrette\Filesystem;
+use Http\Client\HttpClient;
+use Http\Message\MessageFactory;
 use Imagine\Image\Box;
+use Psr\Http\Message\ResponseInterface;
 use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\CoreBundle\Model\Metadata;
 use Sonata\MediaBundle\CDN\CDNInterface;
@@ -28,9 +30,14 @@ use Symfony\Component\Validator\Constraints\NotNull;
 abstract class BaseVideoProvider extends BaseProvider
 {
     /**
-     * @var Browser
+     * @var HttpClient
      */
-    protected $browser;
+    protected $client;
+
+    /**
+     * @var MessageFactory
+     */
+    protected $messageFactory;
 
     /**
      * @var MetadataBuilderInterface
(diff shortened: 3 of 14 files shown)


core23:
Please don't review yet There is still some stuff to do.

greg0ire:
Great initiative!

Soullivaneuh:
:+1: 

But this can be done with BC.

Metadata/AmazonMetadataBuilder.php
```diff
@@ -13,7 +13,7 @@
 
 use Aws\S3\Enum\CannedAcl;
 use Aws\S3\Enum\Storage;
-use Guzzle\Http\Mimetypes;
+use Guzzle\Http\Mimetypes; # TODO: Find something common
```
dbu: @sagikazarmark do we have anything that could replace this?

Resources/config/provider.xml
```diff
@@ -63,7 +63,7 @@
             <argument/>
             <argument/>
             <argument type="service" id="sonata.media.thumbnail.format"/>
-            <argument type="service" id="sonata.media.buzz.browser"/>
+            <argument type="service" id="sonata.media.http.client"/>
```
dbu: this could be a tricky bit to keep BC. maybe a compilar pass that detects if buzz.browser service exists (legacy system) and in that case removes sonata.media.http.client service and instead sets an alias to the buzz.browser service.

Resources/doc/reference/installation.rst
```diff
@@ -26,6 +26,8 @@ Retrieve the bundle with composer:
 .. code-block:: bash
 
     $ php composer.phar require sonata-project/media-bundle
+    $ php composer.phar require php-http/guzzle6-adapter # recommended HTTP client
+
 
 Register these bundles in your AppKernel:
```
dbu: you require the httplug bundle. if it should be used, it needs to be instantiated too.

DependencyInjection/Configuration.php
```diff
@@ -428,22 +428,31 @@ private function addModelSection(ArrayNodeDefinition $node)
...
                     ->addDefaultsIfNotSet()
                     ->children()
-                        ->scalarNode('connector')->defaultValue('sonata.media.buzz.connector.curl')->end()
                         ->arrayNode('client')
                         ->addDefaultsIfNotSet()
                         ->children()
-                            ->booleanNode('ignore_errors')->defaultValue(true)->end()
-                            ->scalarNode('max_redirects')->defaultValue(5)->end()
+                            ->booleanNode('http_errors')->defaultValue(false)->end()
+                            ->arrayNode('allow_redirects')
```
dbu: these things are what HttplugBundle is supposed to provide. ideally, this bundle would just have an option to set the httpclient service name. the default value would be `httplug.client.default` which happens when you configure nothing.

the doc can then link to http://docs.php-http.org/en/latest/integrations/symfony-full-configuration.html to explain how one can configure the client.

Provider/BaseVideoProvider.php
```diff
@@ -83,7 +83,10 @@ public function getReferenceFile(MediaInterface $media)
         } else {
             $referenceFile = $this->getFilesystem()->get($key, true);
             $metadata = $this->metadata ? $this->metadata->get($media, $referenceFile->getName()) : array();
-            $referenceFile->setContent($this->browser->get($this->getReferenceImage($media))->getContent(), $metadata);
+
+            $response = $this->client->request('GET', $this->getReferenceImage($media));
```
dbu: you need the request factory to create a PSR-7 request. the client is no longer a factory for requests with PSR-7.

DependencyInjection/Configuration.php
dbu: and it should be possible to also configure the message factory, defaulting to `httplug.message_factory.default`. you need that in the places where you want to send requests, as the client is no longer a factory for messages.

dbu:
oh, just read the "please don't review yet" part now :-)

anyway, hope that my inputs can help you. regarding BC, you could keep requiring the buzz client, and start requiring the php-http/buzz-adapter and in the constructors allow both php-http client and buzz, and if its buzz wrap it in the buzz adapter to at least have your internal code consistent. which still leaves you with the constructor change for the message factory...

i am not sure its worth the hassle. if we want to be strict with BC, its really tricky with the services and all that. a half-baked BC could be to keep the config section for buzz and create a php-http buzz-adapter service from that configuration if the user explicitly configured something for buzz. then only people that extended one of the media classes would notice the BC break. not sure how likely it is that people extended those classes - and how many compromises you want with regards to BC...

Metadata/AmazonMetadataBuilder.php
sagikazarmark: I don't think so, but I hardly think this should be a responsibility of a HTTP something.

dbu: @core23 is this a package of its own? guzzle/mimetypes or something? if so i think its ok to depend on that specifically and use it here. no need to abstract everything. the mimetype handling has no outside dependencies or influences or configuration like the http clients do.

sagikazarmark: No, it is no. It is part of guzzle3

dbu: is there no guzzle 6 replacement then?

sagikazarmark: No, I couldn't find one.

sagikazarmark: https://packagist.org/search/?q=mime

Soullivaneuh:
@dbu We want to follow Symfony rules for BC.

The less direct BC break we have on next major, the smoother upgrade will be, on less issue we will have. :-)

core23:
This PR depends on https://github.com/sonata-project/dev-kit/pull/165, because HTTPlug requires at least PHP 5.4.

Resources/doc/reference/installation.rst
core23: Fixed

Resources/config/provider.xml
core23: This is now an alias for `httplug.client.default`.

core23:
@dbu Thanks for your great feedback 👍 

I've updated this PR, so it should be fine now. Can you give another review if this works as expected?

Metadata/AmazonMetadataBuilder.php
```diff
@@ -13,7 +13,7 @@
 
 use Aws\S3\Enum\CannedAcl;
 use Aws\S3\Enum\Storage;
-use Guzzle\Http\Mimetypes;
+use Core23\MimeUtil\MimeUtil;
```
sagikazarmark: Actually I think having a library which supports both a static list and PHP's `finfo` would be great.

composer.json
```diff
@@ -60,6 +62,8 @@
     "suggest": {
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
         "liip/imagine-bundle": "^0.9",
+        "php-http/buzz-adapter": "Guzzle HTTP client implementation",
```
sagikazarmark: Actually, the other way around.

sagikazarmark:
Looks good to me. For mocking, you might want to try the mock-client: https://github.com/php-http/mock-client

Metadata/AmazonMetadataBuilder.php
core23: My lib is still under development. Feel free to contribute ;)

Soullivaneuh:
> This PR depends on sonata-project/dev-kit#165, because HTTPlug requires at least PHP 5.4.

Still, I think this can be BC on 3.x. Use running PHP 5.3 can keep the old implementation, the others can start migrating to the new one.

core23:
The problem is travis... If we want to support both, we have to add `HTTPlug` as a dev dependency to run the tests, but the tests for 5.3 would fail.

Soullivaneuh:
Two solutions IMO:

* Do the BC way on 3.x, do not add the dependency on require neither require-dev but just suggest. We can add a target to require it from dev-kit
* If not, a BC way have to be kept in any case. So wait the 4.0 out with PHP 5.6+ support and make a BC way on it that will be removed in 5.0.

sagikazarmark:
If it is only about PHP version: changing it is not a BC break IMO. You can still use the constraint ^3.0 and you will get a >=3.0 version installed. Composer will resolve the version which still supports 5.3.

IMO supporting 5.3 is a bad idea anyway. Projects like this can show the way to people: use secure, up to date software.

dbu:
agree with mark on this one. php 5.3 is horribly outdated, even 5.4 is
end of life by now...

Soullivaneuh:
> If it is only about PHP version: changing it is not a BC break IMO. You can still use the constraint ^3.0 and you will get a >=3.0 version installed. Composer will resolve the version which still supports 5.3.

It is not a BC break if you maintain one branch by minor.

This is not our case, we maintain a 3.x, not a 3.1, 3.2 etc...

So we can't do this.

> agree with mark on this one. php 5.3 is horribly outdated, even 5.4 is
end of life by now...

I know. The new majors was out quick to start the new management system. Next ones dropping old stuff will come quite soon I hope.

core23:
Duplicate of https://github.com/sonata-project/SonataMediaBundle/pull/1181

PR #1070, opened 2016-06-30
Updated knplabs/gaufrette dependency

YaoOcelotl:
## Changelog

```markdown
### Changed
- Allow `knplabs/gaufrette` `^0.2`
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "sonata-project/easy-extends-bundle": "^2.1",
         "sonata-project/doctrine-extensions": "^1.0",
         "sonata-project/classification-bundle": "^3.0",
-        "knplabs/gaufrette": "~0.2",
+        "knplabs/gaufrette": "^0.1.6 || ^0.2",
         "imagine/imagine": "^0.6",
         "kriswallsmith/buzz": "^0.15",
         "jms/serializer-bundle": "^0.13 || ^1.0",


composer.json
```diff
@@ -38,7 +38,7 @@
         "sonata-project/easy-extends-bundle": "^2.1",
         "sonata-project/doctrine-extensions": "^1.0",
         "sonata-project/classification-bundle": "^3.0",
-        "knplabs/gaufrette": "^0.1.6",
+        "knplabs/gaufrette": "~0.2",
```
Soullivaneuh: * Must be `^0.2`
* Please on a minor, keep also `^0.1.6`

Soullivaneuh:
@YaoOcelotl Have you check for BC breaks between those versions?

greg0ire:
Use `git rebase` to get rid of the extra commit.

composer.json
core23: If we should keep `^0.1.6` we could close this PR, because `^0.1.6 | ^0.2` is the same as `^0.1.6`.

greg0ire: > If we should keep ^0.1.6 we could close this PR, because ^0.1.6 | ^0.2 is the same as ^0.1.6.

I don't think it is, actually, [version numbers starting with a zero are a special case](https://getcomposer.org/doc/articles/versions.md#caret)

Soullivaneuh: Exactly @greg0ire.

`0.x` releases are special, BC break are allowed between minors for this.

This also why I asked @YaoOcelotl to check if this bundle is still works with both versions.

YaoOcelotl:
@Soullivaneuh I have not found any BC breaks.

Soullivaneuh:
@YaoOcelotl We have also to be sure if related code is covered for that.

Could you please take a look?

Also please fix https://github.com/sonata-project/SonataMediaBundle/pull/1070#discussion_r69159701

core23:
Please rebase to get travis green @YaoOcelotl

Soullivaneuh:
What is missing to merge this PR:

- [ ] Change version constraint to `^0.1.6 | ^0.2`
- [ ] Rebase the branch to get Travis green

cc @YaoOcelotl 

Thanks.

YaoOcelotl:
Sorry about  the delay.

greg0ire:
> Use git rebase to get rid of the extra commit.

Apparently it did not work, but no worries, I squash-merged into one commit.

PR #1071, opened 2016-07-03
Support for FOSRestBundle 2.0

core23:

## Changelog

```markdown
### Fixed
- Support for FOSRestBundle 2.0
```

## To do

- [X] Update the tests

## Subject

Some methods were removed or changed in the 2.0 version.

### DIFF
--- a/Controller/Api/GalleryController.php
+++ b/Controller/Api/GalleryController.php
@@ -11,12 +11,11 @@
 
 namespace Sonata\MediaBundle\Controller\Api;
 
-use FOS\RestBundle\Controller\Annotations\QueryParam;
+use FOS\RestBundle\Context\Context;
 use FOS\RestBundle\Controller\Annotations\View;
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View as FOSRestView;
 use JMS\Serializer\SerializationContext;
-use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Sonata\DatagridBundle\Pager\PagerInterface;
 use Sonata\MediaBundle\Model\GalleryHasMediaInterface;
 use Sonata\MediaBundle\Model\GalleryInterface;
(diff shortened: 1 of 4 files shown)


Controller/Api/GalleryController.php
```diff
@@ -423,10 +422,19 @@ protected function handleWriteGalleryhasmedia(GalleryInterface $gallery, MediaIn
             $this->galleryManager->save($gallery);
 
             $view = FOSRestView::create($galleryHasMedia);
-            $serializationContext = SerializationContext::create();
-            $serializationContext->setGroups(array('sonata_api_read'));
-            $serializationContext->enableMaxDepthChecks();
-            $view->setSerializationContext($serializationContext);
+
+            // BC for FOSRestBundle < 2.0
+            if (method_exists($view, 'setSerializationContext')) {
```
Soullivaneuh: The reverse would be better.

AFAIK, `setSerializationContext` is still here on v2.0, but deprecated.

Tests/Controller/Api/GalleryControllerTest.php
```diff
@@ -147,7 +147,7 @@ public function testPostGalleryMediaGalleryhasmediaAction()
         $view = $galleryController->postGalleryMediaGalleryhasmediaAction(1, 2, new Request());
 
         $this->assertInstanceOf('FOS\RestBundle\View\View', $view);
-        $this->assertSame(200, $view->getStatusCode(), 'Should return 200');
+        $this->assertSame(200, $view->getResponse()->getStatusCode(), 'Should return 200');
```
Soullivaneuh: Why this is needed?

Controller/Api/GalleryController.php
Soullivaneuh: Forget this, I did a mistake.

Tests/Controller/Api/GalleryControllerTest.php
Soullivaneuh: ping @core23

Soullivaneuh:
Could you please also add a `conflict` key for this dependency on a separate commit?

Tests/Controller/Api/GalleryControllerTest.php
core23: Because the internal logic has changed. In the old version, the method `getResponse` was implicitly called.

Soullivaneuh: Yeah but the `View::getStatusCode` is still here. Why not keep using it?

core23:
Added a `conflict` key

Soullivaneuh:
> Added a `conflict` key

On a separated commit please. :-)

Tests/Controller/Api/GalleryControllerTest.php
core23: The method is still there, but it doesn't call the response anymore.

Btw. I don't use the REST bundle. I simple want to fix the build ;)

Soullivaneuh:
Thank you @core23

PR #1077, opened 2016-07-13
Rename array QueryParam into map to make it work with FOSRestBundle 2…

Bluestart83:

Closes #
Fixes #

## Changelog

```markdown
### Added
- Added some `Class::newMethod` to do great stuff

### Changed

### Deprecated

### Removed

### Fixed

### Security
```

## To do

- [ ] Update the tests
- [ ] Update the documentation
- [ ] Add an upgrade note

## Subject

….0, drop support 1.7 version

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -53,7 +53,7 @@
         "aws/aws-sdk-php": "^2.7",
         "jackalope/jackalope-doctrine-dbal": "^1.1",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "friendsofsymfony/rest-bundle": "^1.7.4 || ^2.0",
+        "friendsofsymfony/rest-bundle": "^2.0",
         "nelmio/api-doc-bundle": "^2.11",
         "sllh/php-cs-fixer-styleci-bridge": "^2.0"
     },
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -53,7 +53,7 @@
         "aws/aws-sdk-php": "^2.7",
         "jackalope/jackalope-doctrine-dbal": "^1.1",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
-        "friendsofsymfony/rest-bundle": "^1.7.4 || ^2.0",
+        "friendsofsymfony/rest-bundle": "^2.0",
```
Soullivaneuh: We can drop a version support like this. We have to support both.

Actually this is BC break.

Soullivaneuh:
Please first... fill the PR template.

This is not here for decoration. No more review will be done before that.

Regards.

Soullivaneuh:
BTW you created a PR with @desmax commit and **without** any template note respect.

So I will close this one and let @desmax open it if he want to propose and modify his changes.

composer.json
Bluestart83: I understand, so do two different versions Tags. Or fix it differently, but in all case your library is not compatible with 2.0 that you was targetting ;-)

Bluestart83:
Go to Hell! Stop writing crappy Library, or open your mind.

greg0ire:
Wow. Way to speak to a benevolent maintainer of a library he did not even write.
! [Dick move](http://m.imgur.com/gallery/7UOWyiK.gif)

Soullivaneuh:
> open your mind.

Well said! Now think about it.

PR #1035, opened 2016-05-31
Changed injection of container to session, added deprecated to container and added test

ahmetakbn:
Refs #1033

### Changelog

```markdown
### Added
- Added test for `Security/SessionDownloadStrategy`

### Deprecated
- `$container` property in `Security/SessionDownloadStrategy`

### Changed
- Injection of `Session` instead of the whole `Container` in `Security/SessionDownloadStrategy`
```

### Subject

See #1032 and #1033

PR #1089, opened 2016-07-29
added upgrade note for deprecated container property

ahmetakbn:
I am targetting this branch, because it is upgrade not of this branch.

refs #1035

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -9,3 +9,19 @@ UPGRADE FROM 3.0 to 3.1
 All files under the ``Tests`` directory are now correctly handled as internal test classes. 
 You can't extend them anymore, because they are only loaded when running internal tests. 
 More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
+
+### Deprecated
+
+`$container` property in `Security/SessionDownloadStrategy` is deprecated. Use `SessionInterface` `$session` instead.
+
+Before:
+
+```php
+    $downloadStrategy = new SessionDownloadStrategy($translator, $container, $times);
+```
+
+After:
+
+```php
+    $downloadStrategy = new SessionDownloadStrategy($translator, $session, $times);
+```


greg0ire:
> it is upgrade not of this brunch.

Makes me hungry ;)

core23:
We don't need this, because the property is allready marked as `@deprecated`.

greg0ire:
> We don't need this, because the property is allready marked as @deprecated.

Is this enough? We are providing deprecation messages, but for each of them, we also have an upgrade note, right?

core23:
IMO, we have enough information for developers. It is marked as `@deprecated` and we throw a `trigger_error` for this. The Upgrade note for the next major release also have short [note](https://github.com/sonata-project/SonataMediaBundle/blob/master/UPGRADE-4.0.md#deprecations).

ahmetakbn:
Was a requested by @Soullivaneuh FYI
https://github.com/sonata-project/SonataMediaBundle/pull/1035#issuecomment-234549024

And it is also written here under the Deprecated title : https://github.com/sonata-project/SonataMediaBundle/releases/tag/3.1.0

greg0ire:
Let's reopen and assign this to @Soullivaneuh , ok?

core23:
@Soullivaneuh should we add this note to the `UPGRADE-3.x.MD` file?

Soullivaneuh:
> @Soullivaneuh should we add this note to the UPGRADE-3.x.MD file?

It was always like this. All deprecation must be documented.

BTW, this should not be just be a line with which is deprecated, but a full description with example of before/after implements.

Because if we just write a line, indeed the `@deprecated` line would be enough. :wink: 

Just see the Symfony upgrades guide for inspiration, it's quite well formatted.

UPGRADE-3.x.md
```diff
@@ -9,3 +9,7 @@ UPGRADE FROM 3.0 to 3.1
 All files under the ``Tests`` directory are now correctly handled as internal test classes. 
 You can't extend them anymore, because they are only loaded when running internal tests. 
 More information can be found in the [composer docs](https://getcomposer.org/doc/04-schema.md#autoload-dev).
+
+### Deprecated
+
+`$container` property in `Security/SessionDownloadStrategy` is deprecated. Use `SessionInterface` `$session` instead.
```
Soullivaneuh: Please add example of before/after implementation. Looks symfony upgrade guides for inspirations.

ahmetakbn: is it ok like this @Soullivaneuh ?

Soullivaneuh: Looks ok but I'm not an expert of media-bundle so I don't know if the example is enough.

cc @sonata-project/contributors for review.

OskarStark: almost...

UPGRADE-3.x.md
```diff
@@ -9,3 +9,19 @@ UPGRADE FROM 3.0 to 3.1
...
+
+Before:
+
+```php
+    $downloadStrategy = new SessionDownloadStrategy($translator, $container, $times);
```
OskarStark: hmm, I am missing the typehint here, maybe its better so show the __construct() method with their typehints here to see the difference? @greg0ire

greg0ire: I think people know what `$container` and `$session` probably refer to ;)

OskarStark: ok, then lets merge as it is! 👍

OskarStark:
Thank you @ahmetakbn 👍

PR #1096, opened 2016-08-07
Fixed missing german translation

core23:
I am targetting this branch, because this adds missing lables.

## Subject

Some german translation keys were missing.

### DIFF
--- a/Resources/translations/SonataMediaBundle.de.xliff
+++ b/Resources/translations/SonataMediaBundle.de.xliff
@@ -567,6 +567,18 @@
                 <source>form.label_class</source>
                 <target>CSS Klasse</target>
             </trans-unit>
+            <trans-unit id="media_">
+                <source>Media</source>
+                <target>Medien</target>
+            </trans-unit>
+            <trans-unit id="gallery_">
+                <source>Gallery</source>
+                <target>Galerie</target>
+            </trans-unit>
+            <trans-unit id="options">
+                <source>Options</source>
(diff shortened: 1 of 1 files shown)


core23:
Ping @OskarStark

Resources/translations/SonataMediaBundle.de.xliff
```diff
@@ -567,6 +567,18 @@
...
+            <trans-unit id="media_">
+                <source>Media</source>
+                <target>Medien</target>
+            </trans-unit>
+            <trans-unit id="gallery_">
+                <source>Gallery</source>
+                <target>Galerie</target>
+            </trans-unit>
+            <trans-unit id="options">
+                <source>Options</source>
```
OskarStark: if we don't use keys there, we should do this in the next major IMO, what do you think @core23 ?

OskarStark:
Thank you @core23 👍

Resources/translations/SonataMediaBundle.de.xliff
core23: Don't know if we use some keys. We have nothing to cover this :(

Soullivaneuh:
Please @core23 do not put both patch and pedantic label. Pedantic means "no need for release".

core23:
Okay, didn't know that

greg0ire:
> Pedantic means "no need for release".

nicely said :bulb:

PR #1106, opened 2016-08-20
Use FQCNs when building mocks

greg0ire:
I am targetting this branch, because it is a critical patch.

## Changelog

No changelog, users do not care about our tests (right?)

## To do

Merge ASAP :rotating_light: 

## Subject

phpunit has no idea what the current namespace is.

This is why builds on most PRs are failing right now.


core23:
LGTM

Issue #600, opened 2014-07-23
Exception has not been caught while uploading any type of image unsupported by GD


jerryladoo:
I am currently using SonataMediaBundle with latest version specified in `dev-master` to upload image files. The current Imagine Driver is GD. But An uncaught exception has been thrown out from `doTransform` in `ImageProvider.php` before I could get any chance to validate the type of image. That is because of the following code,

```
...
try {
            // this will fail if a tiff image or others that are not supported by GD is given
            $image = $this->imagineAdapter->open($media->getBinaryContent()->getPathname());
        } catch (\RuntimeException $e) {
            $media->setProviderStatus(MediaInterface::STATUS_ERROR);
            return;
        }  // <-- only catch the RuntimeException but others flee
```

it tries to open an image file but couldn't do so as result of image unsupported by GD.

Is there any way that can allow me to catch this exception after reversed data back to modal  data from normal data? Because this process is automatically completed by symfony for transforming view data to modal data.

Any Ideas?

Issue #360, opened 2013-07-03
Show an error message when trying to upload files bigger than upload_max_filesize


laherreria:
When I try to upload a file bigger than upload_max_filesize a FileNotFoundException is thrown, so users get a 500 error without any information on what is going on and how they can fix it.

I think this should be checked and it should show an alert to the user.

This is the stacktrace

``````
Symfony\Component\HttpFoundation\File\Exception\FileNotFoundException: The file "" does not exist
    at n/a
        in /var/www/nec/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/File/MimeType/MimeTypeGuesser.php line 115

    at Symfony\Component\HttpFoundation\File\MimeType\MimeTypeGuesser->guess('')
        in /var/www/nec/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/File/File.php line 87

    at Symfony\Component\HttpFoundation\File\File->getMimeType()
        in /var/www/nec/vendor/symfony/symfony/src/Symfony/Component/HttpFoundation/File/File.php line 64

    at Symfony\Component\HttpFoundation\File\File->guessExtension()
        in /var/www/nec/app/cache/dev/classes.php line 15660

    at Sonata\MediaBundle\Provider\FileProvider->generateReferenceName(object(Media))
        in /var/www/nec/app/cache/dev/classes.php line 15610

    at Sonata\MediaBundle\Provider\FileProvider->doTransform(object(Media))
        in /var/www/nec/app/cache/dev/classes.php line 15752

    at Sonata\MediaBundle\Provider\ImageProvider->doTransform(object(Media))
        in /var/www/nec/app/cache/dev/classes.php line 15220

    at Sonata\MediaBundle\Provider\BaseProvider->transform(object(Media))
        in /var/www/nec/vendor/sonata-project/media-bundle/Sonata/MediaBundle/Form/DataTransformer/ProviderDataTransformer.php line 61

    at Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer->reverseTransform(object(Media))
        in /var/www/nec/vendor/symfony/symfony/src/Symfony/Component/Form/Form.php line 994

    at Symfony\Component\Form\Form->normToModel(object(Media))
        in /var/www/nec/vendor/symfony/symfony/src/Symfony/Component/Form/Form.php line 608

    at Symfony\Component\Form\Form->submit(object(Request))
[...]

PR #1107, opened 2016-08-27
Added upload checks

### LINKED ISSUE #593: Error Handling
Is it be possible to have an error handling when the file we send to the server is too heavy? At the moment, we have a 500 error without any error message 

In advance, I thank you

core23:

Rebase of #601 
Fixes #593 #600 #360
## Changelog

``` markdown
### Added
- Added filesize check to upload
- Added empty filename check to upload

```
## Subject

Added some checks while uploading a file.

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -23,6 +23,7 @@ use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Thumbnail\ThumbnailInterface;
 use Symfony\Component\Form\FormBuilder;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\File\Exception\UploadException;
 use Symfony\Component\HttpFoundation\File\File;
 use Symfony\Component\HttpFoundation\File\MimeType\ExtensionGuesser;
 use Symfony\Component\HttpFoundation\File\UploadedFile;
@@ -310,6 +311,13 @@ class FileProvider extends BaseProvider
             throw new \RuntimeException(sprintf('Invalid binary content type: %s', get_class($media->getBinaryContent())));
         }
 
+        if ($media->getBinaryContent() instanceof UploadedFile && $media->getBinaryContent()->getClientSize() == 0) {
+            $errorElement
+               ->with('binaryContent')
+                   ->addViolation('The file is too big, max size: '.ini_get('upload_max_filesize'))
+               ->end();
+        }
+
         if (!in_array(strtolower(pathinfo($fileName, PATHINFO_EXTENSION)), $this->allowedExtensions)) {
             $errorElement
                 ->with('binaryContent')
@@ -317,7 +325,7 @@ class FileProvider extends BaseProvider
                 ->end();
         }
 
-        if (!in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
+        if ($media->getBinaryContent()->getFilename() != '' && !in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
             $errorElement
                 ->with('binaryContent')
                     ->addViolation('Invalid mime type : %type%', array('%type%' => $media->getBinaryContent()->getMimeType()))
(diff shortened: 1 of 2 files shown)


OskarStark:
@core23 can you please check travis?

core23:
Looks like a travis problem. Can't reproduce this on my local machine

Provider/FileProvider.php
```diff
@@ -310,14 +311,21 @@ public function validate(ErrorElement $errorElement, MediaInterface $media)
             throw new \RuntimeException(sprintf('Invalid binary content type: %s', get_class($media->getBinaryContent())));
         }
 
+        if ($media->getBinaryContent() instanceof UploadedFile && $media->getBinaryContent()->getClientSize() == 0) {
```
OskarStark:
```diff
- if ($media->getBinaryContent() instanceof UploadedFile && $media->getBinaryContent()->getClientSize() == 0) {
+ if ($media->getBinaryContent() instanceof UploadedFile && 0 == $media->getBinaryContent()->getClientSize()) {

File: Provider/FileProvider.php [hunk 2]
```diff
@@ -310,14 +311,21 @@ public function validate(ErrorElement $errorElement, MediaInterface $media)
...
+
         if (!in_array(strtolower(pathinfo($fileName, PATHINFO_EXTENSION)), $this->allowedExtensions)) {
             $errorElement
                 ->with('binaryContent')
                 ->addViolation('Invalid extensions')
                 ->end();
         }
 
-        if (!in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
+        if ($media->getBinaryContent()->getFilename() != '' && !in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
```
OskarStark:
```diff
- if ($media->getBinaryContent()->getFilename() != '' && !in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
+ if ('' !=$media->getBinaryContent()->getFilename() && !in_array($media->getBinaryContent()->getMimeType(), $this->allowedMimeTypes)) {
```

Provider/FileProvider.php
```diff
@@ -379,6 +387,12 @@ protected function doTransform(MediaInterface $media)
         $this->fixBinaryContent($media);
         $this->fixFilename($media);
 
+        if ($media->getBinaryContent() instanceof UploadedFile && $media->getBinaryContent()->getClientSize() == 0) {
```
OskarStark: yoda please

Provider/FileProvider.php
```diff
@@ -379,6 +387,12 @@ protected function doTransform(MediaInterface $media)
...
+            $media->setProviderReference(sha1($media->getName().rand(11111, 99999)));
```
greg0ire: This is oddly specific… if you want a unique id, use `uniqid`, maybe?

core23:
Fixed it

core23:
Please review @OskarStark @greg0ire

OskarStark:
Thank you @core23 👍

jordisala1991:
You have to enable validation on the image field: `Assert\Valid()`. Im on my phone atm and not sure if this is documented (I believe it is), but I know 3 ways to do it:

Add validation on your entity. Using annotations or another format (xml, yaml)

There is a function on abstract admin that can do that also (search for validate on the class, or docs)

You can specify constraints on the field in the FormMapper.

lemoinem:
Thank you for your answer!

I've just tried adding `Assert\Valid` to my $file property...
I'm not using SonataAdmin, I'm using my own form, so no FormMapper or abstract admin.

I can put my Entity and my FormType in a gist if you'd like. But, as of now, it looks like this feature is simply not working.

I tried adding a `Assert\Expression` to another field, and as long as my file is small enough not to generate the exception, I see the error.

As far as I can tell, the exception is thrown before the code ever has a chance to call validate. And it makes sense since `doTransform` is called way before any attempt to validate.

On the other hand, if I comment `throw new UploadException('The uploaded file is not found');` and `$media->setContentType($media->getBinaryContent()->getMimeType());` in `doTransform`, then I see the "The file is too big, max size: 2M" message.

I think `doTransform` should be rewritten so it never throws an exception.
At least, the call to `$provider->transform` in `ProviderDataTransformer.php` should be enclosed in a try-catch, maybe logging but definitely not rethrowing any exception.
Would you like me to provide a PR for this?

jordisala1991:
Yes please, lets solve this issue. It will require some testing using sonata forms tho. Try to make ot BC on 3.x branch

Issue #1101, opened 2016-08-08
'callback constraint is not a valid callable ' in the upload 


aminealaoui:
#### Sonata packages

``` bash
$ composer show sonata-project/*
sonata-project/admin-bundle              3.4.0           The missing Symfony Admin Generator
sonata-project/block-bundle              3.1.1           Symfony SonataBlockBundle
sonata-project/cache                     1.0.7           Cache library
sonata-project/classification-bundle     3.x-dev ad97ae7 Symfony SonataClassificationBundle
sonata-project/core-bundle               3.0.3           Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.2             Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.0.2           Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.0.5           Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.1.10          Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.6.0           Lightweight Exporter library
sonata-project/intl-bundle               2.2.4           Symfony SonataIntlBundle
sonata-project/media-bundle              3.1.0           Symfony SonataMediaBundle
sonata-project/notification-bundle       3.0.0           Symfony SonataNotificationBundle

```
#### Symfony packages

``` bash
$ composer show symfony/*
symfony/monolog-bundle     2.11.1  Symfony MonologBundle
symfony/phpunit-bridge     v3.1.3  Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.2.0  Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-intl-icu  v1.2.0  Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring  v1.2.0  Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.2.0  Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php70     v1.2.0  Symfony polyfill backporting some PHP 7.0+ features to lower PHP versions
symfony/polyfill-util      v1.2.0  Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.0  Symfony Security Component - ACL (Access Control List)
symfony/swiftmailer-bundle v2.3.11 Symfony SwiftmailerBundle
[...]

PR #1115, opened 2016-09-14
Fix "callback constraint is not a valid callable" and the error of BaseMediaAdmin incorrectly retrieving providerName

binhle410:
I am targetting this branch to fix issue #1101 and this is BC.
## Changelog

``` markdown
### Changed
- invalid markup in **../Resources/config/validation.xml**
- invalid request param retrieval in **../Admin/BaseMediaAdmin.php**
```

**Change No 1** 
in **../Resources/config/validation.xml** to fix "callback constraint is not a valid callable" 
changed

```
        <constraint name="Callback">
            <option name="methods">
                <value>isStatusErroneous</value>
            </option>
        </constraint>
```

to 
`<constraint name="Callback">isStatusErroneous</constraint>`

**Change No 2**
in **../Admin/BaseMediaAdmin.php** to fix "callback constraint is not a valid callable" and the error of BaseMediaAdmin incorrectly retrieving providerName

changed
`$media->setProviderName($this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true));`
to 
`$media->setProviderName($this->getRequest()->get(sprintf('%s', $this->getUniqid()), null, true)['providerName']);`
## To do
- [ ] Update the tests

### DIFF
--- a/Admin/BaseMediaAdmin.php
+++ b/Admin/BaseMediaAdmin.php
@@ -106,7 +106,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
         if ($this->hasRequest()) {
             if ($this->getRequest()->isMethod('POST')) {
-                $media->setProviderName($this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true));
+                $media->setProviderName($this->getRequest()->get(sprintf('%s', $this->getUniqid()), null, true)['providerName']);
             } else {
                 $media->setProviderName($this->getRequest()->get('provider'));
             }
@@ -154,8 +154,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
             ->addIdentifier('name')
             ->add('description')
             ->add('enabled')
-            ->add('size')
-        ;
+            ->add('size');
     }
 
     /**
--- a/Resources/config/validation.xml
+++ b/Resources/config/validation.xml
@@ -1,11 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <constraint-mapping xmlns="http://symfony.com/schema/dic/constraint-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/constraint-mapping         http://symfony.com/schema/dic/services/constraint-mapping-1.0.xsd">
     <class name="Sonata\MediaBundle\Model\Media">
-        <constraint name="Callback">
-            <option name="methods">
-                <value>isStatusErroneous</value>
-            </option>
-        </constraint>
+        <constraint name="Callback">isStatusErroneous</constraint>
         <constraint name="Sonata\CoreBundle\Validator\Constraints\InlineConstraint">
             <option name="service">sonata.media.pool</option>
             <option name="method">validate</option>


greg0ire:
Do you need help fixing the styleci check or do you know how to proceed?

OskarStark:
@binhle410 please use a good commit msg, which can be used directly for the changelog in the PR header if you want

binhle410:
@greg0ire Please give me instructions or a link for reference since I would like to do this my self, totally new to all this CI stuff.
@OskarStark updated the PR's commit.

greg0ire:
@binhle410 you just need to click the "Details" link in the styleci check line

binhle410:
@greg0ire  how do I fix it, I didn't make any changes to any of those lines of code. The correct formatting / style is displayed on my end.

greg0ire:
I don't think it is, actually. [Download this patch](https://styleci.io/analyses/zY6Al9/diff), and apply it with `git apply wherever/you/dld/the/patch.diff`

binhle410:
@OskarStark  Do you mean that I have to change the content of an existing commit message?
How am I about to do that?

core23:
Please have a look [this](http://ohshitgit.com/#change-last-commit-message) to change your commit message.

binhle410:
@OskarStark I have added a new commit message, should this be ok?

greg0ire:
So, you should have one commit, not 6.

The thing is, when you modify something locally, you can use `git push --force`, because you mean to **overwrite** the remote branch with the new commits. What you probably did instead, is push, get a message about a conflict, pull, thus creating a merge commit, and finally pushed.

If you want to fix the solution easily, just forget all the 6 commits : `git reset --soft origin/3.x`, then  `git commit` to redo one single commit. And finally `git push --force` to replace the 6 commits on github.

binhle410:
@OskarStark  can this be considered a good commit message now ?

OskarStark:
> @OskarStark can this be considered a good commit message now ?

yes, can you try to squash your commits?

`git rebase -i HEAD~6`

afterwards replace `pick` with `f` in all commits except the first one.
Afterwards you should only have one commit, run `git commit --amend` and apply your commit message again.

Finally run `git push --force`

binhle410:
@OskarStark done

OskarStark:
Thank you @binhle410 👍  good job, keep on contributing please 🍺

greg0ire:
@OskarStark what about "Needs: Tests"?

OskarStark:
> @OskarStark what about "Needs: Tests"?

Good catch @greg0ire , any idea how to test this?

greg0ire:
I don't think a unit test would be interesting in that case. If we have end to end test it would be nice to create one, but I'm not sure if we have some. Please check and if not, I think this should be merged without any tests.

greg0ire:
Also, two errors were fixed here, there should be 2 commits not one. Or even 3 for the style fix

@binhle410 please do this : 

1. forget your commit `git reset 'HEAD^'`
2. add the change corresponding to commit 1 : `git add -p`
3. put a nice commit message
4. same as step 2 and 3
5. same as step 2 and 3
6. you should now have 3 commits `git push -f`

Admin/BaseMediaAdmin.php
```diff
@@ -154,8 +154,7 @@ protected function configureListFields(ListMapper $listMapper)
             ->addIdentifier('name')
             ->add('description')
             ->add('enabled')
-            ->add('size')
-        ;
+            ->add('size');
```
greg0ire: Should be in a separate commit. `Fix coding style`

Admin/BaseMediaAdmin.php
```diff
@@ -106,7 +106,7 @@ public function getNewInstance()
 
         if ($this->hasRequest()) {
             if ($this->getRequest()->isMethod('POST')) {
-                $media->setProviderName($this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true));
+                $media->setProviderName($this->getRequest()->get(sprintf('%s', $this->getUniqid()), null, true)['providerName']);
```
greg0ire: Should be in a separate commit

The commits message should explain what was wrong and how you fixed it. I don't understand what you fixed, to me if `sdlfkjsdf[providerName]` was missing, it defaults to null. Now with your change it defaults to null if `sdlfkjsdf` does not exist **but** you still use ['providerName'] on null. BTW, the [third parameter does not exist in sf 3](https://github.com/symfony/symfony/blob/3.1/src/Symfony/Component/HttpFoundation/Request.php#L717), [but exists in version 2](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/HttpFoundation/Request.php#L724) so maybe this isn't the right fix?

Maybe the right fix is : don't touch this line for sf/http-kernel < 3, and do something else for sf/http-kernel > 3

Resources/config/validation.xml
```diff
@@ -1,11 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <constraint-mapping xmlns="http://symfony.com/schema/dic/constraint-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://symfony.com/schema/dic/constraint-mapping         http://symfony.com/schema/dic/services/constraint-mapping-1.0.xsd">
     <class name="Sonata\MediaBundle\Model\Media">
-        <constraint name="Callback">
-            <option name="methods">
-                <value>isStatusErroneous</value>
-            </option>
-        </constraint>
+        <constraint name="Callback">isStatusErroneous</constraint>
```
greg0ire: Should be in a separate commit

Admin/BaseMediaAdmin.php
greg0ire: BTW, maybe it would be worth unit testing this after all, b/c the `null['providerName']` does not look nice.

greg0ire: You can use `version_compare` and [this constant](https://github.com/symfony/http-kernel/blob/2.8/Kernel.php#L62) to achieve that. @sonata-project/contributors : since the constant is part of http-kernel, it's ok.

greg0ire: Or use `Kernel::VERSION_ID` and a simple comparison, it should be easier.

binhle410: 1. ```$this->getUniqid()``` is supposed to be set and as I have seen from the code, this was handled properly unless this is a hacking attempt and the admin intentionally leaves out the "token". So, the "token" sdlfkjsdf should always exist.
2. If we leave it as ```$this->getRequest()->get(sdlfkjsdf[providerName])``` then the result will always null since the get method ultimately will be called upon ParameterBag which uses in_array ``` return array_key_exists($key, $this->parameters) ? $this->parameters[$key] : $default;```
so in this case $this->parameters is an array that looks like
```
['sdlfkjsdf ' => 
                  ['providerName' => 'youtube',
                  'fileName-or-something-from-the-admin-form'=>'some-value' ]
'query_param1' => 'value1',
'query_param2' => 'value2'
 ]```
 so, there is no such key as sdlfkjsdf['providerName'] it's not how in_array works.

Not sure if a unit test for this is really necessary.

Comment (greg0ire) [inline Admin/BaseMediaAdmin.php]: 2. You're probably right about this for sf/http-kernel version 3, but what about version 2 ?

Comment (greg0ire) [inline Admin/BaseMediaAdmin.php]: Here is the thing: this bundle says it is compatible with both versions 2 AND 3.

Your final code should look like this : 

```php
if (Kernel::VERSION_ID <= 20800) {
    // NEXT_MAJOR remove this block
    $media->setProviderName($this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true));
} else {
    $media->setProviderName($this->getRequest()->get($this->getUniqid(), null, true)['providerName'])
}
```

greg0ire: As per the "hacking attempt", I think we should guard this with 
```php
if (!$this->getRequest()->has($this->getUniqId())) {
    throw new \LogicException('form not found in the request');
}
```

And this alone would warrant a unit test IMO. What do you think @sonata-project/contributors

binhle410: I think you meant http-foundation since ```$this->getRequest()``` returns a ParameterBag object which belongs to httpfoundation.
2.8: https://github.com/symfony/http-foundation/blob/2.8/ParameterBag.php#L91
3.1: https://github.com/symfony/http-foundation/blob/3.1/ParameterBag.php#L86
I believe the parameter of deep = true in this case already triggers an error if the app is using http-foundation <=2.8. 

Then Why don't we just remove them all and the problem will be solved without extra checks ?

greg0ire: Argh, you're right about http-foundation vs http-kernel! Gotta find another way to detect the version…

> I believe the parameter of deep = true in this case already triggers an error if the app is using http-foundation <=2.8. 

I don't understand why. The parameter is present in <= 2.8, why would it trigger an error.

> Then Why don't we just remove them all and the problem will be solved without extra checks ?

If you find a way that works with both 2.7, 2.8, 3.0 without any deprecation message being thrown be my guest.

greg0ire: You can maybe try detecting [this method](https://github.com/symfony/http-foundation/compare/2.7...2.8#diff-313a1e2d3e8584eebf87bb466619504eL211) instead, [it should work](https://3v4l.org/JXrBm)

greg0ire: > I believe the parameter of deep = true in this case already triggers an error if the app is using http-foundation <=2.8. 

If you can test that and make sure the code is already bad for sf < 2.8, by all means solve this globally. A unit test will help a lot with proving that, because we run tests for different versions of Symfony on Travis.

binhle410: ok, I'll have to find my time for this.

greg0ire: Thanks a lot!

idchlife:
So, was issue resolved? Still getting

    {"methods":["isStatusErroneous"]} targeted by Callback constraint is not a valid callable

sf3, php7

binhle410:
no, the PR has not yet been merged, you can read what I have fixed in the PR and apply the fix in your custom repo.

elmpp:
Hi guys is there anything i can do to push this through?

Admin/BaseMediaAdmin.php
greg0ire: @elmpp : if you want to carry this PR you should solve this

PR #1086, opened 2016-07-22
Remove dependency of classification-bundle

core23:
I am targetting this branch, because it removes existing functionality.

Rebase of #794
Closes #783 #684

## Changelog

```markdown
### Added
- Added option to disable classification

### Changed
 - made `sonata-project/classification-bundle` dependency optional
```

## Subject

This is a rebase of https://github.com/sonata-project/SonataMediaBundle/pull/794, because there was no further interaction. 

This PR made the classifcation dependency optional.

PR #1120, opened 2016-09-22
Fixed wrong phpdoc of #1086

core23:
I am targetting this branch, because this is just a fix.

Fixes https://github.com/sonata-project/SonataMediaBundle/pull/1086

## Subject

Fixes wrong phpdoc / interface method introduced in https://github.com/sonata-project/SonataMediaBundle/pull/1086

### DIFF
--- a/Model/CategoryManagerInterface.php
+++ b/Model/CategoryManagerInterface.php
@@ -27,11 +27,11 @@ interface CategoryManagerInterface
     public function getRootCategory($context);
 
     /**
-     * @param bool|true $loadChildren
+     * @param bool $loadChildren
      *
      * @return CategoryInterface[]
      */
-    public function getRootCategories($loadChildren);
+    public function getRootCategories($loadChildren = true);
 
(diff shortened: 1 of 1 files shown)


Model/CategoryManagerInterface.php
```diff
@@ -27,11 +27,11 @@
     public function getRootCategory($context);
 
     /**
-     * @param bool|true $loadChildren
+     * @param bool $loadChildren
      *
      * @return CategoryInterface[]
      */
-    public function getRootCategories($loadChildren);
+    public function getRootCategories($loadChildren = true);
```
OskarStark: this is a BC break, we should add an UPGRADE note

core23: No break, because this methods was introduced one commit ago ;) 

Btw. we are talking about the master branch.

OskarStark: you are right 👍

PR #1123, opened 2016-09-26
add listener for console.terminate to allow publishing events on comm…

mvhirsch:
I am targetting this branch, because we should really generate thumbnails on console commands.
## Changelog

``` markdown
### Added
- Generate thumbnails asynchronously if creating Media on console commands.
```
## To do

- [x] Update the tests
- [x] Add an upgrade note
## Subject

I just found a missing event listener while generating thumbnails asynchronously (via SonataNotificationBundle) on console commands.
On web everything works fine, but while generating media using commands any notification will be sent to the messaging queue.

### DIFF
--- a/Tests/Thumbnail/ConsumerThumbnailTest.php
+++ b/Tests/Thumbnail/ConsumerThumbnailTest.php
@@ -12,8 +12,6 @@
 namespace Sonata\MediaBundle\Tests\Thumbnail;
 
 use Sonata\MediaBundle\Thumbnail\ConsumerThumbnail;
-use Symfony\Component\Console\ConsoleEvents;
-use Symfony\Component\HttpKernel\KernelEvents;
 
 class ConsumerThumbnailTest extends \PHPUnit_Framework_TestCase
 {
@@ -27,11 +25,11 @@ class ConsumerThumbnailTest extends \PHPUnit_Framework_TestCase
         $dispatcher = $this->getMock('Symfony\Component\EventDispatcher\EventDispatcherInterface');
         $dispatcher->expects($this->at(0))
             ->method('addListener')
-            ->with($this->equalTo(KernelEvents::FINISH_REQUEST), $this->anything());
+            ->with($this->equalTo('kernel.finish_request'), $this->anything());
 
         $dispatcher->expects($this->at(1))
             ->method('addListener')
-            ->with($this->equalTo(ConsoleEvents::TERMINATE), $this->anything());
+            ->with($this->equalTo('console.terminate'), $this->anything());
 
         $consumer = new ConsumerThumbnail('foo', $thumbnail, $backend, $dispatcher);
         $consumer->generate($provider, $media);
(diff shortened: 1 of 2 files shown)


greg0ire:
Has to do with #1042, right ?

OskarStark:
> Has to do with #1042, right ?

Yes, but we are not creating them directly if we are on the CLI; we are firing the event anyway, but based on another event.

mvhirsch:
> Has to do with #1042, right ?

It's facing the same problem, but I do not like the approach of #1042 because: there it will just generate the thumbnail on-demand instead of using the asynchronous functionality - which is far from being the way I look at the `ConsumerThumbnail`.

So instead of forcing a "directly generate thumbnails" (which the `ConsumerThumbnail` ain't for) I would like to have a working version for web *and* for cli.

greg0ire:
Asynchronous is good, I completely agree with that! On-demand leads to all kind of problems (memory or timeout problems, for instance), so thanks for changing things that way :)

OskarStark:
anything todo here?

greg0ire:
Merging

greg0ire:
Can't merge from phone

greg0ire:
Please do it @Oskarstark

OskarStark:
@mvhirsch are there any tests which need to be updated?

mvhirsch:
> @mvhirsch are there any tests which need to be updated?

There ain’t no tests at all atm. But I'll write some.

OskarStark:
> There ain’t no tests at all atm. But I'll write some.

Thank you, I added `[WIP]` in front of the PR title 👍

core23:
Any news for the tests @mvhirsch ?

mvhirsch:
Thank you for reminding me, just added a small unit test for this.

Tests/Thumbnail/ConsumerThumbnailTest.php
```diff
@@ -0,0 +1,35 @@
...
+use Sonata\MediaBundle\Provider\MediaProviderInterface;
+use Sonata\MediaBundle\Model\MediaInterface;
+use Symfony\Component\Console\ConsoleEvents;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+class ConsumerThumbnailTest extends \PHPUnit_Framework_TestCase
+{
+    public function testGenerateDispatchesEvents()
+    {
+        $thumbnail = $this->createMock(ThumbnailInterface::class);
```
greg0ire: `createMock` doesn't work with old versions of php that we are supporting… so does the `::class` notation. Please replace them with `getMock` (there will be a warning, but it should pass), and the classname as a string

mvhirsch: Sorry, just realized it after pushing :-(

greg0ire: Don't be sorry there's no harm ;)

core23:
Please have a look at travis @mvhirsch

OskarStark:
Good job @mvhirsch 👍

PR #1055, opened 2016-06-10
 medias are not filtered by category 

huziatwork:
Closes #1054 

### Changelog

```markdown
### Fixed
- Medias are not filtered by category

### Changed
- Changed datagrid value of category from object to id of category
```

PR #1130, opened 2016-10-10
medias are not filtered by category

### LINKED ISSUE #1054: medias are not filtered by category
Medias are not filtered by category

| Question | Answer |
| --- | --- |
| Bundle version | 3.x |
| Symfony version | 2.8 |
| php version | 5.5 |
# Steps to reproduce

create medias and assign different categories
# Expected results

Only medias of the current category are visible
# Actual results

All medias are visible

huziatwork:

Rebase of #1055
Fixes #1054
## Changelog

``` markdown

### Changed
- MediaAdminController: call datagrid setValue with category id instead of category object
- Template MediaAdmin:list: check if element is active by category id

### Fixed
- Medias are not filtered by category

```

### DIFF
--- a/Controller/MediaAdminController.php
+++ b/Controller/MediaAdminController.php
@@ -80,7 +80,7 @@ class MediaAdminController extends Controller
         $category = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
 
         if (!$filters) {
-            $datagrid->setValue('category', null, $category);
+            $datagrid->setValue('category', null, $category->getId());
         }
         if ($request->get('category')) {
             $categoryByContext = $this->container->get('sonata.classification.manager.category')->findOneBy(array(
@@ -89,9 +89,9 @@ class MediaAdminController extends Controller
             ));
 
             if (!empty($categoryByContext)) {
-                $datagrid->setValue('category', null, $categoryByContext);
+                $datagrid->setValue('category', null, $categoryByContext->getId());
             } else {
-                $datagrid->setValue('category', null, $category);
+                $datagrid->setValue('category', null, $category->getId());
             }
         }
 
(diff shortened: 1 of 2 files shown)


greg0ire:
Can both commits live without each other? If not, there should be only one commit.

greg0ire:
Also, don't close this PR and reopen a new one when you fix things, please use `git push --force` instead.

greg0ire:
Can you explain how/why this fix works?

huziatwork:
No, they only work together. Currenty I do have to make my commits with the web ui. So I am not able to commit changes of multiple files with one commit.

greg0ire:
Oh ok, that's not really a problem, we can squash them for you when merging.

huziatwork:
I don't know exactly, why this fix works. Obviously it makes a difference if $datagrid->setValue is called with an Id or an object. 

Sonata\AdminBundle\Datagrid calls
$this->form->submit($this->values);
$data = $this->form->getData();

Looking into the form section of the symfony profiler an error message is shown: 

**Message**
This value is not valid.
**Origin**
Cause
**value**
Symfony\Component\Validator\ConstraintViolation

Object(Symfony\Component\Form\Form).children[category].children[value] = Object(Application\Sonata\ClassificationBundle\Entity\Category)

Caused by:
Symfony\Component\Form\Exception\TransformationFailedException

Unable to reverse value for property path "[value]": Expected a string or null.

Caused by:
Symfony\Component\Form\Exception\TransformationFailedException

Expected a string or null.
|

So maybe the explanation is located in the different handling of the form class.

greg0ire:
@ahmetakbn , can you please have a look at this? Looking at da3c13f11b8dc5b16daf942a85cd3ae34dfb4697 , which you wrote, I think there might be some connection.

jordisala1991:
This can be closed, solved on #1196

PR #1131, opened 2016-10-11
Close an unclosed tag in translation file

JonathanBaudoin:
I am targetting this branch, because there's only one change, in a translation file. It's just an omitted end tag, there's no BC-break.

## Changelog

```markdown
### Fixed
- Close an unclosed XML tag in a translation file. 

## Subject

The last <target> tag, in <trans-unit id="form.label_class">,  was not closed in Resources/translations/SonataMediaBundle.sl.xliff.


greg0ire:
Thanks!

greg0ire:
How did xmllint not catch this?

PR #1144, opened 2016-11-16
Display responsive images with the ImageProvider

hug963:
I am targeting this branch, because this feature is Backwards compatible.

## Changelog

```markdown
### Added
- The Image Provider returns responsive images to the twig media helper.
```

## To do

- [x] Add unit tests
- [x] Update documentation

## Subject

The Image providers getHelperProperties() function returns srcset and sizes attributes based on the current image's available sizes.

It only returns sizes smaller than the specified format. This way it is perfectly BC and the format let us control the returned sizes.

Browsers that don't support this feature will just serve the image defined by the 'format' option.

Passing a 'sizes' option to the media tag, easily overrides the predefined sizes to tailor for current template.

I'll update the docs if you are happy with the change.

### DIFF
--- a/Admin/BaseMediaAdmin.php
+++ b/Admin/BaseMediaAdmin.php
@@ -17,6 +17,7 @@ use Sonata\AdminBundle\Form\FormMapper;
 use Sonata\CoreBundle\Model\Metadata;
 use Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer;
 use Sonata\MediaBundle\Model\CategoryManagerInterface;
+use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Sonata\MediaBundle\Provider\Pool;
 
 abstract class BaseMediaAdmin extends AbstractAdmin
@@ -150,7 +151,10 @@ abstract class BaseMediaAdmin extends AbstractAdmin
     {
         $provider = $this->pool->getProvider($object->getProviderName());
 
-        $url = $provider->generatePublicUrl($object, $provider->getFormatName($object, 'admin'));
+        $url = $provider->generatePublicUrl(
+            $object,
+            $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN)
+        );
 
         return new Metadata($object->getName(), $object->getDescription(), $url);
     }
--- a/Provider/BaseProvider.php
+++ b/Provider/BaseProvider.php
@@ -99,7 +99,8 @@ abstract class BaseProvider implements MediaProviderInterface
         if ($media->getId() && $this->requireThumbnails() && !$media->getCdnIsFlushable()) {
             $flushPaths = array();
             foreach ($this->getFormats() as $format => $settings) {
-                if ('admin' === $format || substr($format, 0, strlen($media->getContext())) === $media->getContext()) {
+                if (MediaProviderInterface::FORMAT_ADMIN === $format ||
+                    substr($format, 0, strlen($media->getContext())) === $media->getContext()) {
                     $flushPaths[] = $this->getFilesystem()->get($this->generatePrivateUrl($media, $format), true)->getKey();
                 }
             }
@@ -157,12 +158,12 @@ abstract class BaseProvider implements MediaProviderInterface
      */
     public function getFormatName(MediaInterface $media, $format)
     {
-        if ($format == 'admin') {
-            return 'admin';
+        if (MediaProviderInterface::FORMAT_ADMIN === $format) {
+            return MediaProviderInterface::FORMAT_ADMIN;
         }
 
-        if ($format == 'reference') {
-            return 'reference';
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
+            return MediaProviderInterface::FORMAT_REFERENCE;
         }
 
         $baseName = $media->getContext().'_';
--- a/Provider/BaseVideoProvider.php
+++ b/Provider/BaseVideoProvider.php
@@ -80,7 +80,7 @@ abstract class BaseVideoProvider extends BaseProvider
      */
     public function getReferenceFile(MediaInterface $media)
     {
-        $key = $this->generatePrivateUrl($media, 'reference');
+        $key = $this->generatePrivateUrl($media, MediaProviderInterface::FORMAT_REFERENCE);
 
         // the reference file is remote, get it and store it with the 'reference' format
         if ($this->getFilesystem()->has($key)) {
@@ -251,7 +251,7 @@ abstract class BaseVideoProvider extends BaseProvider
      */
     protected function getBoxHelperProperties(MediaInterface $media, $format, $options = array())
     {
-        if ($format == 'reference') {
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             return $media->getBox();
         }
 
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -221,7 +221,7 @@ class FileProvider extends BaseProvider
      */
     public function generatePublicUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             $path = $this->getReferenceImage($media);
         } else {
             // @todo: fix the asset path
@@ -248,7 +248,7 @@ class FileProvider extends BaseProvider
      */
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             return $this->getReferenceImage($media);
         }
 
@@ -271,7 +271,7 @@ class FileProvider extends BaseProvider
         }
 
         if ($mode == 'http') {
-            if ($format == 'reference') {
+            if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
                 $file = $this->getReferenceFile($media);
             } else {
                 $file = $this->getFilesystem()->get($this->generatePrivateUrl($media, $format));
--- a/Provider/ImageProvider.php
+++ b/Provider/ImageProvider.php
@@ -60,7 +60,7 @@ class ImageProvider extends FileProvider
      */
     public function getHelperProperties(MediaInterface $media, $format, $options = array())
     {
-        if ($format == 'reference') {
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             $box = $media->getBox();
         } else {
             $resizerFormat = $this->getFormat($format);
@@ -82,7 +82,7 @@ class ImageProvider extends FileProvider
             'height' => $box->getHeight(),
         );
 
-        if ($format !== 'admin') {
+        if ($format !== MediaProviderInterface::FORMAT_ADMIN) {
             $srcSet = array();
 
             foreach ($this->getFormats() as $providerFormat => $settings) {
@@ -95,7 +95,11 @@ class ImageProvider extends FileProvider
             }
 
             // The reference format is not in the formats list
-            $srcSet[] = sprintf('%s %dw', $this->generatePublicUrl($media, 'reference'), $media->getBox()->getWidth());
+            $srcSet[] = sprintf(
+                '%s %dw',
+                $this->generatePublicUrl($media, MediaProviderInterface::FORMAT_REFERENCE),
+                $media->getBox()->getWidth()
+            );
 
             $params['srcset'] = implode(', ', $srcSet);
             $params['sizes'] = sprintf('(max-width: %1$dpx) 100vw, %1$dpx', $mediaWidth);
@@ -150,7 +154,7 @@ class ImageProvider extends FileProvider
      */
     public function generatePublicUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             $path = $this->getReferenceImage($media);
         } else {
             $path = $this->thumbnail->generatePublicUrl($this, $media, $format);
--- a/Provider/ImageProvider.php
+++ b/Provider/ImageProvider.php
@@ -72,13 +72,36 @@ class ImageProvider extends FileProvider
             $box = $this->resizer->getBox($media, $resizerFormat);
         }
 
-        return array_merge(array(
+        $mediaWidth = $box->getWidth();
+
+        $params = array(
             'alt' => $media->getName(),
             'title' => $media->getName(),
             'src' => $this->generatePublicUrl($media, $format),
-            'width' => $box->getWidth(),
+            'width' => $mediaWidth,
             'height' => $box->getHeight(),
-        ), $options);
+        );
+
+        if ($format !== 'admin') {
+            $srcSet = array();
+
+            foreach ($this->getFormats() as $providerFormat => $settings) {
+                // Check if format belongs to the current media's context
+                if (strpos($providerFormat, $media->getContext()) === 0) {
+                    $width = $this->resizer->getBox($media, $settings)->getWidth();
+
+                    $srcSet[] = sprintf('%s %dw', $this->generatePublicUrl($media, $providerFormat), $width);
+                }
+            }
+
+            // The reference format is not in the formats list
+            $srcSet[] = sprintf('%s %dw', $this->generatePublicUrl($media, 'reference'), $media->getBox()->getWidth());
+
+            $params['srcset'] = implode(', ', $srcSet);
+            $params['sizes'] = sprintf('(max-width: %1$dpx) 100vw, %1$dpx', $mediaWidth);
+        }
+
+        return array_merge($params, $options);
     }
 
     /**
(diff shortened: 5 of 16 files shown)


greg0ire:
`srcset`? Did a french "and" merge with this word?

Provider/ImageProvider.php
```diff
@@ -72,12 +72,32 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
             $box = $this->resizer->getBox($media, $resizerFormat);
         }
 
+        $mediaWidth = $box->getWidth();
+
+        $srcset = array();
```
greg0ire: Please use `srcSet`, it will be clearer for dummies like me who encounter this for the first time.

Provider/ImageProvider.php
```diff
@@ -72,12 +72,32 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
...
+        $context = $media->getContext();
+
+        foreach ($this->getFormats() as $providerFormat => $settings) {
+            // Check if format belongs to the current media's context
+            if (substr($providerFormat, 0, strlen($context)) == $context) {
```
greg0ire: `strpos($providerFormat, $context) === 0` ?

Provider/ImageProvider.php
```diff
@@ -72,12 +72,32 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
...
+                $width = $this->resizer->getBox($media, $settings)->getWidth();
+
+                if ($width <= $mediaWidth) {
+                    $srcset[] = $this->generatePublicUrl($media, $providerFormat).' '.$width.'w';
+                }
+            }
+        }
+
+        $sizes = '(max-width: '.$mediaWidth.'px) 100vw, '.$mediaWidth.'px';
```
greg0ire: Too much concatenation here IMO, please consider using `sprintf` or simply string interpolation.

Provider/ImageProvider.php
```diff
@@ -72,12 +72,32 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
...
+        $srcset = array();
+        $context = $media->getContext();
+
+        foreach ($this->getFormats() as $providerFormat => $settings) {
+            // Check if format belongs to the current media's context
+            if (substr($providerFormat, 0, strlen($context)) == $context) {
+                $width = $this->resizer->getBox($media, $settings)->getWidth();
+
+                if ($width <= $mediaWidth) {
+                    $srcset[] = $this->generatePublicUrl($media, $providerFormat).' '.$width.'w';
```
greg0ire: Consider using `sprintf` here

Provider/ImageProvider.php
hug963: I stole that from there:
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Thumbnail/FormatThumbnail.php#L79-L80

greg0ire: If it appears a lot, it would be good to have a dedicated method somewhere… can you check how many occurrences of this there is?

hug963: Maybe we can leave it like that for now for consistency and add a $media->hasFormat() function using strpos in a different pull request and update the different places where it is used?

hug963: Three:
https://github.com/sonata-project/SonataMediaBundle/blob/fabd14b939267ab9ef622963f14aed1cf26caa4c/Thumbnail/FormatThumbnail.php#L80
https://github.com/sonata-project/SonataMediaBundle/blob/fabd14b939267ab9ef622963f14aed1cf26caa4c/Provider/BaseProvider.php#L102
https://github.com/sonata-project/SonataMediaBundle/blob/fabd14b939267ab9ef622963f14aed1cf26caa4c/Provider/BaseProvider.php#L169

greg0ire: I'd say do the replacement with `strpos` in a first commit, then apply your commit on top of that. Then, in a different PR, do the refactor into `hasFormat`

hug963:
@greg0ire I made the requested changes.

I had a look in the doc and there is no description anywhere of the built-in Providers.

There is no description neither of the behaviour of the "media helper" for the different providers.

So I am not sure where to fit this? 
On the "helpers" page there is a description of the File Provider behaviour with thumbnails, so maybe there?

greg0ire:
I would extract the generic parts of `reference/create_a_media_provider.rst` to a new article, and move the specific parts to a new cookbook entry. The reference should be about what exists, and the cookbook should be about extending IMO. WDYT @sonata-project/contributors ?

greg0ire:
ping @hug963 are you blocked? Do you need help?

hug963:
I'm a bit stuck.

for the tests, I'd need getBox() to return different values but it is hardcoded here:
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Tests/Provider/ImageProviderTest.php#L25
and I'm not sure how to bypass that.

For the docs I'd just add a new section "media helpers for built-in providers" to `reference/helpers.rst`.
I feel like the `reference/create_a_media_provider.rst` file is quite clear and is a complete use case.

greg0ire:
> for the tests, I'd need getBox() to return different values but it is hardcoded here

You could make the value an optional argument to the function, maybe?

```php
    public function getProvider($allowedExtensions = array(), $allowedMimeTypes = array(), $box = null)
    {
       // if null, use default value
    }
```

or create `getBox()` and `setBox()` methods to do that?

hug963:
I've started digging in for the docs to document the different media helpers, but found a big problem in the way the options are parsed.

if you want to have a media that is a video, you will accept the youtube, vimeo and daylimotion providers.

Then you pass some defaults options:
```twig

    {% media media, 'large' with {
        'html5': true,
        'player_url_parameters': {
             'cc_load_policy' : false,
         },
        'player_parameters': {
            'border': 1,
            'js_api': 1,
        },
        'class': 'video-player',
        'id': 'video-player',
    } %}
```
While the youtube provider has a 'player_url_parameters' option, the vimeo  and daylimotion providers will use the  'player_parameters' to do the same thing.
The 'class' will be used for the vimeo provider as it is hardcoded, but simply ignored for the youtube and daylimotion providers.
The id will just get ignored by all.

Another example: the vimeo provider accepts an ``allowfullscreen`` option, for Daylimotion it is ``allowFullScreen``  and for youtube ``player_url_parameters['fs']``.

If you want your media to be either a video or an image it simply crashes as all the parameters are passed through to the view and you get an 'Array to String' conversion error because of the  'player_url_parameters' and 'player_parameters' arrays.

The only way I see to resolve this would be to separate the attributes from the options and only use the options relevant to the current provider. This wouldn't be BC.

```twig

    {% media media, 'large' with {
        'options': {
            'html5': true,
            'player_url_parameters': {
                 'cc_load_policy' : false,
             },
            'player_parameters': {
                'border': 1,
            },
            'sizes' : '(min-width: 1200px) 50vw, 100vw'
        },
        'attributes: {
            'class': 'main-media',
            'id': 'main-media',
            'data-xxx': 'xxx',
        }
    } %}
```

hug963:
for the tests, I don't know how I could call getBox() several times as the $provider is actualised once.

let's say I add more formats there:
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Tests/Provider/ImageProviderTest.php#L81
```php
        $provider->addFormat('admin', array('width' => 100));
        $provider->addFormat('medium', array('width' => 500));
        $provider->addFormat('large', array('width' => 1000));
```
When we call `getHelperProperties()`, how do I make it make three different calls to getBox()?
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Tests/Provider/ImageProviderTest.php#L89

greg0ire:
You could provide an array argument to give the expected calls, and use `onConsecutiveCalls()` on the mock.

hug963:
I just made some changes to take into account the internal `reference` and `admin` formats.

I added some tests using onConsecutiveCalls() if you want to have a look.

greg0ire:
> While the youtube provider has a 'player_url_parameters' option, the vimeo and daylimotion providers will use the 'player_parameters' to do the same thing.

Not relevant to this PR, right? Maybe create an issue about that, because it is quite ugly indeed.

hug963:
I think I'll prepare a PR for the video providers, youtube and daylimotion are still embedding videos with the old deprecated flash embeds, so most of this options should be removed anyway.

The three video providers could easily extend the same template with a simple iframe and simple similar options (like allowfullscreen !) and optional url parameters, just need to think of a BC way to do this.

hug963:
Added some docs, let me know if you are happy and I'll squash the commits.

Resources/doc/reference/helpers.rst
```diff
@@ -65,7 +65,19 @@ Render the path to a ``sonata.media.provider.file`` context:
...
     {% path media, 'reference' %}
-    
+
+Media helper for images
+-----------------------
+
+The media helper for the ``sonata.media.provider.image`` provider renders a responsive image tag with sensible defaults for ``srcset`` and ``sizes``.
+The size configured will be the one used for the default fallback ``src`` as well as the largest size for the ``srcset``.
+
+To override the sizes to fit your particular design, just pass a sizes option to the helper.
```
greg0ire: Please quote `sizes`

greg0ire: (and you may squash after that)

greg0ire:
@sonata-project/contributors , please give this a final review

greg0ire:
ping @sonata-project/contributors

Provider/ImageProvider.php
```diff
@@ -72,13 +72,44 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
...
             'title' => $media->getName(),
-            'src' => $this->generatePublicUrl($media, $format),
-            'width' => $box->getWidth(),
+            'src' => $src,
+            'width' => $mediaWidth,
             'height' => $box->getHeight(),
-        ), $options);
+        );
+
+        if ($format !== 'admin') {
```
core23: Do we have a constant for this?

Provider/ImageProvider.php
```diff
@@ -72,13 +72,44 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
...
+                    $width = $this->resizer->getBox($media, $settings)->getWidth();
+
+                    if ($width <= $mediaWidth) {
+                        $srcSet[] = sprintf('%s %dw', $this->generatePublicUrl($media, $providerFormat), $width);
+                    }
+                }
+            }
+
+            // The reference format is not in the formats list
+            if ($format == 'reference') {
```
core23: Do we have a constant for this?

Provider/ImageProvider.php
hug963: I don't think so, this is consistent with what I have seen in the repository.

Provider/ImageProvider.php
OskarStark: not yet, could you provide a separate PR @core23 and fix where its needed

@rande is there a reason why we don't have a constant for this?

Provider/ImageProvider.php
core23: Then please create one for this :)

hug963:
@core23 I added constants for both hardcoded formats (admin and reference)

Provider/BaseProvider.php
```diff
@@ -157,12 +157,12 @@ public function removeThumbnails(MediaInterface $media, $formats = null)
      */
     public function getFormatName(MediaInterface $media, $format)
     {
-        if ($format == 'admin') {
-            return 'admin';
+        if ($format == MediaProviderInterface::FORMAT_ADMIN) {
```
OskarStark: please use yoda comparison here like this:

```diff
- if ($format == MediaProviderInterface::FORMAT_ADMIN) {
+ if (MediaProviderInterface::FORMAT_ADMIN == $format) {
```

Provider/BaseProvider.php
```diff
@@ -157,12 +157,12 @@ public function removeThumbnails(MediaInterface $media, $formats = null)
...
+            return MediaProviderInterface::FORMAT_ADMIN;
         }
 
-        if ($format == 'reference') {
-            return 'reference';
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: here too please

Provider/BaseVideoProvider.php
```diff
@@ -251,7 +251,7 @@ protected function getMetadata(MediaInterface $media, $url)
      */
     protected function getBoxHelperProperties(MediaInterface $media, $format, $options = array())
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Provider/FileProvider.php
```diff
@@ -221,7 +221,7 @@ public function updateMetadata(MediaInterface $media, $force = true)
      */
     public function generatePublicUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Provider/FileProvider.php
```diff
@@ -248,7 +248,7 @@ public function getHelperProperties(MediaInterface $media, $format, $options = a
      */
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Provider/FileProvider.php
```diff
@@ -271,7 +271,7 @@ public function getDownloadResponse(MediaInterface $media, $format, $mode, array
         }
 
         if ($mode == 'http') {
-            if ($format == 'reference') {
+            if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Provider/ImageProvider.php
```diff
@@ -60,7 +60,7 @@ public function getProviderMetadata()
      */
     public function getHelperProperties(MediaInterface $media, $format, $options = array())
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Provider/ImageProvider.php
```diff
@@ -127,7 +158,7 @@ public function updateMetadata(MediaInterface $media, $force = true)
      */
     public function generatePublicUrl(MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

Thumbnail/FormatThumbnail.php
```diff
@@ -34,7 +34,7 @@ public function __construct($defaultFormat)
      */
     public function generatePublicUrl(MediaProviderInterface $provider, MediaInterface $media, $format)
     {
-        if ($format == 'reference') {
+        if ($format == MediaProviderInterface::FORMAT_REFERENCE) {
```
OskarStark: same here

hug963:
@OskarStark i spotted this as well and didn't know if I should do it in this commit.

It think it should be a strict comparison everywhere.
It might create problems for people who might have referenced it incorrectly, particularly in a template:

```jinja
    {% media media, 'Reference' %}
    {% media media, 'Admin' %}
```

This is an easy mistake to make and would have worked fine for now.

OskarStark:
@greg0ire WDYT ?

greg0ire:
```
psysh
Psy Shell v0.8.3 (PHP 7.1.4-1+deb.sury.org~yakkety+1 — cli) by Justin Hileman
New version is available (current: v0.8.3, latest: v0.8.6)
>>> 'Reference' == 'reference'
=> false
```

I think you can use strict comparison right now, it's a very little BC-break

hug963:
Alright, I think it's all there!

Provider/MediaProviderInterface.php
```diff
@@ -22,6 +22,9 @@
 
 interface MediaProviderInterface
 {
+    const FORMAT_ADMIN = 'admin';
```
core23: Please add a description for the new constants

Admin/BaseMediaAdmin.php
```diff
@@ -150,7 +151,7 @@ public function getObjectMetadata($object)
     {
         $provider = $this->pool->getProvider($object->getProviderName());
 
-        $url = $provider->generatePublicUrl($object, $provider->getFormatName($object, 'admin'));
+        $url = $provider->generatePublicUrl($object, $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN));
```
OskarStark: please linebreak this, and check the other ones, too.

thank you!

hug963: What's the maximum characters per line?

```php
    $url = $provider->generatePublicUrl(
        $object, 
        $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN)
    );
```

Is that the expected format?

greg0ire: 120 chars; yes.

greg0ire:
Thanks @hug963

PR #1150, opened 2016-12-05
Documented empty `formats`

### LINKED ISSUE #1135: Make "formats" optional
Now you have to specify the "formats" for each media. Would be nice to have this section optional. For instance I want my media to be uploaded but left untouched by the system !  

Is think this is someone possible by create a custom provider but it would be nice to have the feature built in.

hpatoio:
I am targetting this branch, because this PR only change a file in the documentation so it's backward compatible.

Closes #1135

## Subject

Documented the possibility to leave `formats` empty. This PR closes #1135

### DIFF
--- a/Admin/GalleryAdmin.php
+++ b/Admin/GalleryAdmin.php
@@ -134,8 +134,9 @@ class GalleryAdmin extends AbstractAdmin
                 ))
                 ->add('enabled', null, array('required' => false))
                 ->add('name')
-                ->add('defaultFormat', $choiceType, array('choices' => $formats))
-            ->end()
+                ->ifTrue($formats)
+                    ->add('defaultFormat', $choiceType, array('choices' => $formats))
+                ->ifEnd()
             ->with('Gallery')
                 ->add('galleryHasMedias', $collectionType, array(), array(
                     'edit' => 'inline',
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -85,7 +85,6 @@ class Configuration implements ConfigurationInterface
                                 ->end()
                             ->end()
                             ->arrayNode('formats')
-                                ->isRequired()
                                 ->useAttributeAsKey('id')
                                 ->prototype('array')
                                     ->children()
--- a/Model/Gallery.php
+++ b/Model/Gallery.php
@@ -43,7 +43,7 @@ abstract class Gallery implements GalleryInterface
     /**
      * @var string
      */
-    protected $defaultFormat;
+    protected $defaultFormat = 'reference';
 
     /**
      * @var GalleryHasMediaInterface[]
--- a/Resources/config/validation.xml
+++ b/Resources/config/validation.xml
@@ -34,9 +34,5 @@
             <constraint name="NotNull"/>
             <constraint name="NotBlank"/>
         </property>
-        <property name="defaultFormat">
-            <constraint name="NotNull"/>
-            <constraint name="NotBlank"/>
-        </property>
     </class>
 </constraint-mapping>
--- a/Resources/doc/reference/troubleshooting.rst
+++ b/Resources/doc/reference/troubleshooting.rst
@@ -92,7 +92,12 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
                     - sonata.media.provider.youtube
                     - sonata.media.provider.image
                     - sonata.media.provider.file
-
+                # If you omit this section the only format generated will be the `admin` format
+                # which is needed for the backend. This is useful if you use a third party service to create
+                # thumbnails and don't need to create static one using this bundle.
+                #
+                # You can use the helper `media` with `reference` as format to access the media
+                # {% media media, 'reference' %}
                 formats:
                     small: { width: 100 , quality: 70  }
                     big:   { width: 500 , quality: 70  }
--- a/Tests/Validator/FormatValidatorTest.php
+++ b/Tests/Validator/FormatValidatorTest.php
@@ -45,13 +45,40 @@ class FormatValidatorTest extends PHPUnit_Framework_TestCase
         $validator->validate($gallery, new ValidMediaFormat());
     }
 
-    public function testValidateWithValidContext()
+    public function testValidateNotValidDefaultFormat()
+    {
+        $pool = new Pool('defaultContext');
+        $pool->addContext('test', array(), array('format1' => array()));
+
+        $gallery = $this->createMock('Sonata\MediaBundle\Model\GalleryInterface');
+        $gallery->expects($this->once())->method('getDefaultFormat')->will($this->returnValue('non_existing_format'));
+        $gallery->expects($this->once())->method('getContext')->will($this->returnValue('test'));
+
+        // Prefer the Symfony 2.5+ API if available
+        if (class_exists('Symfony\Component\Validator\Context\ExecutionContext')) {
+            $contextClass = 'Symfony\Component\Validator\Context\ExecutionContext';
+        } else {
+            $contextClass = 'Symfony\Component\Validator\ExecutionContext';
+        }
+
+        $context = $this->getMockBuilder($contextClass)
+            ->disableOriginalConstructor()
+            ->getMock();
+        $context->expects($this->once())->method('addViolation');
+
+        $validator = new FormatValidator($pool);
+        $validator->initialize($context);
+
+        $validator->validate($gallery, new ValidMediaFormat());
+    }
+
+    public function testValidateOnlyReferenceIsAllowedIfNotFormats()
     {
         $pool = new Pool('defaultContext');
         $pool->addContext('test');
 
         $gallery = $this->createMock('Sonata\MediaBundle\Model\GalleryInterface');
-        $gallery->expects($this->once())->method('getDefaultFormat')->will($this->returnValue('format1'));
+        $gallery->expects($this->once())->method('getDefaultFormat')->will($this->returnValue('format_that_is_not_reference'));
         $gallery->expects($this->once())->method('getContext')->will($this->returnValue('test'));
 
         // Prefer the Symfony 2.5+ API if available
--- a/Validator/FormatValidator.php
+++ b/Validator/FormatValidator.php
@@ -50,7 +50,10 @@ class FormatValidator extends ConstraintValidator
             }
         }
 
-        if (!array_key_exists($value->getDefaultFormat(), $formats)) {
+        $galleryDefaultFormat = $value->getDefaultFormat();
+
+        if ($galleryDefaultFormat !== 'reference'
+            && !($formats && array_key_exists($galleryDefaultFormat, $formats))) {
             $this->context->addViolation('invalid format');
         }
     }


greg0ire:
Please fill in the PR template ;)

Resources/doc/reference/troubleshooting.rst
```diff
@@ -94,6 +94,9 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
                     - sonata.media.provider.file
 
                 formats:
+                    # You can leave this list empty. If you do the only format generated will be the `admin` format which is needed for the backend. This is useful if you use a third party service to create thumbnails and don't need to create static one using this bundle.
```
greg0ire: Please use `gqq` in vim on this line. Or use your editor and make sure to keep every line under 120 chars

greg0ire:
Your commit message looks really good, good job!

greg0ire:
Please rebase your PR on 3.x

greg0ire:
It looks like you forgot step 2 :P

btribouillet:
By the way if we let the formats empty there might be some impacts.

I experienced an issue on Gallery creation.

screen shot 2016-12-06 at 11 02 57

defaultFormat on Gallery entity can't be null, tried to pass "reference"

    <field name="defaultFormat" type="string" column="default_format" length="255"/>

The quick fix i suppose would be to override the attribute from the extended entity and set the attribute nullable to true. Any non existant formats (not declared in the config) will trigger the error "invalid format" on Gallery creation.

greg0ire:
> The quick fix i suppose would be to override the attribute from the extended entity and set the attribute nullable to true.

A better fix would be to make it nullable depending on whether the "formats" configuration node is empty or not.

greg0ire:
That can be done with [this event](http://docs.doctrine-project.org/projects/doctrine-orm/en/latest/reference/events.html#load-classmetadata-event). If it's the only problem there is, maybe one of you both can contribute that? I don't think we should merge this PR until this is fixed.

hpatoio:
yes sure. Merging can wait.

Admin/GalleryAdmin.php
```diff
@@ -123,9 +123,15 @@ protected function configureFormFields(FormMapper $formMapper)
                     'translation_domain' => 'SonataMediaBundle',
                 ))
                 ->add('enabled', null, array('required' => false))
-                ->add('name')
-                ->add('defaultFormat', 'choice', array('choices' => $formats))
-            ->end()
+                ->add('name');
+
+        if ($formats) {
```
hpatoio: Field is added only if `formats` are presents

DependencyInjection/Configuration.php
```diff
@@ -85,7 +85,6 @@ private function addContextsSection(ArrayNodeDefinition $node)
                                 ->end()
                             ->end()
                             ->arrayNode('formats')
-                                ->isRequired()
```
hpatoio: `formats` is not required anymore.

Resources/config/validation.xml
```diff
@@ -36,9 +36,5 @@
             <constraint name="NotNull"/>
             <constraint name="NotBlank"/>
         </property>
-        <property name="defaultFormat">
-            <constraint name="NotNull"/>
-            <constraint name="NotBlank"/>
-        </property>
```
hpatoio: We can probably add this constraint only if some format is present. WDYT ?

Resources/doc/reference/troubleshooting.rst
```diff
@@ -94,6 +94,12 @@ Finally your settings in your ``sonata_media`` parameters will look like this:
                     - sonata.media.provider.file
 
                 formats:
```
hpatoio: I'll change this commit as well. You can now omit the `formats` key too. But let's first check the code.

Resources/config/validation.xml
greg0ire: It would be best indeed.

hpatoio:
@greg0ire I set the `defaultFormat` as always `nullable`. A user can remove the `formats` section in an already existing installation.

greg0ire:
> A user can remove the formats section in an already existing installation.

Yes but in that case, they would use the doctrine migrations bundle and make a migration, right?

hpatoio:
Well the user must run migrations each time he removes/add the `formats` section. This is not a problem if he goes from a situation where `formats` are presents to one where are not. But it can be an issue if he goes from _no formats_ to _formats_ because he will have rows where `defaultFormat` is null but the declaration do not allow it

greg0ire:
> But it can be an issue if he goes from no formats to formats because he will have rows where defaultFormat is null but the declaration do not allow it

Yes, but shouldn't he? I mean, he should populate these rows, right? Does it make sense to have null values when formats are defined?

hpatoio:
I've pushed one more commit. If a gallery has no `defaultFormat` then "reference" is returned. This ensure compatibility when a gallery has no default format and when user switch from a setup where there is no section `formats` to one where `formats` is present.

Resources/config/doctrine/BaseGallery.orm.xml
```diff
@@ -3,7 +3,7 @@
     <mapped-superclass name="Sonata\MediaBundle\Entity\BaseGallery">
         <field name="name" type="string" column="name" length="255"/>
         <field name="context" type="string" column="context" length="64"/>
-        <field name="defaultFormat" type="string" column="default_format" length="255"/>
+        <field name="defaultFormat" type="string" column="default_format" nullable="true" length="255"/>
```
greg0ire: Since your last commit, there is no need to make this nullable anymore, right?

hpatoio: We set it this field as nullable or we set `reference` as default value. In this second case we don't need the `if` in `getDefaultFormat` anymore.

greg0ire: Using `reference` as a default value seems better.

Model/Gallery.php
```diff
@@ -43,7 +43,7 @@
     /**
      * @var string
      */
-    protected $defaultFormat;
+    protected $defaultFormat = 'reference';
```
greg0ire: Souldn't it also be defined as a default with doctrine ? http://stackoverflow.com/questions/3376881/default-value-in-doctrine#3379560

greg0ire: Answering my own question : no it shouldn't, according to the docs.

Validator/FormatValidator.php
```diff
@@ -50,8 +50,16 @@ public function validate($value, Constraint $constraint)
...
-        if (!array_key_exists($value->getDefaultFormat(), $formats)) {
-            $this->context->addViolation('invalid format');
+        $galleryDefaultFormat = $value->getDefaultFormat();
+
+        if ($formats) {
+            if (!array_key_exists($galleryDefaultFormat, $formats) && $galleryDefaultFormat !== 'reference') {
+                $this->context->addViolation('invalid format');
+            }
+        } else {
+            if ($galleryDefaultFormat !== 'reference') {
```
greg0ire: `elseif` ?

hpatoio: No. I think here `if` is more clear because the flow has 2 difference path depending on the value of `format`.

Model/Gallery.php
hpatoio: Dunno know what you mean. Can you please elaborate ?

greg0ire: Don't worry, you don't have anything to do. I was suggesting changing the doctrine mapping to set this as a default on the database level, but it turns out you can't.

Validator/FormatValidator.php
greg0ire: Yes, but it also increases number of possible paths a.k.a. cyclomatic complexity. If we ignore the condition line 56 there are Only 3 possible paths with `elseif` : inside the first block, inside the second, inside none of them. with this version however, you can go inside the first block, inside the second but not inside the nested if, inside the second and inside the nested if, and inside none of them, which makes 4 possible paths. In the end, it makes it less easy to reason about. Also, I notice that you test for `$galleryDefaultFormat !== 'reference'` in both conditions. Maybe you could rewrite this like this : 


```php
if ($galleryDefaultFormat !== 'reference') {
   if (!$formats || !array_key_exists($galleryDefaultFormat, $formats)) {
       $this->context->addViolation('invalid format');
   }
}
```

Which simplifies into 

```php
if ($galleryDefaultFormat !== 'reference' 
   && !($formats && array_key_exists($galleryDefaultFormat, $formats))) {
       $this->context->addViolation('invalid format');
   }
}
```

Only 2 possible paths in the end.

hpatoio:
Do you want me to squash commits before merge ?

greg0ire:
That would be great :)

hpatoio:
squashed and merged

hpatoio:
@greg0ire just wondering what your are waiting for to merge this PR. I need this feature in a new project and I'm just wondering if connect to my fork or to the main repo.

greg0ire:
PRs are supposed to be approved by at least 2 reviewers. @sonata-project/contributors , please review.

greg0ire:
ping @sonata-project/contributors

Admin/GalleryAdmin.php
core23: Could you use `ifTrue` and `ifEnd`?

hpatoio:
@core23 pushed the changes. I'll squash and rebase once the PR is ok for you. 
Dunno why Travis fails. I think is not directly related to this PR

greg0ire:
It's probably not indeed, see https://github.com/sebastianbergmann/phpunit/issues/2446

greg0ire:
The bug has been fixed, I'm restarting the build

hpatoio:
@greg0ire  @core23 Squashed.

Admin/GalleryAdmin.php
```diff
@@ -124,8 +124,13 @@ protected function configureFormFields(FormMapper $formMapper)
                 ))
                 ->add('enabled', null, array('required' => false))
                 ->add('name')
+                ->ifTrue($format)
```
jordisala1991: I think this should be $formats. Also the indentation could be nested to be clearer.

Admin/GalleryAdmin.php
```diff
@@ -124,8 +124,13 @@ protected function configureFormFields(FormMapper $formMapper)
...
                 ->add('defaultFormat', 'choice', array('choices' => $formats))
-            ->end()
+                ->ifEnd();
+
+        $formMapper->end();
```
jordisala1991: To be consistent with other with and end's calls this should be like before, nested at the same level of 
`->with('Options')`

hpatoio: @jordisala1991 please check if is ok now.

greg0ire:
@jordisala1991 @core23 , please review again.

Tests/Validator/FormatValidatorTest.php
```diff
@@ -45,13 +45,38 @@ public function testValidate()
...
+    }
+
+    public function testValidateOnlyReferenceIsAllowedIfNotFormats()
     {
         $pool = new Pool('defaultContext');
         $pool->addContext('test');
 
-        $gallery = $this->createMock('Sonata\MediaBundle\Model\GalleryInterface');
-        $gallery->expects($this->once())->method('getDefaultFormat')->will($this->returnValue('format1'));
+        $gallery = $this->getMock('Sonata\MediaBundle\Model\GalleryInterface');
```
jordisala1991: Do not use getMock please

Tests/Validator/FormatValidatorTest.php
```diff
@@ -45,13 +45,38 @@ public function testValidate()
         $validator->validate($gallery, new ValidMediaFormat());
     }
 
-    public function testValidateWithValidContext()
+    public function testValidateNotValidDefaultFormat()
+    {
+        $pool = new Pool('defaultContext');
+        $pool->addContext('test', array(), array('format1' => array()));
+
+        $gallery = $this->getMock('Sonata\MediaBundle\Model\GalleryInterface');
```
jordisala1991: use createMock instead

Tests/Validator/FormatValidatorTest.php
```diff
@@ -45,13 +45,38 @@ public function testValidate()
...
+        $gallery->expects($this->once())->method('getContext')->will($this->returnValue('test'));
+
+        // Prefer the Symfony 2.5+ API if available
+        if (class_exists('Symfony\Component\Validator\Context\ExecutionContext')) {
+            $contextClass = 'Symfony\Component\Validator\Context\ExecutionContext';
+        } else {
+            $contextClass = 'Symfony\Component\Validator\ExecutionContext';
+        }
+
+        $context = $this->getMock($contextClass, array(), array(), '', false);
```
jordisala1991: use createMock instead

Admin/GalleryAdmin.php
```diff
@@ -134,8 +134,9 @@ protected function configureFormFields(FormMapper $formMapper)
                 ))
                 ->add('enabled', null, array('required' => false))
                 ->add('name')
-                ->add('defaultFormat', $choiceType, array('choices' => $formats))
-            ->end()
+                ->ifTrue($formats)
+                    ->add('defaultFormat', 'choice', array('choices' => $formats))
```
jordisala1991: Change `'choice'` by `$choiceType` please. To fix compatibility with SF3

Tests/Validator/FormatValidatorTest.php
```diff
@@ -45,13 +45,38 @@ public function testValidate()
...
+        $gallery->expects($this->once())->method('getContext')->will($this->returnValue('test'));
+
+        // Prefer the Symfony 2.5+ API if available
+        if (class_exists('Symfony\Component\Validator\Context\ExecutionContext')) {
+            $contextClass = 'Symfony\Component\Validator\Context\ExecutionContext';
+        } else {
+            $contextClass = 'Symfony\Component\Validator\ExecutionContext';
+        }
+
+        $context = $this->createMock($contextClass, array(), array(), '', false);
```
jordisala1991: this should be 
```
$this->getMockBuilder($contextClass)
    ->disableOriginalConstructor()
    ->getMock();
```

jordisala1991: Sorry, my mistake on the previous review

jordisala1991:
For me its 👍 when the build passes. Thanks @hpatoio!

greg0ire:
Thanks @hpatoio

Issue #1056, opened 2016-06-11
3.0 version does not work with FOSRestBundle 2.0


desmax:
| Question | Answer |
| --- | --- |
| Bundle version | 3.0.0 |
| Symfony version | v2.8.7 |
| php version | PHP 7.0.7-1~dotdeb+8.1 (cli) ( NTS ) |
# Error message

[Doctrine\Common\Annotations\AnnotationException]
  [Creation Error] The annotation @QueryParam declared on method Sonata\MediaBundle\Controller\Api\GalleryController::getGalleriesAction() does not have a property named "array". Available properties: requirements, map, allowBlank, name, key, default, description, strict, nullable, incompatibles
# Steps to reproduce

Install friendsofsymfony/rest-bundle(~2.0@dev) via composer
# Expected results

Installation finishes successfully 
# Actual results

Get previously mentioned error, on step Sensio\Bundle\DistributionBundle\Composer\ScriptHandler::installAssets

Soullivaneuh:
Your bug report does not follow the format required by the template.

Please update it before proceed.

desmax:
updated

desmax:
https://github.com/desmax/SonataMediaBundle/commit/9e4d16396d90405bb1696b5e2ba42bc602dded0a

Bluestart83:
Same problem!

Soullivaneuh:
@desmax Could you please propose a PR for discussion?

Bluestart83:
Yes desmax@9e4d163.
Thanks @desmax

Soullivaneuh:
@Bluestart83 It's a commit. I was asking about a PR. :wink:

Bluestart83:
Sorry it was my first one ^^*
https://github.com/sonata-project/SonataMediaBundle/pull/1077

Soullivaneuh:
Please don't just open PR without reading contributing guide en template.

For this case, let @desmax doing this. The modification can't be accepted as is.

PR #1155, opened 2016-12-21
Api groups moved from string to array

lemorragia:
Following [this](https://github.com/sonata-project/SonataUserBundle/pull/805) and [this](https://github.com/sonata-project/SonataClassificationBundle/pull/245). 

Closes https://github.com/sonata-project/SonataMediaBundle/issues/1056

## Changelog
```markdown
### Changed
- Moved ApiDoc groups from string to array. 

### Fixed
- Moved FosRest groups from string to array and reimplemented the orderBy parameter enabling support for FosRestBundle>=2.0.
```

## Subject

FosRestBundle>=2.0 requirest api groups declared as array and requires a new implementation of the orderBy parameter in the controller which has to be bc-compatible. This fix enable the support for both FosRest 1 and 2, following the two pull requests.

I uniformed the annotations in ApiDoc groups either. This seemed right to do because nelmio-apidoc usually expects groups as an array, and they convert it in ApiDocExtractor with this

```php
// normalize groups
if (isset($input['groups']) && is_string($input['groups'])) {
$input['groups'] = array_map('trim', explode(',', $input['groups']));
}
```
so, given the array need of fos-rest, this also seemed right to do to prevent further work ahead in case they deprecate that one. It does not break BC.

### DIFF
--- a/Controller/Api/GalleryController.php
+++ b/Controller/Api/GalleryController.php
@@ -77,15 +77,14 @@ class GalleryController
      *
      * @ApiDoc(
      *  resource=true,
-     *  output={"class"="Sonata\DatagridBundle\Pager\PagerInterface", "groups"="sonata_api_read"}
+     *  output={"class"="Sonata\DatagridBundle\Pager\PagerInterface", "groups"={"sonata_api_read"}}
      * )
      *
      * @QueryParam(name="page", requirements="\d+", default="1", description="Page for gallery list pagination")
      * @QueryParam(name="count", requirements="\d+", default="10", description="Number of galleries by page")
      * @QueryParam(name="enabled", requirements="0|1", nullable=true, strict=true, description="Enabled/Disabled galleries filter")
-     * @QueryParam(name="orderBy", array=true, requirements="ASC|DESC", nullable=true, strict=true, description="Order by array (key is field, value is direction)")
      *
-     * @View(serializerGroups="sonata_api_read", serializerEnableMaxDepthChecks=true)
+     * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
      * @param ParamFetcherInterface $paramFetcher
      *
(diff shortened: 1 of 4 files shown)


greg0ire:
Please fix the test suite

lemorragia:
I have a doubt. 
It seems that the tests fail passing on the addParam function. I do not fully understand why, my understanding of class-mocking in tests is limited.

Anyway, i looked in sonata bundles and i found an alternative way to declare the ParamFetcher instead of the ParamFetcherInterface within the tests in sonata-user. 
Is replacing
```
        $params = $this->getMock('FOS\RestBundle\Request\ParamFetcherInterface');
        $params->expects($this->once())->method('all')->will($this->returnValue(array('page' => 1, 'count' => 10, 'orderBy' => array('id' => 'ASC'))));
        $params->expects($this->exactly(3))->method('get');

        $this->assertSame(array(), $gController->getGalleriesAction($params));
```
with something like
```
        $paramFetcher = $this->getMockBuilder('FOS\RestBundle\Request\ParamFetcher')
            ->disableOriginalConstructor()
            ->getMock();
        $paramFetcher->expects($this->exactly(3))->method('get');
        $paramFetcher->expects($this->once())->method('all')->will($this->returnValue(array()));

        $this->assertSame(array(), $gController->getGalleriesAction($paramFetcher));
```
an acceptable solution for both classes? In some way i understand that this solution may be not testing the same thing...

greg0ire:
I think what you did is exactly right: we are calling `addParam`, which is not part of the interface, so we'd better have a `ParamFetcher` instance, it will not work with any implementation.

lemorragia:
Ok, i kinda believe i fixed the tests :+1:

greg0ire:
Please fix StyleCI though

lemorragia:
You meant the exception namespace?

greg0ire:
StyleCI is green now so yeah, I guess that's what I meant ;)

greg0ire:
Thanks @lemorragia !

PR #1156, opened 2016-12-29
Set category id as value for category filter

### LINKED ISSUE #1054: medias are not filtered by category

rbs-asimon:
I am targetting this branch, because the commit fixes the original intention of MediaAdminController code : be able to filter the media list by category.

Closes #1054

## Changelog

```markdown
### Fixed
- fix the handling of `category` query parameter in MediaAdminController that prevents medias to be filtered by category
```

## Subject

MediaAdminController was modified to use category id as datagrid value instead of the related entity. The associated twig template was also modified to match the new nature of the category datagrid value.

### DIFF
--- a/Controller/MediaAdminController.php
+++ b/Controller/MediaAdminController.php
@@ -77,9 +77,8 @@ class MediaAdminController extends Controller
         $datagrid->setValue('context', null, $context);
 
         // retrieve the main category for the tree view
-        $rootCategory = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
-        // This should be safe as the root category has to exist for a given context but I do not like fatal errors
-        $rootCategoryId = !empty($rootCategory) ? $rootCategory->getId() : null;
+        $rootCategory = $this->container->get('sonata.classification.manager.category')->getRootCategory($context); 
+        $rootCategoryId = $rootCategory->getId(); // This is safe
 
         if (!$filters) {
             $datagrid->setValue('category', null, $rootCategoryId);
--- a/Controller/MediaAdminController.php
+++ b/Controller/MediaAdminController.php
@@ -77,10 +77,12 @@ class MediaAdminController extends Controller
         $datagrid->setValue('context', null, $context);
 
         // retrieve the main category for the tree view
-        $category = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
+        $rootCategory = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
+        // This should be safe as the root category has to exist for a given context but I do not like fatal errors
+        $rootCategoryId = !empty($rootCategory) ? $rootCategory->getId() : null;
 
         if (!$filters) {
-            $datagrid->setValue('category', null, $category);
+            $datagrid->setValue('category', null, $rootCategoryId);
         }
         if ($request->get('category')) {
             $categoryByContext = $this->container->get('sonata.classification.manager.category')->findOneBy(array(
@@ -89,9 +91,9 @@ class MediaAdminController extends Controller
             ));
 
             if (!empty($categoryByContext)) {
-                $datagrid->setValue('category', null, $categoryByContext);
+                $datagrid->setValue('category', null, $categoryByContext->getId());
             } else {
-                $datagrid->setValue('category', null, $category);
+                $datagrid->setValue('category', null, $rootCategoryId);
             }
         }
 
@@ -104,7 +106,7 @@ class MediaAdminController extends Controller
             'action' => 'list',
             'form' => $formView,
             'datagrid' => $datagrid,
-            'root_category' => $category,
+            'root_category' => $rootCategory,
             'csrf_token' => $this->getCsrfToken('sonata.batch'),
         ));
     }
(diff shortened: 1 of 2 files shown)


Controller/MediaAdminController.php
```diff
@@ -77,10 +77,12 @@ public function listAction(Request $request = null)
         $datagrid->setValue('context', null, $context);
 
         // retrieve the main category for the tree view
-        $category = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
+        $rootCategory = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
+        // This should be safe as the root category has to exist for a given context but I do not like fatal errors
```
greg0ire: Why not throw an exception instead then?

rbs-asimon: To preserve the original behaviour, nothing more.

rbs-asimon: If I had to do it, I would do it on the CategoryManager level. But the actual code assumes the entity exists : 

```php
namespace Sonata\ClassificationBundle\Entity;

class CategoryManager
{
    /**
     * @param ContextInterface $context
     *
     * @return CategoryInterface
     */
    public function getRootCategory($context = null)
    {
        $context = $this->getContext($context);

        $this->loadCategories($context);

        return $this->categories[$context->getId()][0]; // Here we assume the entity always exists, with no verification at all
    }
```

greg0ire: Consider triggering a depreciation notice announcing an exception for next major if you think an exception would have been the right thing to do. See CONTRIBUTING.md

rbs-asimon: I agree but seriously, it took me 1 hour to understand and fix the code (I never read before this part of the sonata bundles) and it takes me more to fit the "administrative part" of the process.

The code is fixed and the original behaviour is preserved. I think this is sufficient for maintenance release. Maybe someone else could do the other part ? It would be great, thank you.

rbs-asimon: OK. I just read a little bit more `Sonata\ClassificationBundle\Entity\CategoryManager` : in fact, `CategoryManager::loadCategories()` creates the missing entity if needed ; I missed this part of the code, my mistake.

I generally do not like to create on the fly objects that have to be, but it is the way it is actually done so my "non null" check is not necessary anymore and the exception throwing is not too.

greg0ire: OK great!

Miljar:
Is there anything more that needs to be done for this PR to be merged?

greg0ire:
Yes:

- fixing style CI
- getting another approval from @sonata-project/contributors

Controller/MediaAdminController.php
```diff
@@ -77,10 +77,11 @@ public function listAction(Request $request = null)
         $datagrid->setValue('context', null, $context);
 
         // retrieve the main category for the tree view
-        $category = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
+        $rootCategory = $this->container->get('sonata.classification.manager.category')->getRootCategory($context);
```
pix-art: I think the failure comes from to long lines. try this:

$categoryManager = $this->container->get('sonata.classification.manager.category');
$rootCategory = $categoryManager->getRootCategory($context); 

...

$categoryByContext = $categoryManager->findOneBy(ar...

greg0ire: The failure comes from a trailing space.

pix-art: Than it's probably this one: 
`--- a/test.php
+++ b/test.php
@@ -109,4 +109,4 @@ class MediaAdminController extends Controller
             'csrf_token' => $this->getCsrfToken('sonata.batch'),
         ));
     }
-}
\ No newline at end of file
+}
`

greg0ire: - no it isn't :P
- just click the "Details" link next to StyleCI and you will no which it is ;)

pix-art: Hmm ^^ This is confusing, details show me nothing, except the line that changed. so i've been trying to figure out where the issue comes from.

greg0ire: In that sort of case, highlighting the lines and you will see trailing spaces ;)

pix-art: I might look a bit retarded now but i still don't see it :P 
screenshot 2017-01-16 15 12 57

greg0ire: In fact, you don't even need to highlight to see it : 

sans titre

pix-art: Aha, i see it now, that's a nice bug for StyleCI :-) In chrome the line is cut off to early, so you dont see the issue. Thanks for pointing that out ^^

pix-art:
ping @rbs-asimon Can you fix that line? ^^ That way we can get this merged.

PR #1157, opened 2016-12-29
Fixed ACL configuration not passed to S3 Filesystem

stloyd:
I am targetting this branch, because we use version 3.3 & we cannot upload some files correctly on S3.

## Changelog

```markdown
### Fixed
- Fixed ACL configuration not passed to S3 Filesystem
```

## To do

- [x] Update the tests

## Subject

In code it uses lowercase `acl` which here it was `ACL`, which lead to always setting `private`.
Also after that change it got error about wrong `acl` due too fact it was not "translated" too Amazon SDK format.

### DIFF
--- a/DependencyInjection/SonataMediaExtension.php
+++ b/DependencyInjection/SonataMediaExtension.php
@@ -12,6 +12,7 @@
 namespace Sonata\MediaBundle\DependencyInjection;
 
 use Sonata\EasyExtendsBundle\Mapper\DoctrineCollector;
+use Sonata\MediaBundle\Metadata\AmazonMetadataBuilder;
 use Symfony\Component\Config\Definition\Processor;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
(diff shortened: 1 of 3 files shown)


hpatoio:
Shouldn't you change also line#76 in AmazonMetadataBuilder.php ?
https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Metadata/AmazonMetadataBuilder.php#L76

Otherwise you "covert" the code twice.

hpatoio:
You can also remove the array at line 38 in AmazonMetadataBuilder.php  https://github.com/stloyd/SonataMediaBundle/blob/40f29865097bc28002e660fb72e2db846f3cb28d/Metadata/AmazonMetadataBuilder.php#L38

stloyd:
@hpatoio It would be BC break as method is `protected` thus why I left it there.

hpatoio:
@stloyd #wise

core23:
Ping @stloyd

stloyd:
@greg0ire @core23 Finally found some time to fix everything related to this PR. Please review.

jordisala1991:
Could you add some test for this on SonataMediaExtension?

DependencyInjection/SonataMediaExtension.php
```diff
@@ -387,33 +388,58 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
 
         // add the default configuration for the S3 filesystem
         if ($container->hasDefinition('sonata.media.adapter.filesystem.s3') && isset($config['filesystem']['s3'])) {
+            $acl = str_replace(
```
jordisala1991: Why don't we use directly the correct value on our config instead of having to replace it here?

I mean, maybe for this PR we can do it in a BC way. Put a message here to remove this replaces, modify our documentation to show our new params, and maybe even trigger a deprecation message if someone use them?

WDYT @stloyd @sonata-project/contributors ?

greg0ire: Not sure why it's like this, uses the correct values directly sounds good

core23:

PR #1165, opened 2017-01-08
Drop support for old versions of Symfony

greg0ire:
I am targetting this branch, because dropping old versions of Symfony is major.

Refs https://github.com/sonata-project/dev-kit/issues/216

## Changelog

```markdown
### Removed
- Support for Symfony 2.3 through 2.7 was dropped.
```

## Subject

Drop old Symfony versions

PR #1166, opened 2017-01-11
fix empty argument in RolesDownloadStrategy

jenyak:
I am targetting this branch, because this branch an error which was introduced in #1165

## Changelog

```markdown
### Fixed
- Using the `SecurityContext` is not possible with Symfony >=2.8
```

## Subject

```
[InvalidArgumentException]                                                                                                                                                    
  Argument 2 should be an instance of Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface or Symfony\Component\Security\Core\SecurityContextInterface  
```

Fix error after merge #1165

### DIFF
--- a/Resources/config/security.xml
+++ b/Resources/config/security.xml
@@ -9,8 +9,7 @@
         </service>
         <service id="sonata.media.security.superadmin_strategy" class="Sonata\MediaBundle\Security\RolesDownloadStrategy">
             <argument type="service" id="translator"/>
-            <argument/>
-            <!-- Either security.context or security.authorization_checker -->
+            <argument type="service" id="security.authorization_checker"/>
             <argument type="collection">
                 <argument>ROLE_SUPER_ADMIN</argument>
                 <argument>ROLE_ADMIN</argument>
@@ -23,8 +22,7 @@
         </service>
         <service id="sonata.media.security.connected_strategy" class="Sonata\MediaBundle\Security\RolesDownloadStrategy">
             <argument type="service" id="translator"/>
-            <argument/>
-            <!-- Either security.context or security.authorization_checker -->
+            <argument type="service" id="security.authorization_checker"/>
             <argument type="collection">
                 <argument>IS_AUTHENTICATED_FULLY</argument>
                 <argument>IS_AUTHENTICATED_REMEMBERED</argument>
--- a/Security/RolesDownloadStrategy.php
+++ b/Security/RolesDownloadStrategy.php
@@ -15,7 +15,6 @@ use Sonata\MediaBundle\Model\MediaInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 use Symfony\Component\Security\Core\Exception\AuthenticationCredentialsNotFoundException;
-use Symfony\Component\Security\Core\SecurityContextInterface;
 use Symfony\Component\Translation\TranslatorInterface;
 
 class RolesDownloadStrategy implements DownloadStrategyInterface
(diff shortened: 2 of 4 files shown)


Resources/config/security.xml
```diff
@@ -9,8 +9,7 @@
         </service>
         <service id="sonata.media.security.superadmin_strategy" class="Sonata\MediaBundle\Security\RolesDownloadStrategy">
             <argument type="service" id="translator"/>
-            <argument/>
-            <!-- Either security.context or security.authorization_checker -->
+            <argument type="service" id="security.authorization_checker"/>
```
greg0ire: This looks wrong. I think this empty argument is supposed to be replaced in `SonataMediaExtension`

OskarStark: hmm, as this is master and we already dropped the support, SF 2.8 is the lowest version, so replacing the argument is not needed anymore IMO

The fix of @jenyak looks correct to me @greg0ire

greg0ire: Oh that's right indeed!

greg0ire: The authorization checker service is not found though :/

OskarStark: if the empty argument is replaced in the compiler pass or the extension, we should remove that code, too

can you please check this @jenyak

greg0ire:
> Merge remote-tracking branch 'origin/fix-empty-argument' into fix-emp… 

It looks like you need to force push

jenyak:
@greg0ire like this ?

greg0ire:
Yup. Looks better!

Security/RolesDownloadStrategy.php
```diff
@@ -37,13 +36,13 @@ class RolesDownloadStrategy implements DownloadStrategyInterface
...
      * @param TranslatorInterface                                    $translator
-     * @param AuthorizationCheckerInterface|SecurityContextInterface $security
+     * @param AuthorizationCheckerInterface $security
      * @param string[]                                               $roles
      */
     public function __construct(TranslatorInterface $translator, $security, array $roles = array())
     {
-        if (!$security instanceof AuthorizationCheckerInterface && !$security instanceof SecurityContextInterface) {
-            throw new \InvalidArgumentException('Argument 2 should be an instance of Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface or Symfony\Component\Security\Core\SecurityContextInterface');
+        if (!$security instanceof AuthorizationCheckerInterface) {
```
OskarStark: this IF should be removed completely

Security/RolesDownloadStrategy.php
```diff
@@ -37,13 +36,13 @@ class RolesDownloadStrategy implements DownloadStrategyInterface
 
     /**
      * @param TranslatorInterface                                    $translator
-     * @param AuthorizationCheckerInterface|SecurityContextInterface $security
+     * @param AuthorizationCheckerInterface $security
      * @param string[]                                               $roles
      */
     public function __construct(TranslatorInterface $translator, $security, array $roles = array())
```
OskarStark: please add a typehint for AuthorizationChecker in front of `$security`

OskarStark:
Please check StyleCI @jenyak

Resources/config/security.xml
OskarStark: the class `SecurityContextCompilerPass` could be removed IMO

jenyak: @OskarStark [this class was removed](https://github.com/sonata-project/SonataMediaBundle/pull/1165/files#diff-84cf3aa70e3d2b787e0810bf807f4406)

OskarStark: ah and this introduced the error 😄

PR #1177, opened 2017-01-21
[BUGFIX] Fixed issue when filtering by category from classification

TheCelavi:
[BUGFIX] Fixed issue when filtering by category from classification bundle.

Prerequisite: You have Sonata Sandbox installed, there is a SonataClassificationBundle, there are some Categories already inserted, lets say in "default" context

Steps necessary to reproduce:

Go to: [YOUR_APP_DOMAIN]/app_dev.php/admin/app/media-media/list
You will be displayed with a list of medias on right and categories in tree view on the left
Click on any category
Expected result: displayed media within selected category
Actual result: See Symfony debug toolbar, section [Forms], go to details, see "category" form field, expand it, see "value" field -> click on it.

You will get the following:

"Vrednost je nevalidna" (It means invalid value - translated).	value
Symfony\Component\Validator\ConstraintViolation
Object(Symfony\Component\Form\Form).children[category].children[value] = Object(AppBundle\Entity\Classification\Category)
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
Unable to reverse value for property path "[value]": Expected a string or null.
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
Expected a string or null.

By stating that 'category' filter field is type of 'doctrine_orm_choice' -> it expects Entity as well.

I am targetting this branch, because: all tests passes.

Closes: No related issue

```markdown
### Fixed
See diff
```

## To do
- N/A
## Subject
Changed form type of classification field in filters builder in MediaAdmin

### DIFF
--- a/Admin/ORM/MediaAdmin.php
+++ b/Admin/ORM/MediaAdmin.php
@@ -36,7 +36,7 @@ class MediaAdmin extends Admin
             ->add('context', null, array(
                 'show_filter' => $this->getPersistentParameter('hide_context') !== true,
             ), 'choice', $options)
-            ->add('category', null, array(
+            ->add('category', 'doctrine_orm_choice', array(
                 'show_filter' => false,
             ))
             ->add('width')


greg0ire:
When building the CHANGELOG.md for the next release, we are going to go through all PRs. "See diff" is not something we can add in the changelog.

TheCelavi:
Bug description: you can not filter media according to categories in Sonata Sandbox

Precondition: You have Sonata Sandbox installed, there is a SonataClassificationBundle, there are some Categories already inserted, lets say in "default" context

Steps necessary to reproduce:

1. Go to: [YOUR_APP_DOMAIN]/app_dev.php/admin/app/media-media/list 
2. You will be displayed with a list of medias on right and categories in tree view on the left
3. Click on any category

Expected result: displayed media within selected category
Actual result: See Symfony debug toolbar, section [Forms], go to details, see "category" form field, expand it, see "value" field -> click on it.

You will get the following:

"Vrednost je nevalidna" (It means invalid value - translated).	value	
Symfony\Component\Validator\ConstraintViolation
Object(Symfony\Component\Form\Form).children[category].children[value] = Object(AppBundle\Entity\Classification\Category)
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
Unable to reverse value for property path "[value]": Expected a string or null.
Caused by:
Symfony\Component\Form\Exception\TransformationFailedException
Expected a string or null.

By stating that 'category' filter field is type of 'doctrine_orm_choice' -> it expects Entity as well.

Is that sufficient to demystify what is wrong?

Admin/ORM/MediaAdmin.php
```diff
@@ -36,7 +36,7 @@ protected function configureDatagridFilters(DatagridMapper $datagridMapper)
             ->add('context', null, array(
                 'show_filter' => $this->getPersistentParameter('hide_context') !== true,
             ), 'choice', $options)
-            ->add('category', null, array(
+            ->add('category', 'doctrine_orm_choice', array(
```
greg0ire: What if the orm is not in use? I had a quick look at the bundle, it seems you can choose to use categories as documents. Also? Why did it work before? I think it worked because the type guessers translated `null` to `doctrine_orm_choice` or something like that if you have the orm installed, am I right? If yes, maybe the right solution would be to get the guessers to work again.

TheCelavi: I dont know, I am not dedicated Sonata user, just started to learning damn thing. I don't know how it did worked before, now it does not :)

If you point me what to checkout and debug, I can take a look and try to fix it. Or you can do it, I don't mind -> issue exists, I think that you can reproduce it.

greg0ire: I think your best chance at solving the bug is starting [here](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/Guesser/TypeGuesser.php#L26). The `null` should be converted in this method IMO.

> I don't know how it did worked before, now it does not :)

If you look at the history of the file, `null` was always here, implicitly or by default.

> Or you can do it, I don't mind -> issue exists, I think that you can reproduce it.

Sorry but I don't even use these bundles so that would be way too cumbersome for me, leaving aside the fact that I have zero benefit in solving this bug.

TheCelavi: Ok, thanks, let's see than what went wrong and why it does not work...

BTW, I just went to try out demo app: http://rftsgb.demo.sonata-project.org/admin/app/media-media/list?category=19

You can reproduce bug there as well.

greg0ire: Nice touch with the demo, and good luck finding the bug :)

greg0ire: Actually since the problem is with a filter (isn't it), the relevant code is probably [here](https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/Guesser/FilterTypeGuesser.php#L26)

PR #1188, opened 2017-02-05
Refactor on media edit image size and measures, SonataIntlBundle is not needed anymore

jordisala1991:
I am targetting this branch, because I think there is no BC break.

## Changelog

```markdown

### Changed
- Media edit size calculations

### Removed
- SonataIntlBundle dependency
```

## To do

- [x] Update the documentation

## Subject

SonataIntlBundle was only used to render sizes of images (300px x 400px) and image size (140 KB). We don't need to pass a filter to parse decimals on the image height and width since there are pixels. Depending on a bundle just to filter image size doesn't make much sense IMO, so I rewrited the size calculation as a macro without decimals.

On a image with size: 130,677 bytes

Before: 
350px x 500px (130,677o)

After:
350 x 500 (127.61 KB)

### DIFF
--- a/Resources/views/MediaAdmin/edit.html.twig
+++ b/Resources/views/MediaAdmin/edit.html.twig
@@ -9,6 +9,23 @@ file that was distributed with this source code.
 
 #}
 
+{% import _self as macros %}
+
+{% macro bytesToSize(bytes) %}
+{% spaceless %}
+    {% set kilobyte = 1024 %}
+    {% set megabyte = kilobyte * 1024 %}
+
+    {% if bytes < kilobyte %}
+        {{ bytes ~ ' B' }}
+    {% elseif bytes < megabyte %}
+        {{ (bytes / kilobyte)|number_format(2) ~ ' KB' }}
+    {% else %}
+        {{ (bytes / megabyte)|number_format(2) ~ ' MB' }}
+    {% endif %}
+{% endspaceless %}
+{% endmacro %}
+
 {% extends 'SonataAdminBundle:CRUD:base_edit.html.twig' %}
 
 {% block stylesheets %}
@@ -50,8 +67,10 @@ file that was distributed with this source code.
                             <table class="table">
                                 <tr>
                                     <th>{{ 'label.size'|trans({}, 'SonataMediaBundle') }}</th>
-                                    <td>{{ object.width|number_format_decimal }}px x {{ object.height|number_format_decimal }}px
-                                        {% if object.size > 0 %}({{ object.size|number_format_decimal }}o){% endif %}</td>
+                                    <td>
+                                        {{ object.width }} x {{ object.height }}
+                                        {% if object.size > 0 %}({{ macros.bytesToSize(object.size) }}){% endif %}
+                                    </td>
                                 </tr>
                                 <tr>
                                     <th>{{ 'label.content_type'|trans({}, 'SonataMediaBundle') }}</th>
--- a/composer.json
+++ b/composer.json
@@ -41,8 +41,7 @@
         "knplabs/gaufrette": "^0.1.6 || ^0.2",
         "imagine/imagine": "^0.6",
         "kriswallsmith/buzz": "^0.15",
-        "jms/serializer-bundle": "^0.13 || ^1.0",
-        "sonata-project/intl-bundle": "^2.2"
+        "jms/serializer-bundle": "^0.13 || ^1.0"
     },
     "require-dev": {
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
         "sonata-project/formatter-bundle": "^3.0",
         "sonata-project/datagrid-bundle": "^2.2",
         "sonata-project/block-bundle": "^3.1.1",
-        "sonata-project/seo-bundle": "^2.0",
+        "sonata-project/seo-bundle": "^2.1",
         "aws/aws-sdk-php": "^2.7",
         "jackalope/jackalope-doctrine-dbal": "^1.1",
         "symfony/phpunit-bridge": "^2.7 || ^3.0",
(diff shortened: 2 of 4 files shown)


Resources/views/MediaAdmin/edit.html.twig
```diff
@@ -50,8 +67,7 @@ file that was distributed with this source code.
                             <table class="table">
                                 <tr>
                                     <th>{{ 'label.size'|trans({}, 'SonataMediaBundle') }}</th>
-                                    <td>{{ object.width|number_format_decimal }}px x {{ object.height|number_format_decimal }}px
-                                        {% if object.size > 0 %}({{ object.size|number_format_decimal }}o){% endif %}</td>
+                                    <td>{{ object.width }} x {{ object.height }} {% if object.size > 0 %}({{ macros.bytesToSize(object.size) }}){% endif %}</td>
```
greg0ire: please linebreak this

greg0ire: Also, I'd use the native [number_format](http://twig.sensiolabs.org/doc/2.x/filters/number_format.html) here

greg0ire: And why remove the px part?

jordisala1991: I was thinking on using number_format(0), just to avoid decimals (because decimals are problematic with . and , depending on country.

For the other part, for me it is hard to read: 350px x 500px. Feel it reads better with just 350 x 500. If we are already talking about size, we know thats pixels.

greg0ire: As stated in the docs, `number_formats` default can be changed globally, which means you should have the right output for each country if the user configured it properly (maybe there is even automatic configuration based on the locale?).

Ok for dropping `px`, it's indeed a bit pedantic.

greg0ire:
In your example, didn't you mean `127,56Ko` ?

Resources/views/MediaAdmin/edit.html.twig
jordisala1991: Do you mean using number_format for height and width or for the size output?

jordisala1991:
Nope, the size was output like 127,87o without the K

greg0ire:
Oh so this is a bugfix then?

jordisala1991:
Something like that. It could be considered bugfix, removal of things, new feature? (because we format the size better, it can hold MB now)

greg0ire:
Let me get one thing clear : Given a file of 127.28 * 1024 bytes, we have
before : 127,28 (so 1024 times less, huge bug)
after : 127,28 KB (no longer in french by default, and correct)

Am I getting this right?

Resources/views/MediaAdmin/edit.html.twig
jordisala1991: Not sure about what you said of `number_format`. It uses php: http://php.net/manual/es/function.number-format.php. And there is nothing about automatic configuration based on the locale.

greg0ire: - I mean for the size output only.
- The twig doc says : "These defaults can be easily changed through the core extension:"

greg0ire: See [the changelog](https://github.com/symfony/symfony/blob/5129c4cf7e294b1a5ea30d6fec6e89b75396dcd2/src/Symfony/Bundle/TwigBundle/CHANGELOG.md#270), I'm right since 2.7.0

jordisala1991:
No,

On a image with size: 130,677

Old: (130,677o)
New: (128 KB)

greg0ire:
Oh ok I get it, the comma is just a thousands separator.

greg0ire:
So not a patch, nor a minor (I don't think we should do a release just for that). Then this is pedantic.

jordisala1991:
Note that I am removing a dependency. I wasn't sure if it was a BC break or not.

greg0ire:
It isn't. If people are using intl, they have to put it in their composer.json

Resources/views/MediaAdmin/edit.html.twig
jordisala1991: So, we change it for a number_format, with or without decimals? or just calling the filter with defaults parameters?

greg0ire: I'd say 2 decimal places, and other settings: default, like this : `number_format(2)`

jordisala1991: Done

jordisala1991:
Done. This PR contains 2 commits. The first one fixes the build, the second one is changing how we render size and height x width on sonata media edit. If you want me to squash them, or create separate PR just let me know :)

composer.json
```diff
@@ -41,16 +41,15 @@
...
+        "jms/serializer-bundle": "^0.13 || ^1.0"
     },
     "require-dev": {
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
         "sonata-project/admin-bundle": "^3.1",
         "sonata-project/formatter-bundle": "^3.0",
         "sonata-project/datagrid-bundle": "^2.2",
         "sonata-project/block-bundle": "^3.1.1",
-        "sonata-project/seo-bundle": "^2.0",
+        "sonata-project/seo-bundle": "^2.1",
```
greg0ire: Why?

greg0ire:
2 commits are just fine, but I'd explain a bit more why and how the first fixes the build. For the second, please try to keep the subject under 50 chars, and put as much as you want in the body of the commit message. See CONTRIBUTING.md

composer.json
jordisala1991: Because seo-bundle 2.0 is a very old release which contains a TestBase no longer included on 2.1, that's why the build broke but not when using lower deps.

greg0ire: Please put this in the commit message body of the first commit.

jordisala1991:
Should be clearer now in the commit messages.

composer.json
jordisala1991: Done!

greg0ire:
Thanks @jordisala1991 !

PR #1069, opened 2016-06-29
Fix request service call

clytemnestra:
I've explained [here](https://github.com/sonata-project/SonataMediaBundle/issues/1068) method `getRequest()` has been removed in SF3
## Changelog

``` markdown
### Added
- Added `getRequest` method on controller for BC with Symfony 2.3+
```

Issue #1068, opened 2016-06-29
Controller not ready for SF3


clytemnestra:
#### Sonata packages

``` bash
$ composer show sonata-project/*
sonata-project/admin-bundle              3.3.2              The missing Symfony Admin Generator
sonata-project/block-bundle              3.0.1              Symfony SonataBlockBundle
sonata-project/cache                     1.0.7              Cache library
sonata-project/classification-bundle     3.0.0              Symfony SonataClassificationBundle
sonata-project/core-bundle               3.0.3              Symfony SonataCoreBundle
sonata-project/datagrid-bundle           2.2                Symfony SonataDatagridBundle
sonata-project/doctrine-extensions       1.0.2              Doctrine2 behavioral extensions
sonata-project/doctrine-orm-admin-bundle 3.0.4              Symfony Sonata / Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/easy-extends-bundle       2.1.10             Symfony SonataEasyExtendsBundle
sonata-project/exporter                  1.5.0              Lightweight Exporter library
sonata-project/intl-bundle               2.2.4              Symfony SonataIntlBundle
sonata-project/media-bundle              dev-master 1eeae18 Symfony SonataMediaBundle
sonata-project/notification-bundle       3.0.0              Symfony SonataNotificationBundle
```
#### Symfony packages

``` bash
$ composer show symfony/*
symfony/monolog-bundle     2.11.1  Symfony MonologBundle
symfony/phpunit-bridge     v3.1.1  Symfony PHPUnit Bridge
symfony/polyfill-apcu      v1.2.0  Symfony polyfill backporting apcu_* functions to lower PHP versions
symfony/polyfill-intl-icu  v1.2.0  Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-mbstring  v1.2.0  Symfony polyfill for the Mbstring extension
symfony/polyfill-php56     v1.2.0  Symfony polyfill backporting some PHP 5.6+ features to lower PHP versions
symfony/polyfill-php70     v1.2.0  Symfony polyfill backporting some PHP 7.0+ features to lower PHP versions
symfony/polyfill-util      v1.2.0  Symfony utilities for portability of PHP codes
symfony/security-acl       v3.0.0  Symfony Security Component - ACL (Access Control List)
symfony/swiftmailer-bundle v2.3.11 Symfony SwiftmailerBundle
symfony/symfony            v3.1.1  The Symfony PHP framework
[...]

PR #1189, opened 2017-02-06
Remove use of request from container directly

jordisala1991:

Closes #1069 #1068 

## Changelog

```markdown
### Added
- Added `getRequest` method on controller for BC with Symfony 2.3+
```

## To do

- [x] Update the tests

## Subject

This is essentially same as #1069. Since the user didn't answer, I created another PR with the same code.

It just fixes the compatibility with SF3, where you can't obtain request from the container, you have to use request_stack, available since SF2.8

~The build will fail until #1188 gets merged.~

### DIFF
--- a/Controller/MediaController.php
+++ b/Controller/MediaController.php
@@ -57,14 +57,14 @@ class MediaController extends Controller
             throw new NotFoundHttpException(sprintf('unable to find the media with the id : %s', $id));
         }
 
-        if (!$this->get('sonata.media.pool')->getDownloadSecurity($media)->isGranted($media, $this->getRequest())) {
+        if (!$this->get('sonata.media.pool')->getDownloadSecurity($media)->isGranted($media, $this->getCurrentRequest())) {
             throw new AccessDeniedException();
         }
 
         $response = $this->getProvider($media)->getDownloadResponse($media, $format, $this->get('sonata.media.pool')->getDownloadMode($media));
 
         if ($response instanceof BinaryFileResponse) {
-            $response->prepare($this->get('request'));
+            $response->prepare($this->getCurrentRequest());
         }
 
         return $response;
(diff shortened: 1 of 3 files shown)


greg0ire:
#1069 lacked tests. Do you know how to add some?

jordisala1991:
Yes, I will add them asap. Not sure how to test the part where I do not have request_stack though.

greg0ire:
I don't know either, as I usually keep my controllers skinny and don't (unit) test them. So if it is too hard, just tell us, we'll merge as is I think.

jordisala1991:
Not super easy to add test here, but it is doable.. I pushed the first one just to know if this is good enough to keep going.

@greg0ire what do you think?

Controller/MediaController.php
```diff
@@ -134,12 +134,27 @@ public function liipImagineFilterAction($path, $filter)
...
         return $response;
     }
+
+    /**
+     * NEXT_MAJOR: Remove this method when bumping Symfony requirement to 2.8+.
+     * Inject the Symfony\Component\HttpFoundation\Request into the actions instead.
+     *
+     * @return Request
+     */
+    public function getRequest()
```
core23: Why public?

jordisala1991: Because we are extending Controller

jordisala1991: I can call it getCurrentRequest and make it private, that could be better

core23: This would be consistent with our other bundles.

greg0ire: Ah ah just like in this PR I made this weekend on the admin bundle!

jordisala1991: hehe

Tests/Controller/MediaControllerTest.php
```diff
@@ -0,0 +1,176 @@
...
+namespace Sonata\MediaBundle\Tests\Controller;
+
+use Sonata\MediaBundle\Controller\MediaController;
+use Sonata\MediaBundle\Tests\Helpers\PHPUnit_Framework_TestCase;
+
+class MediaControllerTest extends PHPUnit_Framework_TestCase
+{
+    protected $controller;
+
+    public function setUp()
```
greg0ire: protected is enough

jordisala1991:
I think I cannot test the liipImagine method as is, because is using a lot of magic php methods to create a temporal file.

greg0ire:
> I think I cannot test the liipImagine method as is, because is using a lot of magic php methods to create a temporal file.

Don't worry, I think we're already very happy with all the testing you've done. Might be doable with https://github.com/adlawson/php-vfs , if you are bored someday. Tell us when you think this is ready.

jordisala1991:
I think is better to just work on a integration with newer liipImagine, and test that instead of testing this very old code. For me this is ready :)

greg0ire:
Ok! @core23 , please review again and do the honors?

OskarStark:
patch or minor @greg0ire ?

greg0ire:
It fixes deprecations, so patch.

PR #1148, opened 2016-11-25
SF3 compatibility

nicodmf:
All is in the title, this is not the first PR, but this is an important commit and as SF3 has more than one year, this seems to me than this should integrated shortly.

Issue #1111, opened 2016-08-29
Error in create gallery page 'The option "cascade_validation" does not exist.'


valiknet18:
### Environment

Bundle version:     3.2.0
Symfony version 3.1.3
php version 5.6
## Subject

Hello, after install sonata media bundle, in create gallery page (admin), i had error `The option "cascade_validation" does not exist.`.  I don't find resolve to this problem :C.
## Actual results

Page with error.

![image](https://cloud.githubusercontent.com/assets/5772019/18064482/5037a346-6e38-11e6-9a82-0836e1c5e9b3.png)

Stack trace:

```
Stack Trace
in vendor/symfony/symfony/src/Symfony/Component/OptionsResolver/OptionsResolver.php at line 685   -
            ksort($clone->defined);
            ksort($diff);
            throw new UndefinedOptionsException(sprintf(
                (count($diff) > 1 ? 'The options "%s" do not exist.' : 'The option "%s" does not exist.').' Defined options are: "%s".',
                implode('", "', array_keys($diff)),
                implode('", "', array_keys($clone->defined))
at OptionsResolver ->resolve (array('sonata_field_description' => object(FieldDescription), 'type' => 'sonata_type_admin', 'modifiable' => true, 'type_options' => array('sonata_field_description' => object(FieldDescription), 'data_class' => 'Application\Sonata\MediaBundle\Entity\GalleryHasMedia'), 'cascade_validation' => true, 'label_render' => false, 'label' => 'form.label_gallery_has_medias')) 
in vendor/symfony/symfony/src/Symfony/Component/Form/ResolvedFormType.php at line 95   + 
at ResolvedFormType ->createBuilder (object(FormFactory), 'galleryHasMedias', array('sonata_field_description' => object(FieldDescription), 'type' => 'sonata_type_admin', 'modifiable' => true, 'type_options' => array('sonata_field_description' => object(FieldDescription), 'data_class' => 'Application\Sonata\MediaBundle\Entity\GalleryHasMedia'), 'cascade_validation' => true, 'label_render' => false, 'label' => 'form.label_gallery_has_medias')) 
in vendor/symfony/symfony/src/Symfony/Component/Form/Extension/DataCollector/Proxy/ResolvedTypeDataCollectorProxy.php at line 81   + 
[...]

PR #1191, opened 2017-02-06
Fixed compatibility with SF3

jordisala1991:

Closes #1115 #1101 #1148 #1111 

## Changelog

```markdown
### Removed
- `cascade_validation` from `GalleryAdmin`

### Fixed
- Callback contraint is not a valid callable on Symfony 3
- `BaseAdmin` incorrectly retrieved `providerName`
```

## Subject

Essentially the same as #1115 and #1148. But trying to do it in a BC way. Not sure about the Callback constraint, don't know if it is BC. Don't merge until we know.

### DIFF
--- a/Admin/BaseMediaAdmin.php
+++ b/Admin/BaseMediaAdmin.php
@@ -106,7 +106,16 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
         if ($this->hasRequest()) {
             if ($this->getRequest()->isMethod('POST')) {
-                $media->setProviderName($this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true));
+                if (method_exists('Symfony\Component\HttpFoundation\JsonResponse', 'transformJsonError')) {
+                    // NEXT_MAJOR remove this block when dropping sf < 2.8 compatibility
+                    $media->setProviderName(
+                        $this->getRequest()->get(sprintf('%s[providerName]', $this->getUniqid()), null, true)
+                    );
+                } else {
+                    $media->setProviderName(
+                        $this->getRequest()->get($this->getUniqid()['providerName'])
+                    );
+                }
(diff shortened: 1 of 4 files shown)


jordisala1991:
I think the Callback constraint is a BC break. Do you use PHP validation for this case? I mean, move this validation to the entity instead of xml.

greg0ire:
> Do you use PHP validation for this case?

Yes, that would be a good solution! Thanks a lot for carying this!

jordisala1991:
I am not sure how do I know I am in SF2.3 to do the magic IF

```
if () {
    $method = array('methods' => array('isStatusErroneous'));
} else {
    $method = 'isStatusErroneous';
}
$metadata->addConstraint(new Assert\Callback($method));
```

Any idea?

jordisala1991:
Well I think I got it. There is an interface that appears on SF2.5 so it should work this way.

greg0ire:
Why 2.5? Is the new callback constraint usage available since 2.5?

jordisala1991:
Is available since SF 2.4, but I didn't found a better check.

greg0ire:
Ok let's look at the CHANGELOG then.

greg0ire:
> added a constraint the uses the expression language

Not very good english but… should do.

jordisala1991:
Didn't found what that ment... There is a LanguageConstraint but also in SF2.3 so...

greg0ire:
https://github.com/symfony/validator/blob/master/Constraints/Expression.php ?

greg0ire:
Confirmed. Use class exist on that.

jordisala1991:
Done, good catch!

greg0ire:
Awesome job! @sonata-project/contributors , please review!

OskarStark:
Thank you @jordisala1991 👍

PR #1193, opened 2017-02-09
Fixed BlockBundle deprecation messages

jordisala1991:

## Changelog

```markdown
### Fixed
- Fixed BlockBundle deprecation messages
```

## Subject

`BaseBlockService` is deprecated since 3.2 of SonataBlockBundle, changed to `AbstractBlockService`.

Not sure about why SonataBlockBundle is a require-dev. Should I add it to suggest?


jordisala1991:
😱  current() expects parameter 1 to be array, null given

Shouldn't it be BC?

greg0ire:
> Not sure about why SonataBlockBundle is a require-dev. Should I add it to suggest?

It's probably an optional dependency

jordisala1991:
Rebased

greg0ire:
Thanks @jordisala1991 !

PR #1192, opened 2017-02-08
Replaced form types for the FQCN's

jordisala1991:
I am targetting this branch, because this is BC.

## Changelog

```markdown
### Changed
- Replaced form types for the FQCN's
```

## To do

- [x] Finish replacements
- [x] Update the tests

## Subject

Trying to make SF3 work with this bundle. I think this is mandatory for SF3, am I wrong? Because I get a lot of deprecations in SF2.8.
Is there a way, or it is implemented in another way in other sonata bundles to avoid all this code?
Also, do I need to replace DataGridMapper types and ListMapper types?

PR #1196, opened 2017-02-12
Fixed category filter on Media List and Test MediaAdminController

jordisala1991:

Closes #1156 #1130 #1054 

## Changelog

```markdown
### Added
- Added test on `MediaAdminController`
### Fixed
- Media List is now filterable by Category again
```

## To do

- [x] Update the tests

## Subject

Same fix as other PR's but with tests.

Not sure about all the mocks I have created. But it will be needed in order to fully test MediaAdminController.

I have done a little bit of refactor on MediaAdminController based on content in CRUDController (changed isGranted by checkAccess, removing some parameters,...)

### DIFF
--- a/Controller/MediaAdminController.php
+++ b/Controller/MediaAdminController.php
@@ -15,7 +15,6 @@ use Sonata\AdminBundle\Controller\CRUDController as Controller;
 use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 
 class MediaAdminController extends Controller
 {
@@ -24,15 +23,15 @@ class MediaAdminController extends Controller
      */
     public function createAction(Request $request = null)
     {
-        if (false === $this->admin->isGranted('CREATE')) {
-            throw new AccessDeniedException();
-        }
+        $this->admin->checkAccess('create');
 
         if (!$request->get('provider') && $request->isMethod('get')) {
+            $pool = $this->get('sonata.media.pool');
+
             return $this->render('SonataMediaBundle:MediaAdmin:select_provider.html.twig', array(
-                'providers' => $this->get('sonata.media.pool')->getProvidersByContext($request->get('context', $this->get('sonata.media.pool')->getDefaultContext())),
-                'base_template' => $this->getBaseTemplate(),
-                'admin' => $this->admin,
+                'providers' => $pool->getProvidersByContext(
+                    $request->get('context', $pool->getDefaultContext())
+                ),
                 'action' => 'create',
             ));
         }
@@ -43,12 +42,12 @@ class MediaAdminController extends Controller
     /**
      * {@inheritdoc}
      */
-    public function render($view, array $parameters = array(), Response $response = null, Request $request = null)
+    public function render($view, array $parameters = array(), Response $response = null)
     {
-        $parameters['media_pool'] = $this->container->get('sonata.media.pool');
+        $parameters['media_pool'] = $this->get('sonata.media.pool');
         $parameters['persistent_parameters'] = $this->admin->getPersistentParameters();
 
-        return parent::render($view, $parameters, $response, $request);
+        return parent::render($view, $parameters, $response);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


Controller/MediaAdminController.php
```diff
@@ -42,22 +41,20 @@ public function createAction(Request $request = null)
     /**
      * {@inheritdoc}
      */
-    public function render($view, array $parameters = array(), Response $response = null, Request $request = null)
+    public function render($view, array $parameters = array(), Response $response = null)
```
greg0ire: For anyone wondering, this is BC 👍 : https://3v4l.org/LWERG

Tests/Controller/MediaAdminControllerTest.php
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Sonata\MediaBundle\Tests\Controller;
+
+use Prophecy\Argument;
+use Sonata\MediaBundle\Controller\MediaAdminController;
+use Sonata\MediaBundle\Tests\Helpers\PHPUnit_Framework_TestCase;
+
+class MediaAdminControllerTest extends PHPUnit_Framework_TestCase
+{
+    protected $controller;
```
greg0ire: Can and should be private (no big deal though)

Controller/MediaAdminController.php
jordisala1991: It is also overriding a method on CRUD controller (also overriding from Controller), none of the parents have Request as a parameter.

jordisala1991:
Build will fail because of the problem with PHP5.3 and the function call + array access (fixed on #1192).

Tests/Controller/MediaAdminControllerTest.php
jordisala1991: will change it

Controller/MediaAdminController.php
core23: This is NOT BC 👎 : https://3v4l.org/dlWbq

jordisala1991: This is what is actually happening. Im not sure how this works, but CRUDController dont have the last parameter.

The example is the other way around, like @greg0ire said. If someone change my parent class (not sure if thats possible) it is considered BC break? This may be important to another PR I have in mind

greg0ire: @core23 : I think you are getting it backwards : @jordisala1991 is removing a parameter, which means the child class, if there is any will have one more parameter, not one less.

core23: Yes, he removes a paramater, but this isn't BC. Adding a parameter with a default value is, but not the other way round.

greg0ire: It is, see my comment above, with a child class which has one more optional parameter (it's optional because it is after an optional parameter)

jordisala1991: http://symfony.com/doc/current/contributing/code/bc.html

@core23 @greg0ire are we following Symfony BC promise?

If yes, then this is BC, because removing the last parameter is BC on symfony.

greg0ire: No we're not, but yes it is BC, as proven in my example. @core23 : your test is for "the other way around"

jordisala1991:
Test done (once Travis pases)

Controller/MediaAdminController.php
greg0ire: @core23 please review again… can we at least agree that you got your example backwards in https://github.com/sonata-project/SonataMediaBundle/pull/1196#discussion_r100724466 ?

greg0ire: ping @core23 , please don't let this rot. @sonata-project/contributors , if you can chime in…

jordisala1991:
I did a little research on the fourth parameter of render... seems like it was removed because adding it was a BC break.

I am okay with adding again fourth parameter on MedAdminController::render... but then lets define what is BC and what is not, because its not clear and there is a lot of opinion.

https://github.com/sonata-project/SonataAdminBundle/pull/3690

greg0ire:
Can you start a new discussion, preferrably based on a PR you would make on dev-kit?

jordisala1991:
I opened an issue first, will do a PR afterwards.

Controller/MediaAdminController.php
core23: My brain was twisted somehow. Everything fine 👍

greg0ire:
Thanks @jordisala1991 !

PR #1199, opened 2017-02-15
Avoid skipped tests

jordisala1991:
I am targetting this branch, because this only changes tests.

## Subject

Every test is now executed, there are no skipped test. We decrease on test quantity because FixMediaContextCommandTest was extending CommandTest which is a Test class from symfony, not a TestCase.

Not sure about adding a deprecated dependency to required-dev but it was needed to test Amazon. Gauffrete itself uses it on its own Amazon class tho. Don't know if there are any other way of doing it.

### DIFF
--- a/Tests/Command/FixMediaContextCommandTest.php
+++ b/Tests/Command/FixMediaContextCommandTest.php
@@ -17,9 +17,8 @@ use Sonata\MediaBundle\Command\FixMediaContextCommand;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
-use Symfony\Component\Console\Tests\Command\CommandTest;
 
-class FixMediaContextCommandTest extends CommandTest
+class FixMediaContextCommandTest extends \PHPUnit_Framework_TestCase
 {
     /**
      * @var \PHPUnit_Framework_MockObject_MockObject|ContainerInterface
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,7 @@
         "knplabs/gaufrette": "^0.1.6 || ^0.2",
         "kriswallsmith/buzz": "^0.15",
         "sonata-project/classification-bundle": "^3.0",
-        "sonata-project/core-bundle": "^3.0",
+        "sonata-project/core-bundle": "^3.2",
         "sonata-project/doctrine-extensions": "^1.0",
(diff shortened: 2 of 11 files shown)


greg0ire:
> We decrease on test quantity because FixMediaContextCommandTest was extending CommandTest which is a Test class from symfony, not a TestCase.

Do you actually mean increase? If we have more actual tests, it should increase, right?

composer.json
```diff
@@ -23,7 +23,7 @@
         "knplabs/gaufrette": "^0.1.6 || ^0.2",
         "kriswallsmith/buzz": "^0.15",
         "sonata-project/classification-bundle": "^3.0",
-        "sonata-project/core-bundle": "^3.0",
+        "sonata-project/core-bundle": "^3.2",
```
greg0ire: Why is this change needed?

jordisala1991:
Decrease, we had 250~ test and now 217~. But they cover more code!

Tests/Command/FixMediaContextCommandTest.php
```diff
@@ -17,9 +17,8 @@
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Tester\CommandTester;
-use Symfony\Component\Console\Tests\Command\CommandTest;
 
-class FixMediaContextCommandTest extends CommandTest
```
jordisala1991: This line was adding extra Symfony own tests @greg0ire

greg0ire:
> CommandTest which is a Test class from symfony, not a TestCase.

 https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Console/Tests/Command/CommandTest.php extends TestCase so yes, it's a test case.

composer.json
jordisala1991: Because on older version we get this when running lower-deps:

Error: Call to protected method PHPUnit_Framework_TestCase::getMockForAbstractClass() from context 'Sonata\CoreBundle\Test\EntityManagerMockFactory'

jordisala1991:
Yes, every test is itself a TestCase, but I mean this class is not ment to be extended, its a test class

Tests/Command/FixMediaContextCommandTest.php
greg0ire: I get it now.

greg0ire:
@sonata-project/contributors please review

greg0ire:
Thanks @jordisala1991 !

PR #1200, opened 2017-02-16
Fixed namespace on test classes

jordisala1991:
I am targetting this branch, because this only changes test.

## Subject

This is pedantic. I found that Block Test were on a incorrect namespace. This PR just fixes that


OskarStark:
Thank you @jordisala1991 👍

PR #1136, opened 2016-10-24
Remove Dependecy to  Classification Bundle

vguardiola:
I am targetting this branch, because the undependecy process of classification-bundle left a minor bug.

Closes #1133
Fixes #1133
## Changelog

``` markdown
### Removed
Type hinting of a variable with a Class of the an optional Bundle
```
## Subject

Remove Dependecy to  Classification Bundle

PR #1201, opened 2017-02-16
Removed SonataClassificationBundle in a BC way

jordisala1991:

Closes #1136 and maybe another issues

## Changelog

```markdown
### Removed
- ClassificationBundle is now an optional dependency
```

## To do

- [x] Make sure this is BC

## Subject

Last attempt to make this BC. I am not completely sure it is, but I think this might be doable. If not, we just close this and change base branch of #1136 to master.

It is essentially same change of https://github.com/sonata-project/SonataMediaBundle/pull/1086 but trying to make it BC.

I removed setCategory method of Media entity and changed with __call method.
~Also changed construct of MediaAdmin to not have categoryManager, not sure about this change tho~. 

It is BC to change required param to optional param on a __construct method? I think it is: https://3v4l.org/Z42At


jordisala1991:
What do you think about this @core23 @greg0ire ?

Admin/BaseMediaAdmin.php
```diff
@@ -38,7 +38,7 @@
      * @param Pool                     $pool
      * @param CategoryManagerInterface $categoryManager
      */
-    public function __construct($code, $class, $baseControllerName, Pool $pool, CategoryManagerInterface $categoryManager)
+    public function __construct($code, $class, $baseControllerName, Pool $pool, CategoryManagerInterface $categoryManager = null)
```
greg0ire: This is a BC-break for child classes.

greg0ire: Wait no it isn't, the LSP does not apply to `__construct`

Command/FixMediaContextCommand.php
```diff
@@ -32,9 +32,13 @@ public function configure()
      */
     public function execute(InputInterface $input, OutputInterface $output)
     {
+        if (!$this->getContainer()->has('sonata.media.manager.category')) {
+            throw new \LogicException('The classification feature is disabled.');
+        }
+
         $pool = $this->getContainer()->get('sonata.media.pool');
         $contextManager = $this->getContainer()->get('sonata.classification.manager.context');
-        $cateoryManager = $this->getContainer()->get('sonata.classification.manager.category');
```
greg0ire: lol that variable name

Model/CategoryManager.php
```diff
@@ -0,0 +1,82 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Model;
+
+use Sonata\ClassificationBundle\Model\CategoryManagerInterface as ManagerInterface;
+
+/**
+ * @author Joao Albuquerque <albuquerque.joao.filipe@gmail.com>
+ * @author Christian Gripp <mail@core23.de>
```
greg0ire: Uuuuh why?

Model/CategoryManagerInterface.php
```diff
@@ -0,0 +1,63 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Model;
+
+use Sonata\ClassificationBundle\Model\CategoryInterface;
+
+/**
+ * @author Joao Albuquerque <albuquerque.joao.filipe@gmail.com>
+ * @author Christian Gripp <mail@core23.de>
```
greg0ire: Same here, why?

jordisala1991: Back porting some code from master, I should not change author.

Model/CategoryManager.php
jordisala1991: Same as below

jordisala1991:
I think we might add some checks on the __call method like this:

```
if (!$category instanceof CategoryInterface) {
    throw new \InvalidArgumentException('Argument 1 should be an instance of Sonata\ClassificationBundle\Model\CategoryInterface or null');
}
```

jordisala1991:
Added that check and a revisited test for Media. Added both cases, when there is a category and when there is not.

Model/Media.php
```diff
@@ -153,6 +155,19 @@ public function __toString()
         return $this->getName() ?: 'n/a';
     }
 
+    public function __call($method, $arguments)
+    {
+        if ($method == 'setCategory') {
+            $category = current($arguments);
+
+            if (null !== $category && !is_a($category, 'Sonata\ClassificationBundle\Model\CategoryInterface')) {
+                throw new \InvalidArgumentException('Argument 1 should be an instance of Sonata\ClassificationBundle\Model\CategoryInterface or null');
```
OskarStark: can you please linebreak this?

core23: Why?  Thought we have a decision

OskarStark: Really? Didn't get that..

Soullivaneuh: @OskarStark SF code style, we don't linebreak error/exception messages. :wink:

greg0ire: Uuuh no we don't… https://github.com/sonata-project/dev-kit/issues/218

greg0ire: Maybe ask @jordisala1991 to vote on this?

jordisala1991: If you ask me, second solution is the one I like the most.

For me when I see a line super long I just split it in a way that it (more or less) fits well... but I don't have the perfect solution that applies for all the cases.

Also like to follow SF code style rules, but we don't know if they are using the rule of no line breaks on error/exception because they like it or because it's hard to change now.

jordisala1991:
I would like to know if there is any BC break on this PR, before changing more things.

Also there is a upcoming bundle for SonataClassification to add back the thumbnail removed on master: https://github.com/sonata-project/SonataClassificationMediaBundle/pull/1.

Could we use that on master to move all this code also to that bundle? I think it could work the same way. @core23 (since you are moving that PR) do you think we can do the same for this case?

greg0ire:
> I would like to know if there is any BC break on this PR, before changing more things.

I don't see any.

DependencyInjection/SonataMediaExtension.php
```diff
@@ -548,6 +553,38 @@ public function configureClassesToCompile()
...
+     * Checks if the classification of media is enabled.
+     *
+     * @param array $config
+     *
+     * @return bool
+     */
+    private function isClassificationEnabled(array $config)
+    {
+        return !$config['force_disable_category'] &&
+            (null !== $config['category_manager'] || interface_exists('Sonata\ClassificationBundle\Model\CategoryInterface'));
```
core23: Should be `&&` instead of `||`

jordisala1991: Are you sure? on master is with an `||` and it makes sense because this line makes a default if category_manager is not defined:

```
$container->setAlias('sonata.media.manager.category', $config['category_manager'] ?: 'sonata.media.manager.category.default');
```

jordisala1991: I did a refactor on this part, to avoid changing some $config variables and default values.

Can you review again @core23 ?

jordisala1991:
@sonata-project/contributors please review, I've double checked everything. Tested with and without SonataClassification (using force_disable and not using it also) and everything works fine.

greg0ire:
Thanks @jordisala1991 !!!

OskarStark:
Minor I guess

jlamur:
Is it a BC break if your application doesn't work after a `composer update`?

If you installed the *MediaBundle* with `composer require sonata-project/media-bundle`, the *ClassificationBundle* was added as dependency.
Now that this dependency is removed, a `composer update` remove the *ClassificationBundle* and your application is broken...

Even though the easy solution is to install the *ClassificationBundle* with `composer require sonata-project/classification-bundle`, it sounds as a BC break for me...

greg0ire:
> Is it a BC break if your application doesn't work after a composer update?

Yes it is one, but I think it is acceptable.

jordisala1991:
> Even though the easy solution is to install the ClassificationBundle with composer require sonata-project/classification-bundle, it sounds as a BC break for me...

If you use `sonata-project/classification-bundle` in your app (you have it declared on AppKernel.php) you should add it to your composer.json

greg0ire:
That's right, everything you reference in your code should appear in your composer.json

jlamur:
Well, so thank you both for your advices and for this fantastic PR @jordisala1991 :)

greg0ire:
See http://stackoverflow.com/questions/42628459/upgrading-symfony-sonata-to-3-x/42628911#42628911 :P

PR #1203, opened 2017-02-19
Refactor on GalleryAdminController

jordisala1991:

## To do

- [x] Update the tests

## Subject

Same changes applied on MediaAdminController here https://github.com/sonata-project/SonataMediaBundle/pull/1196, but applied to GalleryAdminController.

### DIFF
--- a/Controller/GalleryAdminController.php
+++ b/Controller/GalleryAdminController.php
@@ -12,9 +12,9 @@
 namespace Sonata\MediaBundle\Controller;
 
 use Sonata\AdminBundle\Controller\CRUDController as Controller;
+use Symfony\Component\Form\FormView;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
-use Symfony\Component\Security\Core\Exception\AccessDeniedException;
 
 class GalleryAdminController extends Controller
 {
@@ -22,16 +22,15 @@ class GalleryAdminController extends Controller
      * @param string   $view
      * @param array    $parameters
      * @param Response $response
-     * @param Request  $request
      *
      * @return Response
      */
-    public function render($view, array $parameters = array(), Response $response = null, Request $request = null)
+    public function render($view, array $parameters = array(), Response $response = null)
     {
-        $parameters['media_pool'] = $this->container->get('sonata.media.pool');
+        $parameters['media_pool'] = $this->get('sonata.media.pool');
         $parameters['persistent_parameters'] = $this->admin->getPersistentParameters();
 
-        return parent::render($view, $parameters, $response, $request);
+        return parent::render($view, $parameters, $response);
     }
 
     /**
@@ -43,9 +42,7 @@ class GalleryAdminController extends Controller
      */
     public function listAction(Request $request = null)
     {
-        if (false === $this->admin->isGranted('LIST')) {
-            throw new AccessDeniedException();
-        }
+        $this->admin->checkAccess('list');
 
         if ($listMode = $request->get('_list_mode')) {
             $this->admin->setListMode($listMode);
@@ -57,7 +54,7 @@ class GalleryAdminController extends Controller
         $formView = $datagrid->getForm()->createView();
 
         // set the theme for the current Admin Form
-        $this->get('twig')->getExtension('form')->renderer->setTheme($formView, $this->admin->getFilterTheme());
+        $this->setFormTheme($formView, $this->admin->getFilterTheme());
 
         return $this->render($this->admin->getTemplate('list'), array(
             'action' => 'list',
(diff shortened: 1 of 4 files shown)


Controller/GalleryAdminController.php
```diff
@@ -66,4 +62,27 @@ public function listAction(Request $request = null)
...
+     */
+    private function setFormTheme(FormView $formView, $theme)
+    {
+        $twig = $this->get('twig');
+
+        try {
+            $twig
+                ->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')
+                ->setTheme($formView, $theme);
+        } catch (\Twig_Error_Runtime $e) {
```
greg0ire: I know we use that somewhere else, but IMO [there are less violent methods for detected twig bridge 3.2](https://github.com/sonata-project/SonataAdminBundle/blob/3.x/Tests/Controller/CRUDControllerTest.php#L194), and they should be preferred, because it might also be more accurate: what if a `\Twig_Error_Runtime` is thrown in `^3.2` for a totally unrelated reason?

jordisala1991: I will open a PR on admin bundle to change this. The root problem is CRUDController. Once is solved there an we have a stable release we can change it here too. what do you think?

greg0ire: I don't understand why there is a need for a particular order 😕

jordisala1991: On this particular case there is no problem. But on the MediaAdminController there was an issue because we were calling CRUDController render and could be hard to mock there (had to get admin-bundle dependency up because of that). Related PR just got merged.

jordisala1991: Maybe we can use this controller as an example of how to change formTheme. You are right!

jordisala1991:
I did change the setFormTheme detection, have a look a it @greg0ire. Have special case for the mediaAdminController test

Controller/MediaAdminController.php
```diff
@@ -116,16 +116,12 @@ private function setFormTheme(FormView $formView, $theme)
...
-            $twig
-                ->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')
-                ->setTheme($formView, $theme);
-        } catch (\Twig_Error_Runtime $e) {
-            // BC for Symfony < 3.2 where this runtime not exists
-            $twig
-                ->getExtension('Symfony\Bridge\Twig\Extension\FormExtension')
-                ->renderer
-                ->setTheme($formView, $theme);
+        // BC for Symfony < 3.2 where this runtime not exists
```
greg0ire: "not exists" => "does not exist"

Controller/MediaAdminController.php
```diff
@@ -116,16 +116,12 @@ private function setFormTheme(FormView $formView, $theme)
...
+        if (method_exists('Symfony\Bridge\Twig\AppVariable', 'getToken')) {
+            $twig->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')->setTheme($formView, $theme);
+        } else {
+            $twig->getExtension('Symfony\Bridge\Twig\Extension\FormExtension')
+                ->renderer->setTheme($formView, $theme);
```
greg0ire: Using an early return statement, you can make the code even more similar to what it will be when we drop sf <3.2 : 

```php
if (!method_exists('Symfony\Bridge\Twig\AppVariable', 'getToken')) {
    $twig->getExtension('Symfony\Bridge\Twig\Extension\FormExtension')
        ->renderer->setTheme($formView, $theme);
        return;
}
$twig->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')->setTheme($formView, $theme);
```

Tests/Controller/GalleryAdminControllerTest.php
```diff
@@ -0,0 +1,150 @@
...
+use Sonata\MediaBundle\Controller\GalleryAdminController;
+
+class GalleryAdminControllerTest extends \PHPUnit_Framework_TestCase
+{
+    private $container;
+    private $admin;
+    private $request;
+    private $controller;
+
+    public function setUp()
```
greg0ire: `protected` please, no need to widen the visiblity.

Tests/Controller/MediaAdminControllerTest.php
```diff
@@ -173,13 +175,25 @@ private function configureSetFormTheme($formView, $formTheme)
         $twig = $this->prophesize('\Twig_Environment');
         $twigRenderer = $this->prophesize('Symfony\Bridge\Twig\Form\TwigRenderer');
 
-        $twig->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')->willReturn($twigRenderer->reveal());
-        $twigRenderer->setTheme($formView, $formTheme)->shouldBeCalled();
         $this->container->get('twig')->willReturn($twig->reveal());
+
+        // NEXT_MAJOR: Remove this trick when bumping Symfony requirement to 3.2+.
```
greg0ire: I don't think we will bump this requirement when releasing next major.

Tests/Controller/GalleryAdminControllerTest.php
jordisala1991: I do the same mistake over and over

Tests/Controller/MediaAdminControllerTest.php
jordisala1991: Changing it

jordisala1991:
Now it should be okay

jordisala1991:
Ping @OskarStark or @core23

greg0ire:
Thanks @jordisala1991 !

PR #1206, opened 2017-02-20
Calling a macro without being imported is an error on twig 2.0

jordisala1991:
I am targetting this branch, because is a bug fix.

## Changelog

```markdown
### Fixed
- Calling a macro without importing it is an error on twig 2.0
```

## Subject

On Twig 2.0 is mandatory to import every macro you are going to use. When calling a macro inside another macro you also need to import it.


jordisala1991:
Done both changes.

PR #1207, opened 2017-02-22
typo

Chrisvore:
correcting a typo on the xml

I am targetting this branch, because it will not break any BC.

No issue

## Subject
removed `sd">` on line 7


greg0ire:
Nice catch @Chrisvore

PR #1057, opened 2016-06-11
make gaufrette config compatible with sf3 - master branch

hug963:
### Changelog

``` markdown
### Fixed
- `gaufrette.xml` is now compatible with Symfony 3
```
### Subject

The Gaufrette configuration was using the deprecated functions setFactoryClass, setFactoryService and setFactoryMethod.

This removes support for symfony < 2.6

Fix source https://github.com/doctrine/DoctrineBundle/pull/381

PR #1212, opened 2017-02-25
Remove old Symfony 2.3 code

jordisala1991:
I am targetting this branch, because this is removing compatibility with Symfony 2.3.

Closes #1057 

## Changelog

```markdown
### Changed
- Changed `MediaController::downloadAction` signature to include Request as first parameter
- Changed `MediaController::viewAction` signature to include Request as first parameter
- Changed `MediaController::liipImagineFilterAction` signature to include Request as first parameter
```

## Subject

I am removing unused lines of codes, used for symfony 2.3 compatibility on 3.x branch.
Also changing some controllers signatures to accept `$request`as a parameter.

### DIFF
--- a/Controller/MediaController.php
+++ b/Controller/MediaController.php
@@ -15,6 +15,7 @@ use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 use Symfony\Component\Security\Core\Exception\AccessDeniedException;
@@ -44,12 +45,13 @@ class MediaController extends Controller
     /**
      * @throws NotFoundHttpException
      *
-     * @param string $id
-     * @param string $format
+     * @param Request $request
+     * @param string  $id
+     * @param string  $format
      *
      * @return Response
      */
-    public function downloadAction($id, $format = 'reference')
+    public function downloadAction(Request $request, $id, $format = 'reference')
     {
         $media = $this->getMedia($id);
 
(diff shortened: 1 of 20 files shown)


Controller/MediaController.php
```diff
@@ -44,27 +45,28 @@ public function getMedia($id)
...
-     * @param string $id
-     * @param string $format
+     * @param Request $request
+     * @param string  $id
+     * @param string  $format
      *
      * @return Response
      */
-    public function downloadAction($id, $format = 'reference')
+    public function downloadAction(Request $request, $id, $format = 'reference')
```
greg0ire: This is a BC-break, which means there should be something in the `UPGRADE` guide at least, and at most, a PR should be made on the stable branch to simulate both signatures, and deprecate the old one (that would help with migrating calls, but not with migrating child classes).

jordisala1991: Is there a way to simulate this signature on 3.x? I am not sure how to trigger this deprecation

jordisala1991: I did update this PR with a note on UPGRADE-4.0 and a changelog. If we find a way to do this in 3.x, I will create another PR first for 3.x branch

greg0ire: You can simulate any signature with `func_get_args()`

jordisala1991: Do you mean like removing all parameters and get them using `func_get_args()`? The only problem I see with that is the fact that AFAIK Symfony only sends the request if you type hint that parameter.

greg0ire: I meant by keeping the old signature, and check the first parameter's type, but indeed, I overlooked the fact that this is an action and some magic is happening here. So this isn't possible.

UPGRADE-4.0.md
```diff
@@ -9,20 +9,25 @@ Please read [3.x](https://github.com/sonata-project/SonataMediaBundle/tree/3.x)
 
 See also the [diff code](https://github.com/sonata-project/SonataMediaBundle/compare/3.x...4.0.0).
 
+## Controllers
+
+`MediaController` action were changed to introduce `Request $request` as first parameter. You must update those signatures if you want to still extend this class: `downloadAction`, `listAction` and `liipImagineFilterAction`
```
greg0ire: "actions"

greg0ire: Please wrap to 120 chars, cutting your sentences in a semantical way.

jordisala1991:
@sonata-project/contributors Please, review again

OskarStark:
Great job @jordisala1991 👍 🎉

PR #1213, opened 2017-02-26
Remove deprecations from non FQCNs on form types

jordisala1991:

## Subject

This PR tries to avoid deprecation warnings on Symfony2.8 with non FQCNs. 

I didn't spot it because it does not produce an error on Symfony3. There is some magic on Sonata that resolves this `'choice'`to its FQCN but only on SF3... since I don't like magic, I just replaced it with the correct value.

### DIFF
--- a/Admin/ORM/MediaAdmin.php
+++ b/Admin/ORM/MediaAdmin.php
@@ -29,13 +29,18 @@ class MediaAdmin extends Admin
             $options['choices'][$name] = $name;
         }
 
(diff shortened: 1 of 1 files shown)


OskarStark:
Thank you @jordisala1991, can you please add a CHANEGLOG not in the PR header?

jordisala1991:
@OskarStark added Changelog note. I didn't add this before because this changelog note is suposed to be on an earlier PR, those are some edge cases that were left from that one.

PR #1243, opened 2017-04-12
Move to https on Youtube urls

jordisala1991:

## Changelog

```markdown
### Changed
- Change Youtube urls to use https
```

## Subject

Youtube urls for fetching metadata and referenceUrl were using http. It is preferable to just use https.

Not sure if patch or pedantic


greg0ire:
It's a patch for https users, because otherwise they would get mixed content. Tests need to be fixed apparently.

jordisala1991:
It should be fixed now

greg0ire:
Thanks @jordisala1991

PR #1246, opened 2017-04-19
Fix #1107: Catch exception preventing correct handling of upload errors

lemoinem:
I am targeting this branch (3.x), because this a fix for #1107 (and as per @jordisala1991's request).

Closes #1107. Before this PR, when there was an upload error (e.g., file too large), the DataTransformer was throwing an exception which was preventing the form from ever reaching the validation step. This was triggering a 500 Response instead of a 200 with a Form Error.

## Changelog
```markdown
### Changed
- The DataTransformers and MediaTypes (both standard and API) now depend on `Psr\Log\LoggerInterface` in order to log any exception that could arise from `$provider->transform()`

### Added
- Optional dependency on `@logger`. If the service is unavailable the exception will be silently caught and discarded.

### Fixed
- Now we have form errors when uploaded files are too big!
```

## Subject

Error message on upload fail (most of all, file too big).

### DIFF
--- a/Form/DataTransformer/ProviderDataTransformer.php
+++ b/Form/DataTransformer/ProviderDataTransformer.php
@@ -11,11 +11,14 @@
 
 namespace Sonata\MediaBundle\Form\DataTransformer;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Form\DataTransformerInterface;
 
-class ProviderDataTransformer implements DataTransformerInterface
+class ProviderDataTransformer implements DataTransformerInterface, LoggerAwareInterface
 {
     /**
      * @var Pool
@@ -28,6 +31,13 @@ class ProviderDataTransformer implements DataTransformerInterface
     protected $options;
 
     /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Pool   $pool
      * @param string $class
      * @param array  $options
@@ -37,6 +47,18 @@ class ProviderDataTransformer implements DataTransformerInterface
         $this->pool = $pool;
         $this->options = $this->getOptions($options);
         $this->class = $class;
+
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @param LoggerInterface $logger
+     */
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
     }
 
     /**
@@ -94,7 +116,17 @@ class ProviderDataTransformer implements DataTransformerInterface
 
         $provider = $this->pool->getProvider($newMedia->getProviderName());
 
-        $provider->transform($newMedia);
+        try {
+            $provider->transform($newMedia);
+        } catch (\Exception $e) { // NEXT_MAJOR: When switching to PHP 7+, change this to \Throwable
+            // #1107 We must never throw an exception here.
+            // An exception here would prevent us to provide meaningful errors through the Form
+            // Error message inspired from Monolog\ErrorHandler
+            $this->logger->error(
+                sprintf('Caught Exception %s: "%s" at %s line %s', get_class($e), $e->getMessage(), $e->getFile(), $e->getLine()),
+                array('exception' => $e)
+            );
+        }
 
         return $newMedia;
     }
--- a/Form/DataTransformer/ServiceProviderDataTransformer.php
+++ b/Form/DataTransformer/ServiceProviderDataTransformer.php
@@ -11,11 +11,14 @@
 
 namespace Sonata\MediaBundle\Form\DataTransformer;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Symfony\Component\Form\DataTransformerInterface;
 
-class ServiceProviderDataTransformer implements DataTransformerInterface
+class ServiceProviderDataTransformer implements DataTransformerInterface, LoggerAwareInterface
 {
     /**
      * @var MediaProviderInterface
@@ -23,11 +26,30 @@ class ServiceProviderDataTransformer implements DataTransformerInterface
     protected $provider;
 
     /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param MediaProviderInterface $provider
      */
     public function __construct(MediaProviderInterface $provider)
     {
         $this->provider = $provider;
+
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @param LoggerInterface $logger
+     */
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
     }
 
     /**
@@ -47,7 +69,17 @@ class ServiceProviderDataTransformer implements DataTransformerInterface
             return $media;
         }
 
-        $this->provider->transform($media);
+        try {
+            $this->provider->transform($media);
+        } catch (\Exception $e) { // NEXT_MAJOR: When switching to PHP 7+, change this to \Throwable
+            // #1107 We must never throw an exception here.
+            // An exception here would prevent us to provide meaningful errors through the Form
+            // Error message taken from Monolog\ErrorHandler
+            $this->logger->error(
+                sprintf('Caught Exception %s: "%s" at %s line %s', get_class($e), $e->getMessage(), $e->getFile(), $e->getLine()),
+                array('exception' => $e)
+            );
+        }
 
         return $media;
     }
--- a/Form/Type/ApiMediaType.php
+++ b/Form/Type/ApiMediaType.php
@@ -11,6 +11,9 @@
 
 namespace Sonata\MediaBundle\Form\Type;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Form\AbstractType;
@@ -21,7 +24,7 @@ use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 /**
  * @author Hugo Briand <briand@ekino.com>
  */
-class ApiMediaType extends AbstractType
+class ApiMediaType extends AbstractType implements LoggerAwareInterface
 {
     /**
      * @var Pool
@@ -34,6 +37,13 @@ class ApiMediaType extends AbstractType
     protected $class;
 
     /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Pool   $mediaPool
      * @param string $class
      */
@@ -41,6 +51,17 @@ class ApiMediaType extends AbstractType
     {
         $this->mediaPool = $mediaPool;
         $this->class = $class;
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @param LoggerInterface $logger
+     */
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
     }
 
     /**
@@ -48,9 +69,12 @@ class ApiMediaType extends AbstractType
      */
     public function buildForm(FormBuilderInterface $builder, array $options)
     {
-        $builder->addModelTransformer(new ProviderDataTransformer($this->mediaPool, $this->class, array(
+        $dataTransformer = new ProviderDataTransformer($this->mediaPool, $this->class, array(
             'empty_on_new' => false,
-        )), true);
+        ));
+        $dataTransformer->setLogger($this->logger);
+
+        $builder->addModelTransformer($dataTransformer, true);
 
         $provider = $this->mediaPool->getProvider($options['provider_name']);
         $provider->buildMediaType($builder);
--- a/Form/Type/MediaType.php
+++ b/Form/Type/MediaType.php
@@ -11,6 +11,9 @@
 
 namespace Sonata\MediaBundle\Form\Type;
 
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerInterface;
+use Psr\Log\NullLogger;
 use Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Form\AbstractType;
@@ -22,7 +25,7 @@ use Symfony\Component\Form\FormView;
 use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 
-class MediaType extends AbstractType
+class MediaType extends AbstractType implements LoggerAwareInterface
 {
     /**
      * @var Pool
@@ -35,6 +38,13 @@ class MediaType extends AbstractType
     protected $class;
 
     /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Pool   $pool
      * @param string $class
      */
@@ -42,6 +52,17 @@ class MediaType extends AbstractType
     {
         $this->pool = $pool;
         $this->class = $class;
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * NEXT_MAJOR: When switching to PHP 5.4+, replace by LoggerAwareTrait.
+     *
+     * @param LoggerInterface $logger
+     */
+    public function setLogger(LoggerInterface $logger)
+    {
+        $this->logger = $logger;
     }
 
     /**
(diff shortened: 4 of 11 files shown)


greg0ire:
> * Dependence on monolog-bundle (not sure this was the right way to go, but I need `Psr\Log\LoggerInterface` AND the `@logger` service.

Then require `psr/log` ?

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -23,6 +24,11 @@ class ProviderDataTransformer implements DataTransformerInterface
     protected $pool;
 
     /**
+     * @var LoggerInterface
+     */
+    protected $logger;
```
greg0ire: Please use private by default

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -32,9 +38,10 @@ class ProviderDataTransformer implements DataTransformerInterface
      * @param string $class
      * @param array  $options
      */
-    public function __construct(Pool $pool, $class, array $options = array())
+    public function __construct(Pool $pool, $class, LoggerInterface $logger, array $options = array())
```
greg0ire: Please put that argument last for BC's sake, and make it optional. When it is null, detect it and assign a `NullLogger` instance to `$logger`

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -94,7 +101,17 @@ public function reverseTransform($media)
 
         $provider = $this->pool->getProvider($newMedia->getProviderName());
 
-        $provider->transform($newMedia);
+        try {
+            $provider->transform($newMedia);
+        } catch (\Exception $e) { // NEXT_MAJOR: When removing support for PHP5, change this to \Throwable
+            // #1107 We must never throw exception.
+            // This allow us to validate the Media correctly show upload errors.
```
greg0ire: This sentence doesn't parse

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -94,7 +101,17 @@ public function reverseTransform($media)
...
+            // Error message taken from Monolog\ErrorHandler
+            $this->logger->error(
+                sprintf('Uncaught Exception %s: "%s" at %s line %s', get_class($e), $e->getMessage(), $e->getFile(), $e->getLine()),
```
greg0ire: "Uncaught" is wrong ;)

jordisala1991:
Isn't better to make separate PR for this 2 changes?

- Fix form error handling
- Add logs

Form/Type/ApiMediaType.php
```diff
@@ -37,18 +43,19 @@ class ApiMediaType extends AbstractType
      * @param Pool   $mediaPool
      * @param string $class
      */
-    public function __construct(Pool $mediaPool, $class)
+    public function __construct(Pool $mediaPool, $class, LoggerInterface $logger)
```
OskarStark: BC break

Form/Type/MediaType.php
```diff
@@ -38,18 +44,19 @@ class MediaType extends AbstractType
      * @param Pool   $pool
      * @param string $class
      */
-    public function __construct(Pool $pool, $class)
+    public function __construct(Pool $pool, $class, LoggerInterface $logger)
```
OskarStark: BC break

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: Should I switch to private the other properties as well?

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: Ok for last position.

Making it optional really seems wrong thought... We're not talking about debugging/info/warning logging here, but an actual exception that's willingly silence... basically the equivalent of an old times `@` in front of a function call.
It makes debugging impossible. Not logging shouldn't really be an option here..

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: Would this be better?

> An exception here would prevent us to provide meaningful errors through the Form

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: Indeed...

lemoinem:
@greg0ire I thought about it, but `psr/log` does not provide the `@logger` service (see my other comment).

@jordisala1991 I wondered about this as well, but for reasons detailed in another comment (https://github.com/sonata-project/SonataMediaBundle/pull/1246#discussion_r112342551), catching an exception and discarding it silently, without logging it, really feels wrong...

Form/DataTransformer/ProviderDataTransformer.php
greg0ire: If you want to stay BC, you have to put it last, and if you want to put it last, you have to make it optional. Instead of using `NullLogger`, you can choose to rethrow the exception if the logger is null though.

Form/DataTransformer/ProviderDataTransformer.php
greg0ire: Yes 👍

greg0ire:
> it looks like the SonataMediaExtensionTest wasn't included.

What do you mean by that?

lemoinem:
> > it looks like the SonataMediaExtensionTest wasn't included.
>
> What do you mean by that?

I mean neither `phpunit` nor `vendor/bin/simple-phpunit` ran this Test. So I wonder if I missed some configuration or dependencies...

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: rethrowing seems fair enough in this context.

Thanks, I will go with that!

lemoinem:
All right, I think I addressed all your comments, don't hesitate to tell me if there are any more updates you'd like.

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -28,13 +29,19 @@ class ProviderDataTransformer implements DataTransformerInterface
...
+     */
+    private $logger;
+
+    /**
      * @param Pool   $pool
      * @param string $class
      * @param array  $options
      */
-    public function __construct(Pool $pool, $class, array $options = array())
+    public function __construct(Pool $pool, $class, array $options = array(), LoggerInterface $logger = null)
```
OskarStark: next major, this shouldn't be optional, or am I wrong?

Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -23,11 +24,17 @@ class ServiceProviderDataTransformer implements DataTransformerInterface
...
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param MediaProviderInterface $provider
      */
-    public function __construct(MediaProviderInterface $provider)
+    public function __construct(MediaProviderInterface $provider, LoggerInterface $logger = null)
```
OskarStark: same here for next major?

Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -47,7 +54,21 @@ public function reverseTransform($media)
...
         }
 
-        $this->provider->transform($media);
+        try {
+            $this->provider->transform($media);
+        } catch (\Exception $e) { // NEXT_MAJOR: When removing support for PHP5, change this to \Throwable
+            // #1107 We must never throw exception.
+            // An exception here would prevent us to provide meaningful errors through the Form
+            // Error message taken from Monolog\ErrorHandler
+            if (null !== $this->logger) {
```
OskarStark: next major this check should then be removed, if logger is present all the time

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -94,7 +101,21 @@ public function reverseTransform($media)
...
         $provider = $this->pool->getProvider($newMedia->getProviderName());
 
-        $provider->transform($newMedia);
+        try {
+            $provider->transform($newMedia);
+        } catch (\Exception $e) { // NEXT_MAJOR: When removing support for PHP5, change this to \Throwable
+            // #1107 We must never throw exception.
+            // An exception here would prevent us to provide meaningful errors through the Form
+            // Error message taken from Monolog\ErrorHandler
+            if (null !== $this->logger) {
```
OskarStark: next major this check should then be removed, if logger is present all the time

Form/Type/ApiMediaType.php
```diff
@@ -34,13 +35,19 @@ class ApiMediaType extends AbstractType
...
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Pool   $mediaPool
      * @param string $class
      */
-    public function __construct(Pool $mediaPool, $class)
+    public function __construct(Pool $mediaPool, $class, LoggerInterface $logger = null)
```
OskarStark: optional in next major?

Form/Type/MediaType.php
```diff
@@ -35,13 +36,19 @@ class MediaType extends AbstractType
...
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    /**
      * @param Pool   $pool
      * @param string $class
      */
-    public function __construct(Pool $pool, $class)
+    public function __construct(Pool $pool, $class, LoggerInterface $logger = null)
```
OskarStark: optional in next major?

Resources/config/api_form_doctrine_mongodb.xml
```diff
@@ -13,6 +13,7 @@
             <tag name="form.type" alias="sonata_media_api_form_media"/>
             <argument type="service" id="sonata.media.pool"/>
             <argument>%sonata.media.admin.media.entity%</argument>
+            <argument type="service" id="logger" on-invalid="null"/>
```
OskarStark: next major still `on-invalid`?

Resources/config/api_form_doctrine_orm.xml
```diff
@@ -13,6 +13,7 @@
             <tag name="form.type" alias="sonata_media_api_form_media"/>
             <argument type="service" id="sonata.media.pool"/>
             <argument>%sonata.media.admin.media.entity%</argument>
+            <argument type="service" id="logger" on-invalid="null"/>
```
OskarStark: same here?

Resources/config/api_form_doctrine_phpcr.xml
```diff
@@ -13,6 +13,7 @@
             <tag name="form.type" alias="sonata_media_api_form_media"/>
             <argument type="service" id="sonata.media.pool"/>
             <argument>%sonata.media.admin.media.entity%</argument>
+            <argument type="service" id="logger" on-invalid="null"/>
```
OskarStark: same here?

Resources/config/form.xml
```diff
@@ -5,6 +5,7 @@
             <tag name="form.type" alias="sonata_media_type"/>
             <argument type="service" id="sonata.media.pool"/>
             <argument/>
+            <argument type="service" id="logger" on-invalid="null"/>
```
OskarStark: same here?

Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -47,7 +54,21 @@ public function reverseTransform($media)
...
+                $this->logger->error(
+                    sprintf('Caught Exception %s: "%s" at %s line %s', get_class($e), $e->getMessage(), $e->getFile(), $e->getLine()),
+                    array('exception' => $e)
+                );
+            } else {
+                throw new \Exception('Uncaught Exception', 0, $e);
```
greg0ire: This exception brings nothing new, please just `throw $e` instead.

Form/DataTransformer/ProviderDataTransformer.php
greg0ire: This would imply requiring the monolog bundle (which would be ok though)

OskarStark: ok fine

jordisala1991: If it is not required now, atleast it should be suggested right?

greg0ire: Yes

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: Unless we want to move the argument's position around, I don't think we can make the argument mandatory itself...

Form/DataTransformer/ProviderDataTransformer.php
lemoinem: I will add a deprecation warning for null logger in this PR as well.

Form/DataTransformer/ProviderDataTransformer.php
core23: Could you use the `LoggerAwareTrait` instead and use the `NullLogger` by default?

lemoinem: Using LoggerAwareTrait could definitely be another option...

I've had a few alternatives offered for this issue...
I will add a global PR comment later today ([EDT](https://www.timeanddate.com/time/zones/edt)), summarize them, and wait for a definitive answer. I'd like to find a solution satisfying everyone.

lemoinem:
I've had several contradictory or incompatible suggestions regarding the PR, I'd like us to reach a consensus before making the modifications (to avoir doing it 4 times...)

@core23 @OskarStark @greg0ire @jordisala1991 Since you are the four who gave me suggestions, here is a list of the choices I could gather:

1. Should we depend on psr/log and suggests symfony/monolog-bundle OR depend on symfony/monolog-bundle directly?
2. Should this be implemented by adding an argument in constructors (current implementation) OR using `LoggerAwareTrait`?
3. Should we use `NullLogger` as a default logger for this PR
4. NEXT MAJOR: Should the service become mandatory? (implies depending on symfony/monolog-bundle)
5. NEXT MAJOR: IF 2 is constructor and 3 is YES, Should we shift the constructors' parameters around so the constructor parameter can be made mandatory?
6. NEXT MAJOR: IF 4 is NO: Should we use `NullLogger` as a default logger.

I feel important to mention that Points 3 and 6 imply that we might end up silently catching and discarding errors and/or exceptions, which, to me is an extremely bad idea. It like having an old time `@` in front of somewhere, except you can't even use `xdebug.scream=on` to bypass it. It makes for tricky bugs and maintenance difficulties.

IMHO:
1. symfony/monolog-bundle
2. `LoggerAwareTrait` (mostly because it's a transversal concern)
3. NO
4. YES
5. N/A
6. N/A

Once a consensus is reached, I will prepare a new implementation.
Thanks for your time and patience.

greg0ire:
1. suggest
2. `LoggerAwareTrait` + `NullLogger` in constructor (but no logger arg), and remove `NullLogger` initialization in next major
3. yes
4. yes
5. N/A
6. N/A

lemoinem:
Thanks for the feedback!

@greg0ire @core23 Here is an implementation following Gregoire and Oskar's advice.
Tests are :green_heart:, everything should be fine (:

OskarStark:
Please Check travis

Form/Type/MediaType.php
```diff
@@ -54,7 +77,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
             'context' => $options['context'],
             'empty_on_new' => $options['empty_on_new'],
             'new_on_update' => $options['new_on_update'],
-        )));
+        ), $this->logger));
```
jordisala1991: This probably needs to be changed

Form/Type/ApiMediaType.php
```diff
@@ -50,7 +73,7 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder->addModelTransformer(new ProviderDataTransformer($this->mediaPool, $this->class, array(
             'empty_on_new' => false,
-        )), true);
+        ), $this->logger), true);
```
jordisala1991: This needs to be replaced

lemoinem:
@OskarStark @jordisala1991 Last fixes applied. All :green_heart:!

OskarStark:
@sonata-project/contributors please review

greg0ire:
I remove some things from the changelog, and I think it could be made even shorter, and more directed at end users.

lemoinem:
I changed the "Added" portion of the change log to better reflect the fact that we now use NullLogger instead of rethrowing. Makes it shorter too.

I'm not sure what you mean by making an upgrade note. The fix itself is backward compatible. So nothing should be required for upgrading.

Let me know if there is anything more I can do.

Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -11,23 +11,47 @@
...
-class ServiceProviderDataTransformer implements DataTransformerInterface
+class ServiceProviderDataTransformer implements DataTransformerInterface, LoggerAwareInterface
 {
     /**
      * @var MediaProviderInterface
      */
     protected $provider;
 
     /**
+     * The logger instance.
```
OskarStark: comment is not needed

Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -11,23 +11,47 @@
...
      */
     public function __construct(MediaProviderInterface $provider)
     {
         $this->provider = $provider;
+
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * Sets a logger.
```
OskarStark: comment is not needed

Form/Type/ApiMediaType.php
```diff
@@ -34,23 +37,46 @@ class ApiMediaType extends AbstractType
     protected $class;
 
     /**
+     * The logger instance.
```
OskarStark: comment is not needed

Form/Type/ApiMediaType.php
```diff
@@ -34,23 +37,46 @@ class ApiMediaType extends AbstractType
...
      */
     public function __construct(Pool $mediaPool, $class)
     {
         $this->mediaPool = $mediaPool;
         $this->class = $class;
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * Sets a logger.
```
OskarStark: comment is not needed

Form/Type/MediaType.php
```diff
@@ -35,26 +38,49 @@ class MediaType extends AbstractType
     protected $class;
 
     /**
+     * The logger instance.
```
OskarStark: comment is not needed

Form/Type/MediaType.php
```diff
@@ -35,26 +38,49 @@ class MediaType extends AbstractType
...
      */
     public function __construct(Pool $pool, $class)
     {
         $this->pool = $pool;
         $this->class = $class;
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * Sets a logger.
```
OskarStark: comment is not needed

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -37,6 +48,19 @@ public function __construct(Pool $pool, $class, array $options = array())
         $this->pool = $pool;
         $this->options = $this->getOptions($options);
         $this->class = $class;
+
+        $this->logger = new NullLogger();
+    }
+
+    /**
+     * Sets a logger.
```
OskarStark: comment is not needed

Form/DataTransformer/ProviderDataTransformer.php
```diff
@@ -28,6 +31,14 @@ class ProviderDataTransformer implements DataTransformerInterface
     protected $options;
 
     /**
+     * The logger instance.
```
OskarStark: comment is not needed

greg0ire:
No need for an upgrade note indeed

lemoinem:
@OskarStark Useless comments removed.

OskarStark:
Great job @lemoinem 🎉

lemoinem:
Thanks! Any idea regarding when we'll have a 3.5.2 patch release including this?

PR #1250, opened 2017-04-25
Fixed some hardcoded paths related to integration with SonataEasyExtendsBundle

vyshkant:
I am targeting this branch, because my patch is fixing the bug, arisen due to the hardcoded class-path.

## Changelog

```markdown
### Fixed
- Fixed hardcoded paths to classes in `.xml.skeleton` files of config
```

## Subject

Fixed some hardroded paths, which become invalid after enabling '--namespace' option in SonataEasyExtendsBundle's last release (https://github.com/sonata-project/SonataEasyExtendsBundle/releases/tag/2.2.0).

### DIFF
--- a/Resources/config/doctrine/Gallery.mongodb.xml.skeleton
+++ b/Resources/config/doctrine/Gallery.mongodb.xml.skeleton
@@ -11,7 +11,7 @@
             xml mapping          : http://www.doctrine-project.org/projects/orm/2.0/docs/reference/xml-mapping/en
             association mapping  : http://www.doctrine-project.org/projects/orm/2.0/docs/reference/association-mapping/en
     -->
-    <document name="Application\Sonata\MediaBundle\Document\Gallery" >
+    <document name="{{ namespace }}\Document\Gallery" >
         <field fieldName="id" id="true" />
     </document>
 </doctrine-mongo-mapping>
--- a/Resources/config/doctrine/Gallery.orm.xml.skeleton
+++ b/Resources/config/doctrine/Gallery.orm.xml.skeleton
@@ -9,7 +9,7 @@
             association mapping  : http://www.doctrine-project.org/projects/orm/2.0/docs/reference/association-mapping/en
     -->
     <entity
-        name="Application\Sonata\MediaBundle\Entity\Gallery"
+        name="{{ namespace }}\Entity\Gallery"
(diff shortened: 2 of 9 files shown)


jordisala1991:
Will this work even if you dont define your custom namespace when calling easy-extends?

vyshkant:
Yes, this will work if we don't define custom `--namespace` when calling `sonata:easy-extends:generate`.

The variable `namespace` in edited skeletons contains the actual PHP namespace of the bundle like `Application\Sonata\MediaBundle` or `SomeOtherNamespace\Sonata\MediaBundle`, where `Application` is the default value and `SomeOtherNamespace` is the value of SonataEasyExtendsBundle's `--namespace` option.

jordisala1991:
This will probably require more PR on other bundles that rely on easy-extend. Classification or UserBundle for instance. This is just a remined, in case you want to help with that ones also :)

jordisala1991:
I think we normally don't add requirements update of composer.json on CHANGELOG. Correct me if I am wrong @greg0ire

vyshkant:
> This will probably require more PR on other bundles that rely on easy-extend. Classification or UserBundle for instance. This is just a remined, in case you want to help with that ones also :)
I'm not sure how fast, but I'll try to take care of them too.

vyshkant:
Let's proceed as follows: I'll fix the `sonata-project` bundles, which depend to SonataEasyExtendsBundle after you accept and merge this pull request (just to understand all the details).

greg0ire:
> I think we normally don't add requirements update of composer.json on CHANGELOG. Correct me if I am wrong @greg0ire

Does the end-user care? I don't think they do. So yeah, we should remove that part.

vyshkant:
@jordisala1991  I'have changed the commit message as you asked.

jordisala1991:
Not sure if minor or patch, @greg0ire or @OskarStark will know. Anyway this is ready to merge now, once travis passes. 

Thanks for your first contribution :)

jordisala1991:
IMO this is a patch, but not 100% sure.

vyshkant:
So I've made similar pull requests to other bundles of `sonata-project`. I did not make pull requests to non-bundles branches (such as `ecommerce`), because I'm not sure about how they work.

PR #1255, opened 2017-05-14
Fix sonata_template_box templates compatibility with twig 2.0

xaben:
I am targeting this branch, because this fixes a bug with twig 2.0.

## Changelog

```markdown

### Fixed
- fixed bug against twig 2.0 as `translationBundle` cannot be null

```

## Subject

After upgrading to twig 2.0 I got some errors from following templates. 
As I understand ```sonata_template_box``` requires 2 parameters now in order to be compliant, so I added the translation domain in this PR to address the issue.

```
- The template "SonataMediaBundle:Gallery:index.html.twig" contains an error: Using "null" for the value of node "translationBundle" of "Sonata\CoreBundle\Twig\Node\TemplateBoxNode" is not supported. You must pass a Twig_Node instance.
- The template "SonataMediaBundle:Gallery:view.html.twig" contains an error: Using "null" for the value of node "translationBundle" of "Sonata\CoreBundle\Twig\Node\TemplateBoxNode" is not supported. You must pass a Twig_Node instance.
- The template "SonataMediaBundle:Media:view.html.twig" contains an error: Using "null" for the value of node "translationBundle" of "Sonata\CoreBundle\Twig\Node\TemplateBoxNode" is not supported. You must pass a Twig_Node instance.
```

### DIFF
--- a/src/Resources/translations/SonataMediaBundle.bg.xliff
+++ b/src/Resources/translations/SonataMediaBundle.bg.xliff
@@ -427,10 +427,6 @@
                 <source>sonata_media_gallery_index</source>
                 <target>Галерия</target>
             </trans-unit>
-            <trans-unit id="sonata_template_box_media_gallery_block">
-                <source>sonata_template_box_media_gallery_block</source>
-                <target>Това блокът на медиа галерията. Свободни сте да го промените по ваш вкус. </target>
-            </trans-unit>
             <trans-unit id="list.label_description">
                 <source>list.label_description</source>
                 <target>Описание</target>
--- a/src/Resources/translations/SonataMediaBundle.de.xliff
+++ b/src/Resources/translations/SonataMediaBundle.de.xliff
@@ -423,10 +423,6 @@
                 <source>sonata_media_gallery_index</source>
                 <target>Galerie</target>
             </trans-unit>
-            <trans-unit id="sonata_template_box_media_gallery_block">
-                <source>sonata_template_box_media_gallery_block</source>
-                <target>as ist ein Galerie Media Block den Sie überschreiben können.</target>
-            </trans-unit>
             <trans-unit id="list.label_description">
                 <source>list.label_description</source>
                 <target>Beschreibung</target>
--- a/src/Resources/translations/SonataMediaBundle.en.xliff
+++ b/src/Resources/translations/SonataMediaBundle.en.xliff
@@ -423,10 +423,6 @@
                 <source>sonata_media_gallery_index</source>
                 <target>Gallery</target>
             </trans-unit>
-            <trans-unit id="sonata_template_box_media_gallery_block">
-                <source>sonata_template_box_media_gallery_block</source>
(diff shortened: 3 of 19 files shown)


Resources/views/Gallery/index.html.twig
```diff
@@ -9,7 +9,7 @@ file that was distributed with this source code.
 
 #}
 
-{% sonata_template_box 'This is the gallery index template. Feel free to override it.' %}
+{% sonata_template_box 'This is the gallery index template. Feel free to override it.' 'SonataMediaBundle'%}
```
core23: This will always call the translator, also we don't have any translation

xaben: Well on https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Resources/views/Block/block_gallery.html.twig#L19 it is already used also with Twig 2.0 the following appears in the debug toolbar:
screen shot 2017-05-14 at 8 38 10 pm

xaben: I investigated the other approach would be to merge this PR to CoreBundle:

https://github.com/sonata-project/SonataCoreBundle/pull/402

core23:
Do we still need this after the bugfix in the core bundle @xaben ?

xaben:
@core23 Depends if those string need to be translated, if not then no.

Resources/views/Gallery/index.html.twig
core23: After the fix in the core bundle. Do we still need this fix?

IMHO we should stop calling the translator. This message is for development only, so no reason for a translation.

OskarStark:
@xaben @core23 can you guys finalize this PR?

xaben:
@OskarStark I've checked the code, and as such we have the following options:

- Merge this branch, then all messages will be translated.
- Leave as is (do not merge), then the calls to `sonata_template_box` will be inconsistent (some have translations some don't).
- Change this PR, so we remove the translation from https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Resources/views/Block/block_gallery.html.twig#L19. This implies removing of the already translated string from multiple translation languages.

For me, it does not matter as long as we no longer have any error messages.

Resources/views/Gallery/index.html.twig
core23: WDYT @greg0ire @OskarStark 

Should we translate all template box texts?

OskarStark: I would say yes 👍

core23:
Just thinking about this...

Can you find the same approach like https://github.com/sonata-project/SonataMediaBundle/blob/3.x/Resources/views/Block/block_gallery.html.twig#L19. So we don't have a whole text that get translated

OskarStark:
ping @xaben

xaben:
@core23 @greg0ire I've changed the PR in order not to use Translator for Dev messages. 

Also cleanup translation files and remove the already translated string.

PR #1261, opened 2017-06-07
Remove Twig nodes from classes to compile

nicolas-grekas:
Twig nodes are template-compilation time only. There is no reason to inline them, they should be far from required all the time, isn't it?

## Changelog

Remove Twig nodes from classes to compile


fabpot:
👍 Nodes are only ever used at compile time.

greg0ire:
This will hardly get a better approval, so let's merge :sweat_smile: 
Thanks @nicolas-grekas !

PR #696, opened 2015-01-16
Fix Remove media files in preRemove instead of postRemove

srascar:
Exception is thrown by Gaufrette Filesystem
https://github.com/KnpLabs/Gaufrette/blob/master/src/Gaufrette/Filesystem.php#L161

- [x] #504
- [x] #559
- [x] #558

PR #1270, opened 2017-07-13
Fixed deleting a file if deleting throw the error.

haltaction:
I am targeting this branch, because it is patch.

## Changelog
```markdown
### Fixed
- Prevent file from being removed if an error occurred while deleting its database entry.
```

## Subject

Moved deleting a file from file system from preRemove to postRemove.
Without it first deleting file and then trying to remove record. And if record, for example, has dependecies, it not deleted, but the file will already be deleted.

### DIFF
--- a/Provider/BaseProvider.php
+++ b/Provider/BaseProvider.php
@@ -187,6 +187,13 @@ abstract class BaseProvider implements MediaProviderInterface
      */
     public function preRemove(MediaInterface $media)
     {
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function postRemove(MediaInterface $media)
+    {
         $path = $this->getReferenceImage($media);
 
         if ($this->getFilesystem()->has($path)) {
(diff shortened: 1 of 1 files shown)


jlamur:
Please edit your PR message to add a CHANGELOG. The original PR template can be found [here](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/.github/PULL_REQUEST_TEMPLATE.md).

For example:
```markdown
### Fixed
- Prevent file from being removed if an error occured while deleting its database entry.
```

jlamur:
IMO, this is not BC.

For example, if someone decided to override the `preRemove()` method to not delete the file, it would be deleted anyway. That may be the case when implementing, for example, a soft delete feature.

OskarStark:
I updated the Changelog

OskarStark:
thank you @haltaction 👍

aderuwe:
This re-introduces the bug that was fixed here: https://github.com/sonata-project/SonataMediaBundle/pull/696, which I've now run into.
Any hints?

PR #1286, opened 2017-08-30
Take client-reported mimetype into consideration for setting content …

nevstokes:
I am targeting this branch, because this fixes a bug where SVG file uploads were identified as having "text/html" Content Type.

In case of bug fix, `3.x` **MUST** be targeted.

## Changelog

```markdown
### Fixed
Client / browser reported mimetype of uploaded files was being ignored leading to misidentified media item content types for plain text files (SVG for example).
```

## Subject

Added a check for `UploadedFile` and use that class's `getClientMimeType` method to identify the file type instead of simply using `getMimeType` method of `File` which is unable to correctly identify plain text files like `text/css` or `image/svg+xml` for example and was resulting in `text/html` being used incorrectly.

### DIFF
--- a/Provider/FileProvider.php
+++ b/Provider/FileProvider.php
@@ -401,7 +401,10 @@ class FileProvider extends BaseProvider
             $media->setProviderReference($this->generateReferenceName($media));
         }
 
-        if ($media->getBinaryContent() instanceof File) {
+        if ($media->getBinaryContent() instanceof UploadedFile) {
+            $media->setContentType($media->getBinaryContent()->getClientMimeType());
+            $media->setSize($media->getBinaryContent()->getSize());
+        } elseif ($media->getBinaryContent() instanceof File) {
             $media->setContentType($media->getBinaryContent()->getMimeType());
             $media->setSize($media->getBinaryContent()->getSize());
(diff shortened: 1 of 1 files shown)


Provider/FileProvider.php
```diff
@@ -401,7 +401,10 @@ protected function doTransform(MediaInterface $media)
             $media->setProviderReference($this->generateReferenceName($media));
         }
 
-        if ($media->getBinaryContent() instanceof File) {
+        if ($media->getBinaryContent() instanceof UploadedFile) {
+            $media->setContentType($media->getBinaryContent()->getClientMimeType());
```
core23: Can the client mimetype be null?

nevstokes: `UploadedFile` defaults to a mimetype of `application/octet-stream` should the mimetype constructor argument be null.

https://github.com/symfony/http-foundation/blob/master/File/UploadedFile.php#L91

OskarStark:
Please add a testcase

OskarStark:
@nevstokes can you please finalize this PR by adding a test case?

core23:

PR #1300, opened 2017-10-01
Fix unit test warning if Twig 2.x is used

dmarkowicz:
I am targeting this branch, because it is bug fix.

## Changelog

There is no need to describe this in changelog

## To do

- [x] Make it work on PHP <= 5.5

## Subject

`Twig_TemplateInterface` is deprecated since Twig 1.12 and removed in Twig 2.0. Composer require minimum Twig 1.28 so it is ok. Fix warning in PHP 7.x platform tests where Twig 2 is used (because require PHP 7.x).
```
There was 1 warning:

1) MediaExtensionTest::testThumbnailHasAllNecessaryAttributes
Cannot stub or mock class or interface "Twig_TemplateInterface" which does not exist
```

### DIFF
--- a/Tests/Twig/Extension/MediaExtensionTest.php
+++ b/Tests/Twig/Extension/MediaExtensionTest.php
@@ -24,7 +24,7 @@ class MediaExtensionTest extends PHPUnit_Framework_TestCase
     private $provider;
 
     /**
-     * @var Twig_TemplateInterface
+     * @var Twig_Template
      */
     private $template;
 
@@ -101,7 +101,10 @@ class MediaExtensionTest extends PHPUnit_Framework_TestCase
     public function getTemplate()
     {
         if (is_null($this->template)) {
-            $this->template = $this->createMock('Twig_TemplateInterface');
+            $this->template = $this->getMockBuilder('Twig_Template')
+                                   ->disableOriginalConstructor()
+                                   ->setMethods(array('render'))
+                                   ->getMockForAbstractClass();
         }
 
         return $this->template;


dmarkowicz:
Not so trivial as i thought before. Need to rework for ancient PHP 😢

dmarkowicz:
Ready for review.

Tests/Twig/Extension/MediaExtensionTest.php
```diff
@@ -101,7 +101,12 @@ public function getProvider()
     public function getTemplate()
     {
         if (is_null($this->template)) {
-            $this->template = $this->createMock('Twig_TemplateInterface');
+            // NEXT_MAJOR: remove the check when requirement of PHP is >= 5.6
+            if (version_compare(PHP_VERSION, '5.6', '>=')) {
```
core23: Why not use something like `if (class_exists('Twig_TemplateInterface')) {`

Tests/Twig/Extension/MediaExtensionTest.php
```diff
@@ -101,7 +101,9 @@ public function getProvider()
     public function getTemplate()
     {
         if (is_null($this->template)) {
-            $this->template = $this->createMock('Twig_TemplateInterface');
+            $this->template = $this->createMock(
+                class_exists('Twig_Template') ? 'Twig_Template' : 'Twig_TemplateInterface'
```
jlamur: The `NEXT_MAJOR` comment disappeared.

dmarkowicz: I tried `class_exists` as @core23 suggested but it [fails](https://travis-ci.org/sonata-project/SonataMediaBundle/builds/281986768) on PHP < 5.6. Not sure if my fault or twig specific. I go back to previous test with `version_compare`.

dmarkowicz:
Simple and ready ☺️

PR #1307, opened 2017-10-09
Drop old versions of php/sf

jlamur:

## Changelog

```markdown
### Removed
- Support for old versions of php and Symfony.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -17,7 +17,7 @@
         }
     ],
     "require": {
-        "php": "^5.3 || ^7.0",
+        "php": "^7.1",
         "imagine/imagine": "^0.6",
         "jms/serializer-bundle": "^0.13 || ^1.0 || ^2.0",
         "knplabs/gaufrette": "^0.1.6 || ^0.2 || ^0.3",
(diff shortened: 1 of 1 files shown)


greg0ire:
Travis is all red...

jlamur:
@greg0ire it's the same bug as here: https://github.com/sonata-project/SonataAdminBundle/pull/4686

I will fix it tonight in this PR.

composer.json
```diff
@@ -27,21 +27,21 @@
...
+        "symfony/finder": "^2.8 || ^3.0",
+        "symfony/form": "^2.8 || ^3.0",
+        "symfony/http-foundation": "^2.8 || ^3.0",
+        "symfony/http-kernel": "^2.8 || ^3.0",
+        "symfony/options-resolver": "^2.8 || ^3.0",
+        "symfony/routing": "^2.8 || ^3.0",
+        "symfony/security":"^2.8 || ^3.0",
+        "symfony/templating":"^2.8 || ^3.0",
+        "symfony/translation": "^2.8 || ^3.0",
+        "symfony/validator": "^2.8 || ^3.0",
```
greg0ire: Should be ^3.2 I think

composer.json
```diff
@@ -58,15 +58,15 @@
...
+        "symfony/phpunit-bridge": "^3.3"
     },
     "suggest": {
         "liip/imagine-bundle": "^0.9",
         "rackspace/php-opencloud": "^1.6",
         "sonata-project/classification-bundle": "If you want to categorize your media items.",
         "sonata-project/doctrine-orm-admin-bundle": "^3.0",
         "sonata-project/seo-bundle": "^2.1",
-        "symfony/monolog-bundle": "^2.4",
+        "symfony/monolog-bundle": "^2.8",
```
jordisala1991: This version does not affect anything, we should change it for a description like: "If you want to log exceptions produced when dealing with images"

jlamur: Right! IMO, I should add a conflict with `"symfony/monolog-bundle": "<2.8 || >=3.0 <3.2"`. Is it the correct way to do it?

jordisala1991: Why conflict? If we detect something that makes us unable to work with some version, lets do it. But I think this is not the case here

jlamur: > If we detect something that makes us unable to work with some version, lets do it.

That's exactly what we agreed to not do when we decided to drop support of SF < 2.8.
Currently, monolog is only an optionnal dependency, and we use it when it is installed.

So, why should we not handle optionnal dependencies like "normal" ones?

greg0ire: optional dependencies might be required in the project to do something else entirely

greg0ire: But they we can't disable their usage if installed in this case, so we should indeed conflict.

jordisala1991: If you want to keep that conflict the way it is, fine by me, but monolog-bundle does not follow same version as symfony does, I am not sure on what is based that conflict requirement.

From monolog readme:

```
As of v2.4.0 of the bundle, the release cycle is de-synchronized from the framework's. It means you can just require "symfony/monolog-bundle": "~2.4" in your composer.json and Composer will automatically pick the latest version of the bundle that works with your current version of Symfony. The minimum version of Symfony2 for this workflow is 2.3.0.
```

composer.json
```diff
@@ -75,7 +75,8 @@
         "sonata-project/block-bundle": "<3.1.1 || >=4.0",
         "sonata-project/classification-bundle": "<3.0 || >= 5.0",
         "sonata-project/formatter-bundle": "<3.2",
-        "sonata-project/seo-bundle": "<2.1 || >=3.0"
+        "sonata-project/seo-bundle": "<2.1 || >=3.0",
+        "symfony/monolog-bundle": "<2.8 || >=3.0 <3.2"
```
core23: There is an operator missing

jlamur: I didn't test it but the [documentation](https://getcomposer.org/doc/04-schema.md#conflict) says this is an "and": 

> Note that when specifying ranges like `<1.0 >=1.1` in a `conflict` link, this will state a conflict with all versions that are less than 1.0 and equal or newer than 1.1 at the same time

core23: This is valid? WTF?

composer.json
jlamur: I didn't know that this bundle does not follow the same release cycle as symfony.
I changed the conflict to `< 2.4`.

jlamur:
@core23 @greg0ire @jordisala1991 Please review again and take in account this [comment](https://github.com/sonata-project/SonataMediaBundle/pull/1307#discussion_r144360958).

jordisala1991:
Thank you @jlamur !

PR #1325, opened 2017-11-16
Use FormRenderer runtime to maintain compatibility with Symfony 3.4

hanzi:
I am targeting this branch, because it fixes a bug (Admin controllers do not work with Symfony 3.4)

## Changelog

```markdown
### Fixed
- Use FormRenderer runtime to maintain compatibility with Symfony 3.4
```

## Subject

Symfony 3.4 deprecates `TwigRenderer` (TwigBridge) in favour of `FormRenderer` (Form component). While the old class is still available, it is not registered as a Twig runtime anymore. ([See this related changeset in the TwigBridge component.](https://github.com/symfony/symfony/pull/23437/files#diff-5eb0532c96734f6e23428e921225b2f0))

`MediaAdminController` and `GalleryAdminController` try to call this runtime by class name, which throws an Exception (tested with Symfony 3.4-BETA4).

```
Twig_Error_Runtime:
Unable to load the "Symfony\Bridge\Twig\Form\TwigRenderer" runtime.

  at vendor/twig/twig/lib/Twig/Environment.php:931
  at Twig_Environment->getRuntime('Symfony\\Bridge\\Twig\\Form\\TwigRenderer')
     (vendor/sonata-project/media-bundle/Controller/MediaAdminController.php:128)
  at Sonata\MediaBundle\Controller\MediaAdminController->setFormTheme(object(FormView), array('SonataDoctrineORMAdminBundle:Form:filter_admin_fields.html.twig'))
     (vendor/sonata-project/media-bundle/Controller/MediaAdminController.php:100)
  at Sonata\MediaBundle\Controller\MediaAdminController->listAction(object(Request))
  at call_user_func_array(array(object(MediaAdminController), 'listAction'), array(object(Request)))
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:153)
  at Symfony\Component\HttpKernel\HttpKernel->handleRaw(object(Request), 1)
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php:68)
  at Symfony\Component\HttpKernel\HttpKernel->handle(object(Request), 1, true)
     (vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php:202)
  at Symfony\Component\HttpKernel\Kernel->handle(object(Request))
     (web/app_dev.php:21)
  at require('/home/hanzi/Projects/gi/get-in-it/web/app_dev.php')
     (vendor/symfony/symfony/src/Symfony/Bundle/WebServerBundle/Resources/router.php:42)
```

This PR changes both controllers to use the new `FormRenderer` runtime instead.

The `method_exists` check is based on [this PR](https://github.com/symfony/symfony/pull/24064) which got merged for the same version as the PR that caused this bug (Symfony 3.4-BETA1.)

### DIFF
--- a/tests/Controller/GalleryAdminControllerTest.php
+++ b/tests/Controller/GalleryAdminControllerTest.php
@@ -121,13 +121,21 @@ class GalleryAdminControllerTest extends TestCase
     private function configureSetFormTheme($formView, $formTheme)
     {
         $twig = $this->prophesize('\Twig_Environment');
-        $twigRenderer = $this->prophesize('Symfony\Bridge\Twig\Form\TwigRenderer');
+
+        // Remove this trick when bumping Symfony requirement to 3.4+
+        if (method_exists('Symfony\Bridge\Twig\Command\DebugCommand', 'getLoaderPaths')) {
+            $rendererClass = 'Symfony\Component\Form\FormRenderer';
+        } else {
+            $rendererClass = 'Symfony\Bridge\Twig\Form\TwigRenderer';
+        }
+
+        $twigRenderer = $this->prophesize($rendererClass);
 
         $this->container->get('twig')->willReturn($twig->reveal());
 
         // Remove this trick when bumping Symfony requirement to 3.2+.
         if (method_exists('Symfony\Bridge\Twig\AppVariable', 'getToken')) {
-            $twig->getRuntime('Symfony\Bridge\Twig\Form\TwigRenderer')->willReturn($twigRenderer->reveal());
+            $twig->getRuntime($rendererClass)->willReturn($twigRenderer->reveal());
         } else {
             $formExtension = $this->prophesize('Symfony\Bridge\Twig\Extension\FormExtension');
             $formExtension->renderer = $twigRenderer->reveal();
--- a/tests/Controller/GalleryAdminControllerTest.php
+++ b/tests/Controller/GalleryAdminControllerTest.php
@@ -14,6 +14,11 @@ namespace Sonata\MediaBundle\Tests\Controller;
 use PHPUnit\Framework\TestCase;
 use Prophecy\Argument;
 use Sonata\MediaBundle\Controller\GalleryAdminController;
+use Symfony\Bridge\Twig\AppVariable;
+use Symfony\Bridge\Twig\Command\DebugCommand;
+use Symfony\Bridge\Twig\Extension\FormExtension;
+use Symfony\Bridge\Twig\Form\TwigRenderer;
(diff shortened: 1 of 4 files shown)


hanzi:
Ah sorry, missed that one.

jordisala1991:
No need to close it, just wait until we figure out the solution and apply the same here :)

hanzi:
The PR you referenced fixes this in `CRUDController`, which serves as parent class for both `MediaAdminController` and `GalleryAdminController`.

Apparently, this method exists even now and implements the same behaviour as "our" controllers.

Wouldn't it make more sense to just use the parent method instead? The only problem right now is that it's a private method. So that would have to be changed.

Otherwise we'd be copying the same code again. The solution in that PR is the same as mine, actually.

Re-opening this PR anyway, since I guess the test changes could still be used.

jordisala1991:
Can you copy the method from my PR? We need to start moving to MyClass:class instead of strings

hanzi:
@jordisala1991 done

tests/Controller/GalleryAdminControllerTest.php
```diff
@@ -121,13 +121,21 @@ private function configureSetCsrfToken($intention)
     private function configureSetFormTheme($formView, $formTheme)
     {
         $twig = $this->prophesize('\Twig_Environment');
-        $twigRenderer = $this->prophesize('Symfony\Bridge\Twig\Form\TwigRenderer');
+
+        // Remove this trick when bumping Symfony requirement to 3.4+
+        if (method_exists('Symfony\Bridge\Twig\Command\DebugCommand', 'getLoaderPaths')) {
```
jordisala1991: DebugCommand::class

hanzi: Spelling out fully-qualified class names as strings is done *all over the place* in those test classes.

Even if I change this particular instance, or all the lines I've touched, or even all instances in those particular method, the test class would actually more inconsistent style-wise.

Might I propose to leave it as is for now and update the code style in a separate PR instead?

greg0ire: @hanzi how about you do it in a separate commit, before your changes?

hanzi:
@jordisala1991 @greg0ire I have adjusted the two test methods modified in this PR to use `ClassName::class` syntax throughout.

If you're ok with this, I can create an issue for changing the rest of the code base to match. Might have time to work on that early next week.

jordisala1991:
Thank you @hanzi !

PR #1329, opened 2017-11-16
Reflect the current docs path in its usages

greg0ire:


jordisala1991:
It would be updated tonight by dev-kit

greg0ire:
Yeah but some people would like to do some work :P

PR #1335, opened 2017-11-28
Return correct size cropped image

rvanginneken:
I am targeting this branch, because the wrong size is returned for a cropped image.

## Changelog

```markdown
### Fixed
- Return size after resize in cropping flow, not just the cropped dimensions
```
## Subject

Function `getBox` is supposed to return the dimensions for given media with applied settings. However, when an image is cropped, that size was returned directly. Resizing is the next step and should also be done before returning the actual size.

### DIFF
--- a/src/Resizer/SquareResizer.php
+++ b/src/Resizer/SquareResizer.php
@@ -126,7 +126,7 @@ class SquareResizer implements ResizerInterface
             }
 
             if ($higher - $lower > 0) {
-                return new Box($lower, $lower);
+                $size = new Box($lower, $lower);
             }
         }
 
--- a/tests/Resizer/SquareResizerTest.php
+++ b/tests/Resizer/SquareResizerTest.php
@@ -57,7 +57,7 @@ class SquareResizerTest extends TestCase
     public static function getBoxSettings()
     {
         return [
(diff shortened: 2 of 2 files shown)


greg0ire:
The build failure does not seem to be related to your PR :( Not quite sure yet what broke it.

rvanginneken:
@greg0ire 
Isn't it this?

> 1) Sonata\MediaBundle\Tests\Resizer\SquareResizerTest::testGetBox with data set #0 (array(90, 90), Imagine\Image\Box Object (...), Imagine\Image\Box Object (...))
Failed asserting that 90 is identical to 100.

greg0ire:
> The build failure does not seem to be **entirely** related to your PR :( Not quite sure yet what broke it.

rvanginneken:
@greg0ire I fixed the failing test. The other failures are not this PR related I think? Do we need to wait for that to get fixed?

greg0ire:
> Do we need to wait for that to get fixed?

We litterally can't click the merge button unless they get fixed. I'll try to have a look this evening, but I think it should be easy to fix.

greg0ire:
Thanks @rvanginneken !

PR #1336, opened 2017-11-28
Always put src format in srcset

rvanginneken:
I am targeting this branch, because when defining srcsets, the src format needs to be added again.

## Changelog

```markdown
### Added
- Added automatically adding src format to srcset
```

## Subject

I just noticed that when defining your srcset formats, the src format itself wasn't printed. Now it's automatically added if it wasn't done by the user.


greg0ire:
Thanks @rvanginneken !

PR #1259, opened 2017-06-02
RemoveThumbsCommand dialog error fix

vyshkant:
I am targeting this branch, because my patch is fixing the bug existing in this branch.

Closes #1258 

## Changelog

```markdown
### Fixed
- Silent `sonata:media:remove-thumbnails` command when running this command without arguments.
```

## To do

- [x] Update the tests

## Subject

I'have fixed the call to undefined helper 'dialog'. Intead of 'dialog' the 'question' helper is used.

PR #1344, opened 2017-12-04
Remove unnecesary bundle

jordisala1991:

## Subject

The bundle was added here: https://github.com/sonata-project/SonataMediaBundle/pull/1259 without a good reason. It is making SF 4 compatibility impossible, since it won't update to symfony4 because of https://github.com/symfony/maker-bundle

To sumarize, not a good idea to add a dependency on a bundle that has no future

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,7 +23,6 @@
         "knplabs/gaufrette": "^0.1.6 || ^0.2 || ^0.3",
         "kriswallsmith/buzz": "^0.15",
         "psr/log": "^1.0",
-        "sensio/generator-bundle": "^3.1",
         "sonata-project/core-bundle": "^3.2",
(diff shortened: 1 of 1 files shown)


vyshkant:
Just for clarity.

As far as I remember, `sensio/generator-bundle` was required to use `Sensio\Bundle\GeneratorBundle\Command\Helper\QuestionHelper` and `Sensio\Bundle\GeneratorBundle\Command\Helper\DialogHelper` in commands (see https://github.com/sonata-project/SonataMediaBundle/pull/1259#issuecomment-312598646).

But then we SUDDENLY found out that these classes are not used (see https://github.com/sonata-project/SonataMediaBundle/pull/1259#discussion_r144399174).

So dependency removal obviously will not cause any problems.

PR #1346, opened 2017-12-07
Ensure that repository calculations work with UUIDs

### LINKED ISSUE #1342: Math operation with Id throws warning if Id is generated with UUID strategy
# The problem
If you use UUID strategy for MediaElement ID property in PHP 7.1 a warning is thrown because a math operation is done with a string value.

The PHP warning:
Warning: A non-numeric value encountered

Version tested: 3.4
Version possible affected: dev-master (affected code didnt change)

# Example
Extend MediaElement with your own one with UUID strategy
```
class MediaElement extends Sonata\MediaBundle\Entity\BaseMedia
{
	/**
	 * @ORM\Id
	 * @ORM\Column(type="guid")
	 * @ORM\GeneratedValue(strategy="UUID")
	 */
	protected $id;
}
```

Use it by changing config.yml
```
sonata_media:
    class:
        media: AppBundle\Entity\MediaElement
```

# Possible solution
Detect if Id is a uuid and use the first two groups for hexadeciaml to decimal conversion. Use the resulting integer for math operation.

Change this code:
https://github.com/sonata-project/SonataMediaBundle/blob/248064b65ea9adac12a33d22d3d49c1910e79e0a/src/Generator/DefaultGenerator.php#L43-L44

To smth like this:
```
$mediaId = $media->getId();
 if (preg_match("/^\{?[a-f\d]{8}-(?:[a-f\d]{4}-){3}[a-f\d]{12}\}?$/i", $mediaId)) {
     $mediaId = (int) hexdec(implode("", array_slice(explode("-", $mediaId), 0, 2)));
}
$rep_first_level = (int) ($mediaId / $this->firstLevel);
$rep_second_level = (int) (($mediaId - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
```

benjaminrau:
I am targeting this branch, because this is not a BC but affects all versions of MediaBundle.
Closes #1342 

## Changelog
```markdown
### Fixed
- Fix thrown exception if media::id is UUID on PHP 7.1
```

### Fixed
Here media::id is used for some calculations to define a repository foldername. If media::id is UUID (string) instead of integer, this calculation will throw an exception in PHP 7.1 becauce math operations on strings are not allowed.

This commit uses the first two segments of UUID and converts their hex value to decimal to use as base of calculation.

### DIFF
--- a/src/Generator/DefaultGenerator.php
+++ b/src/Generator/DefaultGenerator.php
@@ -40,8 +40,12 @@ class DefaultGenerator implements GeneratorInterface
      */
     public function generatePath(MediaInterface $media)
     {
-        $rep_first_level = (int) ($media->getId() / $this->firstLevel);
-        $rep_second_level = (int) (($media->getId() - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
+        $mediaId = $media->getId();
+        if (preg_match("/^\{?[a-f\d]{8}-(?:[a-f\d]{4}-){3}[a-f\d]{12}\}?$/i", $mediaId)) {
+            $mediaId = (int) hexdec(implode("", array_slice(explode("-", $mediaId), 0, 2)));
+        }
+        $rep_first_level = (int) ($mediaId / $this->firstLevel);
+        $rep_second_level = (int) (($mediaId - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
 
         return sprintf('%s/%04s/%02s', $media->getContext(), $rep_first_level + 1, $rep_second_level + 1);
     }


OskarStark:
Please add a testcase

jordisala1991:
This could generate wrong paths if I saved images with the old logic (not in php 7.1, but on older versions) and now this routeGeneration works different, right? (If yes, this is BC break and should target master, if not, it is not a BC break and should target 3.x)

benjaminrau:
@jordisala1991 you are right, i wasnt aware that generated paths are not persistent.
I am unsure how to fix this in a smart way. It should be generated in a different way only for new media. Any idea how to modify my PR to achieve this?

This looks quite ugly:
```

 if ($media->getCreatedAt() > (new \DateTime("2017-12-14")) && preg_match("/^\{?[a-f\d]{8}-(?:[a-f\d]{4}-){3}[a-f\d]{12}\}?$/i", $mediaId)) {
        $mediaId = (int) hexdec(implode("", array_slice(explode("-", $mediaId), 0, 2)));
}
```

bigfoot90:
Any news on this?

benjaminrau:
My fix does not ensure backwards compatibility, i will close the pull request as i think this needs to be fixed in an completely different approach.

PR #1347, opened 2017-12-07
Properly resolve paramters in application config

### LINKED ISSUE #1271: AddProviderCompilerPass parameter replacement
After the change made by #1218 we no longer have the parameter replacement working in the providers configuration.

For example I use different filesystem for dev/staging/prod and to defined that I used a parameter:

```yaml
providers:
        file:
            filesystem: %sonata_media.filesystem%
```

Now I get an error saying that the service is %sonata_media.filesystem% doesn't exist.

benjaminrau:
I am targeting this branch, because it fixes problems with latest 3.X release.

Closes #1271 

## Changelog
```markdown
### Fixed
- Fixes problem with parameter resolution in extension config
```

### Fixed
Recently the AddCompilerPass was refactored and an ugly hack was removed to resolve paramters in extension config. Now parameters (from parameters.yml) and used in config.yml are no longer replaced. With this PR the parameters are resolved again.

### DIFF
--- a/src/DependencyInjection/Compiler/AddProviderCompilerPass.php
+++ b/src/DependencyInjection/Compiler/AddProviderCompilerPass.php
@@ -85,14 +85,14 @@ class AddProviderCompilerPass implements CompilerPassInterface
                     $definition = $container->getDefinition($id);
 
                     $definition
-                        ->replaceArgument(1, new Reference($config['filesystem']))
-                        ->replaceArgument(2, new Reference($config['cdn']))
-                        ->replaceArgument(3, new Reference($config['generator']))
-                        ->replaceArgument(4, new Reference($config['thumbnail']))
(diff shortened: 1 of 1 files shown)


jordisala1991:
What happens if some of these values is not a parameter?

benjaminrau:
@jordisala1991 It is untouched then. If there is nothing to resolve the method resolveValue is just returning the input value.

OskarStark:
Can you add a test case for both cases?

benjaminrau:
@jordisala1991 I think i would need some assistance or at least somebody pointing to a direction for that

OskarStark:
https://github.com/SymfonyTest/SymfonyDependencyInjectionTest might be a good starting point

OskarStark:
Any news on this @benjaminrau ?

PR #1349, opened 2017-12-07
Drop old Symfony and PHP code

jordisala1991:

### Subject

Removing old code that never gets executed because we dropped SF 2.3 to 2.7

### DIFF
--- a/tests/Validator/FormatValidatorTest.php
+++ b/tests/Validator/FormatValidatorTest.php
@@ -15,6 +15,7 @@ use PHPUnit\Framework\TestCase;
 use Sonata\MediaBundle\Provider\Pool;
 use Sonata\MediaBundle\Validator\Constraints\ValidMediaFormat;
 use Sonata\MediaBundle\Validator\FormatValidator;
+use Symfony\Component\Validator\Context\ExecutionContext;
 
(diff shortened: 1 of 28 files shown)


tests/Validator/FormatValidatorTest.php
```diff
@@ -81,14 +68,7 @@ public function testValidateOnlyReferenceIsAllowedIfNotFormats()
...
 
-        // Prefer the Symfony 2.5+ API if available
-        if (class_exists('Symfony\Component\Validator\Context\ExecutionContext')) {
-            $contextClass = 'Symfony\Component\Validator\Context\ExecutionContext';
-        } else {
-            $contextClass = 'Symfony\Component\Validator\ExecutionContext';
-        }
-
-        $context = $this->getMockBuilder($contextClass)
+        $context = $this->getMockBuilder(ExecutionContext::class)
```
OskarStark: please use `createMock()` whenever possible

jordisala1991: Will try, but only on the code I changed, because this is not the objective of the PR

jordisala1991: Done!

greg0ire:
Thanks @jordisala1991 !

PR #1356, opened 2017-12-14
Use createMock in test files where it's possible

ZRazor:
I am targeting this branch, because it's BC.

## Changelog
```markdown
### Fixed
- Replaced `getMockBuilder` with `createMock` where it was possible
```


greg0ire:
Thanks @ZRazor !

PR #1370, opened 2018-01-12
Avoid replacing galleryHasMedias collection (improves Doctrine 2.6.0 compatibility)

gremo:
Experimental fix for Doctrine 2.6.0 compatibility, see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/issues/789.

### DIFF
--- a/src/Entity/BaseGallery.php
+++ b/src/Entity/BaseGallery.php
@@ -13,6 +13,7 @@ namespace Sonata\MediaBundle\Entity;
 
 use Doctrine\Common\Collections\ArrayCollection;
 use Sonata\MediaBundle\Model\Gallery;
+use Sonata\MediaBundle\Model\GalleryHasMediaInterface;
 
 /**
  * Bundle\MediaBundle\Entity\BaseGallery.
@@ -25,6 +26,28 @@ abstract class BaseGallery extends Gallery
     public function __construct()
     {
         $this->galleryHasMedias = new ArrayCollection();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setGalleryHasMedias($galleryHasMedias)
+    {
+        $this->galleryHasMedias = $galleryHasMedias;
+        foreach ($this->galleryHasMedias as $galleryHasMedia) {
+            $this->addGalleryHasMedias($galleryHasMedia);
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addGalleryHasMedias(GalleryHasMediaInterface $galleryHasMedia)
+    {
+        $galleryHasMedia->setGallery($this);
+        if (!$this->galleryHasMedias->contains($galleryHasMedia)) {
+            $this->galleryHasMedias->add($galleryHasMedia);
+        }
     }
 
     /**


jordisala1991:
Doctrine has a console method to generate getters and setters, why dont you use that?

jordisala1991:
This is how doctrine would generate it (on the 2.6 version):

```
    public function addGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia)
    {
        $this->galleryHasMedias[] = $galleryHasMedia;

        return $this;
    }

    public function removeGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia)
    {
        return $this->galleryHasMedias->removeElement($galleryHasMedia);
    }

    public function getGalleryHasMedias()
    {
        return $this->galleryHasMedias;
    }

    public function __construct()
    {
        $this->galleryHasMedias = new \Doctrine\Common\Collections\ArrayCollection();
    }
```

Note that there is no direct setter method.
And this is how I usually do it:

```
public function addGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia)
    {
        $this->galleryHasMedias[] = $galleryHasMedia;

        $galleryHasMedia->setGallery($this);

        return $this;
    }

    public function removeGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia)
    {
        $this->galleryHasMedias->removeElement($galleryHasMedia);

        $galleryHasMedia->setGallery();
    }

    public function getGalleryHasMedias()
    {
        return $this->galleryHasMedias;
    }

    public function __construct()
    {
        $this->galleryHasMedias = new \Doctrine\Common\Collections\ArrayCollection();
    }
```

No setter method, and I add a setGallery for the galleryHasMedia because otherwise it doesn't get registered, atleast on doctrine 2.5.x (to be checked with 2.6, but it does not hurt that's for sure)

gremo:
> Doctrine has a console method to generate getters and setters, why dont you use that?

For BC compatibility, I've changed only the Entity/Gallery class.

jordisala1991:
IMO BC should not compromise a bugfix. WDYT @sonata-project/contributors ?
Not sure about the best way to fix this but IMO adding more wrong named methods is not the best. Even if it works

gremo:
@jordisala1991 I agree that method names are wrong but I don't think they are even related to this "bug". We should put the method refactoring in another PR.

IIRC methods "adder" and "remover" are related to Symfony forms (in particular, to the collection type) and they have nothing to do with Doctrine.

jordisala1991:
You are adding another method that is wrong named, how it is supposed to not be related to this PR?

gremo:
> You are adding another method that is wrong named, how it is supposed to not be related to this PR?

@jordisala1991 not adding, overriding.

jordisala1991:
Can you modify directly the model methods then?

gremo:
> Can you modify directly the model methods then?

Why? This is related to ORM not to the model itself. We should not change the (inherited) behaviour of the PHPRC BaseGallery class.

jordisala1991:
addGalleryHasMedias is already overriden by the PHPCR class so no problem there

gremo:
@jordisala1991 yes, but https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/PHPCR/BaseGallery.php#L101 will still call the model method (which we need to change).

jordisala1991:
Lets wait for more opinions

jordisala1991:
Ping @sonata-project/contributors

gremo:
Ping...

src/Entity/BaseGallery.php
```diff
@@ -27,6 +27,27 @@ public function __construct()
         $this->galleryHasMedias = new ArrayCollection();
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function setGalleryHasMedias($galleryHasMedias)
+    {
+        foreach ($galleryHasMedias as $galleryHasMedia) {
```
core23: No clear needed if we try to replease the collection?

greg0ire: replace? ping @gremo you should answer this

core23: > replace?

AFAIK this is the default behavior in doctrine. If you want to add many childrens at once, you should create a `addAllGalleryHasMedia` method

greg0ire: Yes, I was just taking a guess at what "replease" could possibly mean :P

gremo: What I should answer?
I've just found the bug and the fix: It works with both version of Doctrine (2.5 and 2.6). But I'm waiting for some authoritative  opinion about this...

But again this is huge: think about how many people are using SonataMediaBundle with 3.4 LTS (may affect other Sonata bundles too).

greg0ire: > What I should answer?

Explain to @core23 why you don't want to call `clear()` on the collection.

gremo: Well @core23 IMHO there is no need: the collection is already up-to-date as soon as you submit the form using the admin. What clear() is supposed to do?

core23: > the collection is already up-to-date as soon as you submit the form using the admin. 

What if you use the `setGalleryHasMedias` outside the admin form? You could also do this by hand in a custom controller.

janpetersen:
See https://github.com/sonata-project/SonataMediaBundle/pull/1394

This is easy to put into your extended Gallery in your own project's model. Just overwrite the two affected methods, in case this bug takes another month to be resolved.

gremo:
@greg0ire @core23 I changed the setter to assign the `$galleryHasMedias` property. This should solve the issue about cleaning the collection, right? Let me know... it seems to work fine with both version of Doctrine.

@janpetersen I know what you mean but... do you really think that not releasing an update is a good choice? Consider how many people just ignore this issue: they will loose their data.

src/Entity/BaseGallery.php
```diff
@@ -27,6 +28,28 @@ public function __construct()
...
     }
 
+    /**
+     * {@inheritdoc}
+     */
+    public function setGalleryHasMedias($galleryHasMedias)
+    {
+        $this->galleryHasMedias = $galleryHasMedias;
+        foreach ($this->galleryHasMedias as $galleryHasMedia) {
+            $this->addGalleryHasMedias($galleryHasMedia);
```
jordisala1991: Why do you need this? Shouldn't be replaced by simply: `$galleryHasMedia->setGallery($this);` ?

gremo: Do you want to remove addGalleryHasMedias method?

jordisala1991: If we don't need it yes

PR #1378, opened 2018-01-26
Update installation documentation for Symfony Flex

kunicmarko20:
I am targeting this branch, because I am updating documentation for Symfony Flex.

## Subject

One more bundle that was hard to install following current docs, this should help people.

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -26,7 +26,25 @@ Retrieve the bundle with composer:
 
     $ composer require sonata-project/media-bundle
 
-Register these bundles in your AppKernel:
+Register these bundles in your ``bundles.php`` file:
+
+.. code-block:: php
+
+    <?php
+
+    // config/bundles.php
+
+    return [
+        //...
+        Sonata\MediaBundle\SonataMediaBundle::class => ['all' => true],
+        Sonata\EasyExtendsBundle\SonataEasyExtendsBundle::class => ['all' => true],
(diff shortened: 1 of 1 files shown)


docs/reference/installation.rst
```diff
@@ -212,19 +298,42 @@ Now that your module is generated, you can register it
...
 
 If they are not already created, you need to add specific folder to allow uploads from users:
 
 .. code-block:: bash
 
-    $ mkdir web/uploads
-    $ mkdir web/uploads/media
+    $ mkdir public/uploads
+    $ mkdir public/uploads/media
     $ chmod -R 0777 web/uploads
```
greg0ire: 1. Oh my god.
2. Should be `public/uploads` here too

kunicmarko20: Good catch :smile:

docs/reference/installation.rst
```diff
@@ -212,19 +298,42 @@ Now that your module is generated, you can register it
...
 If they are not already created, you need to add specific folder to allow uploads from users:
 
 .. code-block:: bash
 
-    $ mkdir web/uploads
-    $ mkdir web/uploads/media
-    $ chmod -R 0777 web/uploads
+    $ mkdir public/uploads
+    $ mkdir public/uploads/media
+    $ chmod -R 0777 public/uploads
```
greg0ire: I don't think we should recommend this. Maybe just add a comment instead, because the right way will depend on the installation. Something like "make sure your http user can write to this directory, using `chmod` or `setfacl`

PR #890, opened 2015-11-14
Remove 'wrap' property from GalleryBlock

core23:
### Changelog

```markdown
### Removed
 * Removed `wrap` property from `GalleryBlockService`
```

### Subject
Removes an unused property form `GalleryBlockService`.

### To do

- [x] Add an upgrade note

PR #1383, opened 2018-01-31
Manual merge

kunicmarko20:
Just approve, don't merge.

### DIFF
--- a/src/Block/GalleryBlockService.php
+++ b/src/Block/GalleryBlockService.php
@@ -96,7 +96,7 @@ class GalleryBlockService extends AbstractAdminBlockService
             'format' => false,
             'pauseTime' => 3000,
             'startPaused' => false,
-            'template' => 'SonataMediaBundle:Block:block_gallery.html.twig',
+            'template' => '@SonataMedia/Block/block_gallery.html.twig',
             'galleryId' => null,
         ]);
     }
--- a/src/Templating/Helper/MediaHelper.php
+++ b/src/Templating/Helper/MediaHelper.php
@@ -23,6 +23,9 @@ use Symfony\Component\Templating\Helper\Helper;
  * MediaHelper manages action inclusions.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @deprecated Since version 3.10, will be removed in 4.0.
+ * NEXT_MAJOR : remove this class
  */
 class MediaHelper extends Helper
 {
--- a/.flintci.yml
+++ b/.flintci.yml
@@ -1,3 +1,4 @@
 services:
+  composernormalize: true
   phpcsfixer: true
   yamllint: true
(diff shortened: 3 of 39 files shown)


src/Block/GalleryBlockService.php
```diff
@@ -96,7 +96,8 @@ public function configureSettings(OptionsResolver $resolver): void
             'format' => false,
             'pauseTime' => 3000,
             'startPaused' => false,
-            'template' => 'SonataMediaBundle:Block:block_gallery.html.twig',
+            'wrap' => true,
```
kunicmarko20: This is missing on master since last CI merge, not sure why.

src/Templating/Helper/MediaHelper.php
```diff
@@ -23,6 +23,9 @@
  * MediaHelper manages action inclusions.
  *
  * @author Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * @deprecated Since version 3.10, will be removed in 4.0.
+ * NEXT_MAJOR : remove this class
```
kunicmarko20: Should I just remove this class?

src/Block/GalleryBlockService.php
greg0ire: Does it have to do with ca9a642 ?

src/Templating/Helper/MediaHelper.php
greg0ire: Probably not in that PR, which should be kept minimal, although I think I've done that in the past…

src/Block/GalleryBlockService.php
kunicmarko20: Not sure, as it present on current `3.x` https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Block/GalleryBlockService.php#L97 also https://github.com/sonata-project/SonataMediaBundle/blob/bb60020eafa6874f214b1b129437db4e8ac89afb/src/Block/GalleryBlockService.php#L97 commit on master 18 days ago, and your commit is from 2015.

src/Templating/Helper/MediaHelper.php
kunicmarko20: ok, will leave it for now then.

src/Block/GalleryBlockService.php
kunicmarko20: I think CI fails because of this.

jordisala1991: be careful because it probably was removed on master directly. I remember this wrap attribute. can you check master?

src/Templating/Helper/MediaHelper.php
jordisala1991: Better on a cleanup pr indeed

src/Block/GalleryBlockService.php
greg0ire: It was indeed purposefully removed directly on master: https://github.com/sonata-project/SonataMediaBundle/pull/890

kunicmarko20: ah, sorry, will remove it.

PR #1339, opened 2017-11-28
Call new FOSRest API

greg0ire:
I am targeting this branch, because this is BC.

Refs #1267

## Changelog

```markdown
### Fixed
- FOSRest-related deprecations
```
## Subject

Apparently setMaxDepth was a misnomer.
See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1506

PR #1389, opened 2018-02-12
Remove compatibility with FOSRest < 2.1

jordisala1991:

## Changelog

```markdown
### Removed
- Removed compatibility with older versions of FOSRestBundle (<2.1)
```

## Subject

Since dropping a dependency is not a BC break, this is a good start, because we are doing a lot of code just to keep compat with FOSRest 1.

Those are the PR adding this code, to check if I did correctly or not: https://github.com/sonata-project/SonataMediaBundle/pull/1071 https://github.com/sonata-project/SonataMediaBundle/pull/1155 https://github.com/sonata-project/SonataMediaBundle/pull/1339

### DIFF
--- a/src/Controller/Api/GalleryController.php
+++ b/src/Controller/Api/GalleryController.php
@@ -16,7 +16,6 @@ use FOS\RestBundle\Controller\Annotations\QueryParam;
 use FOS\RestBundle\Controller\Annotations\View;
 use FOS\RestBundle\Request\ParamFetcherInterface;
 use FOS\RestBundle\View\View as FOSRestView;
-use JMS\Serializer\SerializationContext;
 use Nelmio\ApiDocBundle\Annotation\ApiDoc;
 use Sonata\DatagridBundle\Pager\PagerInterface;
 use Sonata\MediaBundle\Model\GalleryHasMediaInterface;
@@ -81,6 +80,7 @@ class GalleryController
      * @QueryParam(name="page", requirements="\d+", default="1", description="Page for gallery list pagination")
      * @QueryParam(name="count", requirements="\d+", default="10", description="Number of galleries by page")
      * @QueryParam(name="enabled", requirements="0|1", nullable=true, strict=true, description="Enabled/Disabled galleries filter")
+     * @QueryParam(name="orderBy", map=true, requirements="ASC|DESC", nullable=true, strict=true, description="Order by array (key is field, value is direction)")
      *
      * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
@@ -90,20 +90,6 @@ class GalleryController
      */
     public function getGalleriesAction(ParamFetcherInterface $paramFetcher)
     {
-        $orderByQueryParam = new QueryParam();
-        $orderByQueryParam->name = 'orderBy';
-        $orderByQueryParam->requirements = 'ASC|DESC';
-        $orderByQueryParam->nullable = true;
-        $orderByQueryParam->strict = true;
-        $orderByQueryParam->description = 'Query groups order by clause (key is field, value is direction)';
-        if (property_exists($orderByQueryParam, 'map')) {
-            $orderByQueryParam->map = true;
-        } else {
-            $orderByQueryParam->array = true;
-        }
-
-        $paramFetcher->addParam($orderByQueryParam);
-
         $supportedCriteria = [
             'enabled' => '',
         ];
@@ -434,26 +420,12 @@ class GalleryController
             $gallery->addGalleryHasMedias($galleryHasMedia);
             $this->galleryManager->save($gallery);
 
+            $context = new Context();
+            $context->setGroups(['sonata_api_read']);
+            $context->enableMaxDepth();
+
             $view = FOSRestView::create($galleryHasMedia);
-
-            // BC for FOSRestBundle < 2.0
-            if (method_exists($view, 'setSerializationContext')) {
-                $serializationContext = SerializationContext::create();
-                $serializationContext->setGroups(['sonata_api_read']);
-                $serializationContext->enableMaxDepthChecks();
-                $view->setSerializationContext($serializationContext);
-            } else {
-                $context = new Context();
-                $context->setGroups(['sonata_api_read']);
-
-                // NEXT_MAJOR: simplify when dropping FOSRest < 2.1
-                if (method_exists($context, 'disableMaxDepth')) {
-                    $context->disableMaxDepth();
-                } else {
-                    $context->setMaxDepth(0);
-                }
-                $view->setContext($context);
-            }
+            $view->setContext($context);
 
             return $view;
         }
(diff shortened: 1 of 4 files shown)


src/Controller/Api/MediaController.php
```diff
@@ -406,24 +392,11 @@ protected function handleWriteMedium(Request $request, MediaInterface $media, Me
...
-
-                // NEXT_MAJOR: simplify when dropping FOSRest < 2.1
-                if (method_exists($context, 'disableMaxDepth')) {
-                    $context->disableMaxDepth();
-                } else {
-                    $context->setMaxDepth(0);
-                }
-                $view->setContext($context);
-            }
+            $context = new Context();
```
kunicmarko20: Is this even needed, can't it be replaced with `View` Annotation?

kunicmarko20:
This looks more like a minor than a patch

src/Controller/Api/MediaController.php
jordisala1991: I am not sure about that, It was originally implemented like that. Do you know a better way? I am not a user of FOSRest atm, just doing the cleanup

kunicmarko20: If I am correct, annotation [here](https://github.com/jordisala1991/SonataMediaBundle/blob/79952066c3f9dac4d227f05ac55dca943b196b44/src/Controller/Api/GalleryController.php#L85 )

`@View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)`

Can replace:

```
            $view = FOSRestView::create($gallery);
            $context = new Context();
            $context->setGroups(['sonata_api_read']);
            $context->disableMaxDepth();
            $view->setContext($context);
            return $view;
```

to just 

```
return $gallery;
```

Only change `serializerEnableMaxDepthChecks` to `false`.

kunicmarko20: Hopefully, someone else is using FOSRest and could confirm this.

jordisala1991: But on our case we are doing this only on the `if`, on the else part we are returning a form

kunicmarko20: That does not affect form, form is not valid it returns validation errors, maybe I am using FOSRest wrong but I have it like that everywhere :smile:

jordisala1991: is this what you had in mind?

kunicmarko20: yes, but someone needs to test/confirm this will work.

src/Controller/Api/GalleryController.php
```diff
@@ -412,6 +398,8 @@ public function deleteGalleryAction($id)
     /**
      * Write a GalleryHasMedia, this method is used by both POST and PUT action methods.
      *
+     * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=false)
```
jordisala1991: If this is the default value: `serializerEnableMaxDepthChecks=false` maybe we can remove it also?

kunicmarko20:
Tests are failing because of change to View Annotation, will try to test this later today and confirm that we can go this way before you start fixing the tests.

jordisala1991:
Okay, I will fix them tonight probably and I will create PR for all bundles using FOSRest

kunicmarko20:
It works,

example (as in my comments):
https://github.com/kunicmarko20/fos-rest-test/blob/master/src/Controller/LabelController.php#L21

tests:
https://github.com/kunicmarko20/fos-rest-test/blob/master/tests/Controller/LabelControllerTest.php

travis passing:
https://travis-ci.org/kunicmarko20/fos-rest-test

The only thing I encountered that docs aren't good, so you could probably update that. [Here](https://sonata-project.org/bundles/media/3-x/doc/reference/api.html#setup) it says: 
```
sensio_framework_extra:
    view:    { annotations: false }
    router:  { annotations: true }
    request: { converters: true }
```

but view should be `true` and that is the default value so that part can even be removed. I also needed to add rule for `format_listener` [here](https://github.com/kunicmarko20/fos-rest-test/blob/master/config/packages/fos_rest.yaml#L5).

jordisala1991:
Perfect, tests fixed too! :) Thank you for your time

src/Controller/Api/GalleryController.php
```diff
@@ -520,6 +486,8 @@ protected function getMediaManager()
     /**
      * Write a Gallery, this method is used by both POST and PUT action methods.
      *
+     * @View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
```
jordisala1991: Wait, @kunicmarko20  because this method is protected and called from the public methods, that's probably why it was created like that, I think I need to revert this change

jordisala1991: I have decided to revert this, because it was used on a protected method, and will probably break BC that a method that returns a view, suddenly starts returning a media.

Let's not create any issue with this change. wdyt?

jordisala1991:
If the code is clear enough to use the annotation, I will change to that ( like notificationbundle ). For cases like classification, here or pagebundle I will go for the code option

kunicmarko20:
Well, it is just adding serialization group that is the same for every public method that calls that protected method, I am maybe missing something but that looks clear to me.

jordisala1991:
I guess it is because I don't have the necessary knowledge about FOSRest

kunicmarko20:
Ping @sonata-project/contributors, anyone using FOSRest and can help with this?

jordisala1991:
I have doubts on when to use the annotation and when to use the php to create the view.
As far as I can see, the php code to create the view is only needed if you want to change the response code. Otherwise, you should always use annotations. Is that right @kunicmarko20 ?

kunicmarko20:
You can change the status code in annotation also. Here is an example of how I use it with forms:

```
    /* *
     * @Rest\View(
     *  serializerGroups={
     *      "userComplete"
     *  },
     *  statusCode=Response::HTTP_CREATED
     * )
     */
    public function postAction(Request $request, UserService $userService)
    {
        $form = $this->createForm(UserType::class, $userDTO = new UserDTO());

        $form->submit($request->request->all());

        if ($form->isValid()) {
            return $userService->create($userDTO);
        }

        return $form;
    }
```

Using annotation looks cleaner to me than writing whole context and that stuff, but I do not want to block this so we can just leave it as is.

jordisala1991:
Yes, you can change the code, but we have some cases where an `if` changes the status code. You can do that with annotations?

kunicmarko20:
Don't think so but you can just do `return View::create($user, 201)` etc.

jordisala1991:
We clarified a little bit how to deal with the api. Conclusions:

- Not all public methods needs to return a View. If there is no view, there is a listener on FOSRest that will wrap the object into a View.
- It is preferable to use annotations to define Views, unless you have to deal with special cases and status codes.
- We won’t change the protected methods to use annotations since it could be BC break. Maybe it could be done on master if it doesn’t conflict with point 2.
- In any case, why we want to use views with code or annotations if there is a listener that can deal with that? The reason is to define our custom serializedGroups when dealing with classes, and other methods like: `enableMaxDepth`.

This PR should be ready with all that in mind. Please review @sonata-project/contributors

kunicmarko20:
Could you also update the docs per my [comment](https://github.com/sonata-project/SonataMediaBundle/pull/1389#issuecomment-365391737)

jordisala1991:
Rebased, and docs fixed too! Should be good to go.

jordisala1991:
Can we get this merged?

greg0ire:
Sure

PR #1345, opened 2017-12-05
Selected gallery context remained untranslated in dropdown

janpetersen:
I am targeting this branch, because it's just a missing translation.

## Changelog

```markdown
### Fixed
- Selected gallery context is now translated just like in the dropdown.
```

## Subject

This fixes a small translation bug: The name of the selected gallery context remained untranslated, although the options in the dropdown list are being translated correctly.

PR #1394, opened 2018-02-19
Fixed Bug that would clear a Gallery

janpetersen:
I am targeting this branch, because I use it.

## Changelog

```markdown
### Fixed
- Fixed Bug that would clear a Galery
- Selected gallery context remained untranslated in dropdown
```

## Subject

When saving an existing gallery, all assigned images were removed:

DELETE FROM media__gallery_media WHERE gallery_id = 1;

This was due to replacing the PersistentCollection with a fresh new ArrayCollection, probably without proper cascade=persist on Doctrine level.


src/Model/Gallery.php
```diff
@@ -144,8 +144,6 @@ public function getDefaultFormat()
      */
     public function setGalleryHasMedias($galleryHasMedias)
     {
-        $this->galleryHasMedias = new ArrayCollection();
-
```
greg0ire: Shouldn't the collection be cleared first? There might be things to remove in there.

janpetersen: I don't know why we should clear the collection first.

Deleting an image from the Gallery works just fine.

greg0ire: Because anyone might call `setGalleryHasMedias` with any custom code expect a replacement to happen, not a merge.

src/Resources/views/GalleryAdmin/list.html.twig
```diff
@@ -18,7 +18,7 @@ file that was distributed with this source code.
                 {{ "label.select_context"|trans({}, 'SonataMediaBundle') }}
                 <div class="btn-group">
                     <button type="button" class="btn btn-default dropdown-toggle" data-toggle="dropdown">
-                        <strong class="text-info">{{ persistent_parameters.context }}</strong> <span class="caret"></span>
+                        <strong class="text-info">{{ persistent_parameters.context|trans({}, 'SonataMediaBundle') }}</strong> <span class="caret"></span>
```
jordisala1991: How is this related to the fix?

src/Model/Gallery.php
janpetersen: Why would anyone want to replace all images of an existing Gallery with a complete set of new images? In that case, he should better add a custom method like "replaceGalleryHasMedia", since he most probably will have the Gallery model extended anyways.

This issue just removed all images from my local development instance of a website. I think you should focus on fixing this before someone experiences some real trouble.

src/Resources/views/GalleryAdmin/list.html.twig
janpetersen: Sorry, I did not see that. It was part of a pull request from December that I thought were already merged... https://github.com/sonata-project/SonataMediaBundle/pull/1345 ?

src/Model/Gallery.php
greg0ire: > Why would anyone want to replace all images of an existing Gallery with a complete set of new images? In that case, he should better add a custom method like "replaceGalleryHasMedia", since he most probably will have the Gallery model extended anyways.

I don't know, you don't know, let's not take any chances, ok? I don't know why you insist on removing this, you or @gremo? The `Collection` interface has a `clear` method: https://github.com/doctrine/collections/blob/master/lib/Doctrine/Common/Collections/Collection.php#L43

Just call, let me approve this PR, and get that bug fixed.

src/Resources/views/GalleryAdmin/list.html.twig
greg0ire: Do you know how to get rid of this with `git rebase -i`?

src/Model/Gallery.php
greg0ire: Oh sorry I get it now… not sure how to solve this.

gremo:
Why we don't talk about this PR in the original https://github.com/sonata-project/SonataMediaBundle/pull/1370

PR #1403, opened 2018-02-25
NotificationBundle is an optional dependency

jordisala1991:

## Changelog

```markdown
### Removed
- Removed `SonataNotificationBundle` as a required dependency
```

## Subject

It was added here (on composer.json and code): 

PR: https://github.com/sonata-project/SonataMediaBundle/pull/151/files
Commit: https://github.com/sonata-project/SonataMediaBundle/commit/fa34e9e9c93c2a3ea6a7aab02463eae78c5c7f82

It was always made optional (not loaded if bundle is not present), so no need to make it a hard requirement on composer


greg0ire:
Thanks @jordisala1991 !

PR #1412, opened 2018-03-10
Fix outdated twig docs

covex-nn:
I am targeting this branch, because this is docs.

## Subject

`@SonataMedia/Form/media_widgets.html.twig` template must be added to `twig.form_themes`, not to `twig.form.resources`


greg0ire:
Thanks @covex-nn

PR #1398, opened 2018-02-20
Composer improvements

jordisala1991:
I am targeting this branch, because this is BC.

## Changelog

```markdown
### Added
- Added compatibility with Gaufrette `^0.4` and `^0.5`
- Added compatibility with Buzz `^0.16`

### Removed
- Removed compatibility with Gaufrette `^0.1`and `^0.2`
- Removed compatibility with JMS serializer `^0.13`
```

## Subject

A few changes on composer, this was by far the hardest repo to update:

- Up some sonata dependencies
- Add missing symfony dependencies
- Replace `security` by `security-core` on required section and `security-csrf` on required-dev section (only used in tests)
- Removed deprecated Amazon dependency: only used in tests.
- Upgrade Gaufrette (allow 0.4 and 0.5 and remove 0.1 and 0.2) 
- Upgrade serializer (remove 0.13)
- Upgrade Buzz (add 0.16)
- Upgrade phpunit-bridge

There is a missing thing:

- Allow aws ^3.0. This will be done in another PR because it will need some update on our code and it was too much already for this PR.

PR #1417, opened 2018-04-16
Update dependencies

covex-nn:
I am targeting this branch, because this is a BC patch.

## Changelog

```markdown
### Added
- Added dependency on symfony/twig-bundle and symfony/security-bundle
```

## Subject

This bundle requires `twig` and `security.authorization_checker` services during container compilation in `GlobalVariablesCompilerPass` and `SecurityContextCompilerPass` accordingly. This PR allows these services to be available.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,10 +48,12 @@
         "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
         "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
         "symfony/routing": "^2.8 || ^3.2 || ^4.0",
+        "symfony/security-bundle": "^2.8 || ^3.2 || ^4.0",
         "symfony/security-core": "^2.8 || ^3.2 || ^4.0",
         "symfony/templating": "^2.8 || ^3.2 || ^4.0",
         "symfony/translation": "^2.8 || ^3.2 || ^4.0",
         "symfony/twig-bridge": "^2.8 || ^3.2 || ^4.0",
+        "symfony/twig-bundle": "^2.8 || ^3.2 || ^4.0",
         "symfony/validator": "^2.8 || ^3.2 || ^4.0",
         "twig/twig": "^1.35 || ^2.0"
     },


kunicmarko20:
IIRC changelog is not needed for this.

greg0ire:
It depends if this is pedantic or not (warrants a release or not?)

composer.json
```diff
@@ -48,10 +48,11 @@
         "symfony/http-kernel": "^2.8 || ^3.2 || ^4.0",
         "symfony/options-resolver": "^2.8 || ^3.2 || ^4.0",
         "symfony/routing": "^2.8 || ^3.2 || ^4.0",
-        "symfony/security-core": "^2.8 || ^3.2 || ^4.0",
```
greg0ire: Do we really not reference any of the classes in that package in our code?

composer.json
```diff
@@ -80,8 +81,7 @@
         "sonata-project/formatter-bundle": "^3.4",
         "sonata-project/notification-bundle": "^3.3",
         "sonata-project/seo-bundle": "^2.5",
-        "symfony/phpunit-bridge": "^4.0",
-        "symfony/security-csrf": "^2.8 || ^3.2 || ^4.0"
```
greg0ire: Why are you removing this?

kunicmarko20:
based on what @jordisala1991 did, eg https://github.com/sonata-project/SonataAdminBundle/pull/5010 I think it is pedantic

composer.json
covex-nn: Because only `symfony/security-bundle` should (have been?) added in #1398, and security-csrf and security-core are part of security-bundle

greg0ire: Ok I'll let @jordisala1991 do the review then :+1:

kunicmarko20:
Test blew up

covex-nn:
Tests are failing even if i run phpunit against current 3.x branch

kunicmarko20:
ah, forgot about that, that is because of new stuff in AdminBundle

covex-nn:
I added this lines to `Sonata\MediaBundle\Tests\Controller\GalleryAdminControllerTest::configureCRUDController`:

```php
$templateRegistry = $this->prophesize(TemplateRegistryInterface::class);
$this->admin->getCode()->willReturn('admin_code');
$this->container->get('admin_code.template_registry')->willReturn($templateRegistry);
```

and tests in `GalleryAdminControllerTest` were passed. @kunicmarko20 is this a new stuff, that you mentioned?

kunicmarko20:
Yes, that was released with new version of admin bundle and now media, formatter and classification. Since you already found the fix, could you make a PR, please?

composer.json
jordisala1991: https://github.com/sonata-project/SonataMediaBundle/blob/772d0b78fae60b5ad27f2dfc6db47928817b4fe5/src/Security/RolesDownloadStrategy.php#L16

composer.json
jordisala1991: https://github.com/sonata-project/SonataMediaBundle/blob/100ecd16c8a5eae3dc8183949f0b88a325cbd40e/tests/Controller/GalleryAdminControllerTest.php#L35

composer.json
jordisala1991: Is this an exception? because twig-bundle includes twig-bridge

And framework-bundle includes the 2 http components

IMO we should follow a clear way to declare things on composer.json. To me the rule was, add everything that is referenced on the code explicitly (correct me if im wrong @greg0ire)

To me it makes sense, because sometimes you can say I only want security-bundle 2.8 at minimum. Then you do composer update with prefer-lowest and you can end with security-core < 2.8 because you didnt fix that package too.

wdyt?

covex-nn: Fixed

PR #1418, opened 2018-04-16
Fix tests

covex-nn:

## Subject

With new version of admin bundle and now media bundle is failing


kunicmarko20:
We should rewrite a lot of stuff, we need a list 😅

jordisala1991:
Thank you @covex-nn

PR #1431, opened 2018-05-22
Connection to S3 without credentials but with IAM role

bdejacobet:
I am targeting this branch  3.x, because it used for backward compatible PRs.

## Changelog
```Markdown
### Added
- S3 Connection without use of credentials (access and secret keys) but with IAM roles (Amazon instances must have been configured with it). 
```

## Subject

To connect to an Amazon S3, you can use IAM role (cf. https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/guide_credentials.html#instance-profile-credentials). But sonata media always set credentials has an array with secret and acces keys.
Sonata admin could send a credential set at `false` if  `filesystem.s3.accessKey` and `filesystem.s3.secretKey` are `null` in sonata_media configuration. So aws_sdk set an empty Credentials object (cf. https://github.com/aws/aws-sdk-php/blob/master/src/ClientResolver.php#L405)  and use IAM roles.

### DIFF
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -458,16 +458,22 @@ class SonataMediaExtension extends Extension implements PrependExtensionInterfac
             ;
 
             if (3 === $config['filesystem']['s3']['sdk_version']) {
-                $container->getDefinition('sonata.media.adapter.service.s3')
-                ->replaceArgument(0, [
-                    'credentials' => [
-                        'secret' => $config['filesystem']['s3']['secretKey'],
-                        'key' => $config['filesystem']['s3']['accessKey'],
-                    ],
+                $arguments = [
+                    'credentials' => false,
                     'region' => $config['filesystem']['s3']['region'],
                     'version' => $config['filesystem']['s3']['version'],
-                ])
-            ;
+                ];
+
+                if (isset($config['filesystem']['s3']['secretKey'], $config['filesystem']['s3']['accessKey'])) {
+                    $arguments['credentials'] = [
+                        'secret' => $config['filesystem']['s3']['secretKey'],
+                        'key' => $config['filesystem']['s3']['accessKey'],
+                    ];
+                }
+
+                $container->getDefinition('sonata.media.adapter.service.s3')
+                    ->replaceArgument(0, $arguments)
+                ;
             } else {
                 $container->getDefinition('sonata.media.adapter.service.s3')
                     ->replaceArgument(0, [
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -455,12 +455,18 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
             ;
 
             if (3 === $config['filesystem']['s3']['sdk_version']) {
-                $container->getDefinition('sonata.media.adapter.service.s3')
-                ->replaceArgument(0, [
-                    'credentials' => [
+                if (isset($config['filesystem']['s3']['secretKey']) && isset($config['filesystem']['s3']['secretKey'])) {
```
kunicmarko20: you have the same isset 2 times?

bdejacobet: my bad, I test `secretKey`and `accessKey`

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -455,12 +455,18 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
             ;
 
             if (3 === $config['filesystem']['s3']['sdk_version']) {
-                $container->getDefinition('sonata.media.adapter.service.s3')
```
kunicmarko20: What about writing something like this:

```
$arguments = [
    'credentials' => false,
    'region' => $config['filesystem']['s3']['region'],
    'version' => $config['filesystem']['s3']['version'],
];

if (isset($config['filesystem']['s3']['secretKey'], $config['filesystem']['s3']['accessKey'])) {
    $arguments['credentials'] = [
        'secret' => $config['filesystem']['s3']['secretKey'],
        'key' => $config['filesystem']['s3']['accessKey'],
    ];
}

$container->getDefinition('sonata.media.adapter.service.s3')
    ->replaceArgument(0, $arguments);
```
?

bdejacobet: You're right, it's more readable

kunicmarko20:
The only thing left are the tests

bdejacobet:
@kunicmarko20 @OskarStark  testcase added

kunicmarko20:
tests are still failing, please fix that.

greg0ire:
It does fail because of deprecations, and this should absolutely be fixed it it isn't already. That's giving me  _déjà vu_

bdejacobet:
hello @kunicmarko20 @OskarStark, I rebase my PR with latest comits and now tests are green

bdejacobet:
hello @OskarStark, can you review this PR please? thx

PR #1432, opened 2018-05-23
Make providers, metadata, and managers services public

enekochan:
I am targeting this branch, because is a patch and backwards compatible.

Closes #1379

## Changelog

```markdown
### Fixed
- `sonata.media.manager.media` is now `public`
- `sonata.media.manager.gallery` is now `public`
- `sonata.media.metadata.proxy` is now `public`
- `sonata.media.metadata.amazon` is now `public`
- `sonata.media.metadata.noop` is now `public`
- `sonata.media.provider.image` is now `public`
- `sonata.media.provider.file` is now `public`
- `sonata.media.provider.youtube` is now `public`
- `sonata.media.provider.dailymotion` is now `public`
- `sonata.media.provider.vimeo` is now `public`
```

## Subject

Set providers, metadata and manager services public so applications using SonataMediaBundle can upgrade to Symfony 4.0.

PR #1433, opened 2018-05-23
Stop using Sonata\AdminBundle\Controller\CRUDController::render and use renderWithExtraParams

enekochan:
I am targeting this branch, because is Deprecation and BC.

## Changelog

```markdown
### Changed
- Make `Sonata\MediaBundle\Controller\MediaAdminController` use `renderWithExtraParams`
instead of the deprecated `render` from `Sonata\AdminBundle\Controller\CRUDController`
- Make `Sonata\MediaBundle\Controller\GalleryAdminController` use `renderWithExtraParams`
instead of the deprecated `render` from `Sonata\AdminBundle\Controller\CRUDController`
```

## Subject

Stop using Sonata\AdminBundle\Controller\CRUDController::render and use renderWithExtraParams

### DIFF
--- a/tests/Controller/Api/GalleryControllerTest.php
+++ b/tests/Controller/Api/GalleryControllerTest.php
@@ -124,6 +124,9 @@ class GalleryControllerTest extends TestCase
         $this->assertSame([$media], $gController->getGalleryMediasAction(1));
     }
 
+    /**
+     * @group legacy
+     */
     public function testPostGalleryMediaGalleryhasmediaAction()
     {
         $media = $this->createMock(MediaInterface::class);
@@ -184,6 +187,9 @@ class GalleryControllerTest extends TestCase
         $this->assertSame(400, $view->getResponse()->getStatusCode(), 'Should return 400');
     }
 
+    /**
+     * @group legacy
+     */
     public function testPutGalleryMediaGalleryhasmediaAction()
     {
         $media = $this->createMock(MediaInterface::class);
--- a/src/Controller/GalleryAdminController.php
+++ b/src/Controller/GalleryAdminController.php
@@ -35,7 +35,7 @@ class GalleryAdminController extends Controller
         $parameters['media_pool'] = $this->get('sonata.media.pool');
         $parameters['persistent_parameters'] = $this->admin->getPersistentParameters();
 
-        return parent::render($view, $parameters, $response);
+        return parent::renderWithExtraParams($view, $parameters, $response);
     }
 
     /**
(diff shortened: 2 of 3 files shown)


greg0ire:
The build fails because someone (@supersmile2009 ?) introduced a new deprecation but did not mark the tests that use the deprecated paths with the `@group legacy` annotation I think.

enekochan:
Once this is approved I think #1432 will pass too because the failing test is just the same.

tests/Controller/Api/GalleryControllerTest.php
```diff
@@ -124,6 +124,9 @@ public function testGetGalleryMediaAction()
         $this->assertSame([$media], $gController->getGalleryMediasAction(1));
     }
 
+    /**
+     * @group legacy
+     */
```
greg0ire: @supersmile2009 please review. Is it right to mark this as legacy? Are we testing the legacy path here? Are there tests for the non-legacy path?

enekochan: For what I've seen the test is creating a mock gallery object that implements `GalleryInterface` which in fact has the deprecated method `addGalleryHasMedias` that makes the test not pass.

There is an if...else in `GalleryController::handleWriteGalleryhasmedia` that checks if it has to call `addGalleryHasMedias` (deprecated) or `addGalleryHasMedia` (the new method) depending on the interface the gallery implements (`GalleryInterface` or `GalleryMediaCollectionInterface`).

```
// NEXT_MAJOR: remove this if/else block. Just call `$gallery->addGalleryHasMedia($galleryHasMedia);`
if ($gallery instanceof GalleryMediaCollectionInterface) {
    $gallery->addGalleryHasMedia($galleryHasMedia);
} else {
    $gallery->addGalleryHasMedias($galleryHasMedia);
}
$this->galleryManager->save($gallery);
```

So if the test is testing a legacy path it should be marked as legacy.
Otherwise if the test is intended to test the new `addGalleryHasMedia` we should change the mock gallery object to be an instance of `GalleryMediaCollectionInterface` so the if in `GalleryController` runs `addGalleryHasMedia`. I'm not sure how to do this.

May be the solution for now is to leave the test as legacy an once the `addGalleryHasMedias` method is removed (along with the if in the controller) also remove the legacy tag because it will call the new `addGalleryHasMedia` method.

greg0ire: Ideally there should be 2 tests, but your changes can be merged IMO

greg0ire:
Thanks @enekochan!

greg0ire:
I edited the changelog to be more compact and readable.

PR #1393, opened 2018-02-18
LiipImagineThumbnail thumbnail path resolve

schauer:
I continue the PR from cengizhancaliskan https://github.com/sonata-project/SonataMediaBundle/pull/1242
"I am targeting this branch, because , The LiipImagine dependency being used is outdated and no longer maintained. Actually i fixed relative image path."

Closes #994

## Changelog

```markdown


### Fixed
- LiipImagine generatePublicUrl updated to work with latest version of that bundle
- Relative path when path it is already an url

```

PR #1437, opened 2018-05-25
3.13.1

greg0ire:
Approve and squash merge or rebase

### DIFF
--- a/src/Controller/MediaController.php
+++ b/src/Controller/MediaController.php
@@ -110,7 +110,7 @@ class MediaController extends Controller
      *
      * @return Response
      *
-     * @deprecated since 3.x, to be removed in 4.0.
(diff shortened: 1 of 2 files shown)


src/Controller/MediaController.php
```diff
@@ -110,7 +110,7 @@ public function viewAction($id, $format = MediaProviderInterface::FORMAT_REFEREN
      *
      * @return Response
      *
-     * @deprecated since 3.x, to be removed in 4.0.
+     * @deprecated since 3.12, to be removed in 4.0.
```
greg0ire: Check on your own if you think this is a mistake.

covex-nn: It was deprecated since 3.12, see #1393

PR #1439, opened 2018-05-26
Make sonata.media.manager.category service alias public

enekochan:
I am targeting this branch because is a BC patch.

## Changelog

```markdown
### Changed
- Make service alias `sonata.media.manager.category` public
```

## Subject

The `sonata.media.manager.category` service alias is created from service `sonata.media.manager.category.default` when SonataClassificationBundle is enabled. This service alias is retrieved from the container and checked for its existence in several places so it will fail in Symfony 4.0.


src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -131,7 +131,8 @@ public function load(array $configs, ContainerBuilder $container)
 
         if ($this->isClassificationEnabled($config)) {
             $loader->load('category.xml');
-            $container->setAlias('sonata.media.manager.category', $config['category_manager']);
+            $categoryManager = $container->setAlias('sonata.media.manager.category', $config['category_manager']);
```
greg0ire: Looks like `setAlias` does not behave consistently for all Travis jobs

enekochan: Yes, Symfony 2.8 and 3.3 don't return the alias with `setAlias`. Fixing it right now.

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already pushed. Don't forget the `--force` option otherwise git will try to merge both things together.

PR #1451, opened 2018-07-05
Add getters to FileProvider

tifabien:
I am targeting this branch, because it adds helpful getters to FileProvider class.

## Changelog

```markdown
### Added
- Added getters for `allowedExtensions` and  `allowedMimeTypes` properties
```

Useful when we inject the provider in twig.

### DIFF
--- a/src/Provider/FileProvider.php
+++ b/src/Provider/FileProvider.php
@@ -89,7 +89,7 @@ class FileProvider extends BaseProvider
     }
 
     /**
-     * @return array
+     * @return string[]
      */
     public function getAllowedExtensions()
     {
@@ -97,7 +97,7 @@ class FileProvider extends BaseProvider
     }
 
     /**
-     * @return array
+     * @return string[]
      */
     public function getAllowedMimeTypes()
     {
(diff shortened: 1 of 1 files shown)


OskarStark:
Could you please add a test case and do we need to change/extend an Interface?

kunicmarko20:
There is no interface and these properties are only in FileProvider, about the Tests, do we really need to test getters? :smile:

src/Provider/FileProvider.php
```diff
@@ -88,6 +88,22 @@ public function getReferenceFile(MediaInterface $media)
         return $this->getFilesystem()->get($this->getReferenceImage($media), true);
     }
 
+    /**
+     * @return array
```
greg0ire: please use the `itemType[]` notation

OskarStark:
> There is no interface and these properties are only in FileProvider, about the Tests, do we really need to test getters? 😄

I was on mobile yesterday and I am fine without the tests, but I really do getter tests because of misconfigured setters mostly ;-)

After @tifabien applied @greg0ire 's comment, we are ready to merge 👍

PR #1453, opened 2018-07-07
 Fixed wrong PHPDoc 

core23:

## Changelog

```markdown
### Fixed
- Fixed case typo for \DateTime 
```

## Subject
Some PHPDoc fixes and one typo fix.


kunicmarko20:
Does this really need a changelog? It looks more like a `pedantic` then a fix, but maybe I am wrong.

core23:
Not sure, because I actually changed code from `\Datetime` to `\DateTime`

greg0ire:
It's pedantic, because it will not change anything for the end user, will it?

PR #1455, opened 2018-07-09
Add show_mosaic_button parameter in config file

tifabien:
I am targeting this branch, because it could be useful to configure individually the show_mosaic_button option in media and gallery admins.

## Changelog

```markdown
### Added
- Added `sonata.media.admin.media.show_mosaic_button` and `sonata.media.admin.gallery.show_mosaic_button` parameters
```

### DIFF
--- a/src/Resources/config/doctrine_mongodb_admin.xml
+++ b/src/Resources/config/doctrine_mongodb_admin.xml
@@ -8,10 +8,12 @@
         <parameter key="sonata.media.admin.media.class">Sonata\MediaBundle\Admin\ODM\MediaAdmin</parameter>
         <parameter key="sonata.media.admin.media.controller">SonataMediaBundle:MediaAdmin</parameter>
         <parameter key="sonata.media.admin.media.translation_domain">SonataMediaBundle</parameter>
+        <parameter key="sonata.media.admin.media.show_mosaic_button">true</parameter>
         <!-- GALLERY -->
         <parameter key="sonata.media.admin.gallery.class">Sonata\MediaBundle\Admin\GalleryAdmin</parameter>
         <parameter key="sonata.media.admin.gallery.controller">SonataMediaBundle:GalleryAdmin</parameter>
         <parameter key="sonata.media.admin.gallery.translation_domain">%sonata.media.admin.media.translation_domain%</parameter>
+        <parameter key="sonata.media.admin.gallery.show_mosaic_button">true</parameter>
         <!-- GALLERY HAS MEDIA -->
         <parameter key="sonata.media.admin.gallery_has_media.class">Sonata\MediaBundle\Admin\GalleryHasMediaAdmin</parameter>
         <parameter key="sonata.media.admin.gallery_has_media.controller">SonataAdminBundle:CRUD</parameter>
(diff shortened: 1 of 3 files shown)


kunicmarko20:
https://symfony.com/doc/master/bundles/SonataAdminBundle/reference/action_list.html#mosaic-view-button

It already exists 😅

tifabien:
The problem is the fact that the config is global.
What I want is to deactivate globally the mosaic button for all my Admin classes but want to have this mosaic view in the media list.

kunicmarko20:
But then, global won't work on media anymore, right?

src/Resources/config/doctrine_mongodb_admin.xml
```diff
@@ -8,18 +8,20 @@
         <parameter key="sonata.media.admin.media.class">Sonata\MediaBundle\Admin\ODM\MediaAdmin</parameter>
         <parameter key="sonata.media.admin.media.controller">SonataMediaBundle:MediaAdmin</parameter>
         <parameter key="sonata.media.admin.media.translation_domain">SonataMediaBundle</parameter>
+        <parameter key="sonata.media.admin.media.show_mosaic_button">true</parameter>
```
core23: Please don't add new parameters. We should get rid of this

tifabien: How do you want we manage this option?

core23: Don't use a variable, use hardcoded `true` or `false` values

tifabien: Humm I'm not sure to understand. If I put an hardcoded value in the show_mosaic_button option, how the user will choose if he/she wants to show or hide the button?

OskarStark:
Any news here?

tifabien:
I can work again on this but I asked for help in my last comment. Any suggestion are welcome :)

OskarStark:
Sorry...

@core23 can you be more precise on your comment?

tifabien:
I think this is still relevant. I just need some guidance about the implementation. 
Regarding 
>  But then, global won't work on media anymore, right?

We could use the `%sonata.admin.configuration.show.mosaic.button%` param instead of `true/false` this way we could stick to the global config but able to override admin by admin.

jordisala1991:
Hi @tifabien , are you still interested in this PR? 

Maybe a solution is to add a new key on configuration, then on the extension convert to parameter, and then injecting it on the services file. The default value should be false to keep bc, but it will allow you to change it to true on your projects

jordisala1991:
Other way is to just override the DI config for the media admin and add this on your side. 

I dont really know the benefits on doing this change here

jordisala1991:
@sonata-project/contributors wdyt?

phansys:
I think we could leverage the current configuration adding the possibility to accept an array with the admin ids which should show the mosaic layout. And additionally, we should provide a method to control this individually at each media admin, in order to let the users to configure this setting in their service definitions.

I mean something like:
```yaml
app.media.admin:
    # ...
    tags:
        - { name: sonata.admin, list_mode: mosaic }
```

PR #1459, opened 2018-08-06
Remove useless duplicated code

tifabien:
I am targeting this branch, because it removes useless duplicated code. The code is already the same in the `FileProvider::getReferenceImage` parent method

### DIFF
--- a/src/Provider/ImageProvider.php
+++ b/src/Provider/ImageProvider.php
@@ -130,17 +130,6 @@ class ImageProvider extends FileProvider
     /**
      * {@inheritdoc}
      */
-    public function getReferenceImage(MediaInterface $media)
-    {
-        return sprintf('%s/%s',
-            $this->generatePath($media),
-            $media->getProviderReference()
-        );
-    }
-
-    /**
-     * {@inheritdoc}
-     */
     public function updateMetadata(MediaInterface $media, $force = true)
     {
         try {


src/Provider/ImageProvider.php
```diff
@@ -127,17 +127,6 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
         return array_merge($params, $options);
     }
 
-    /**
-     * {@inheritdoc}
-     */
-    public function getReferenceImage(MediaInterface $media)
```
kunicmarko20: Method is public and removing it would be a BC Break, you can deprecate it if you want.

greg0ire: Why would it be a BC break if the parent method exists?

kunicmarko20: My bad, didn’t understand

tifabien:
I can see that indeed the lint build failed but I don't know why... Is there a way to rebuild the task ?

kunicmarko20:
restarted

kunicmarko20:
@greg0ire do we need a changelog here?

OskarStark:
> @greg0ire do we need a changelog here?

lets keep the changelog and tag it as minor

greg0ire:
I would remove the changelog and keep it pedantic: imagine you update the lib, you will not notice any changes, will you? I mean the public API is the same, and the behavior is the same.

tifabien:
I agree to not include the changelog as the public API is the same. That's invisible to the user.

greg0ire:
It's settled then, I removed the changelog

greg0ire:
Thanks @tifabien !

PR #1465, opened 2018-08-29
align MediaAdminController test code with parameter requirement

esserj:
I am targeting this branch, because the test is broken for code in the branch.

## Subject

The admin bundle CRUDController had its setFormTheme signature updated in https://github.com/sonata-project/SonataAdminBundle/commit/06e1e7ffd837fcf117bd0ee317f885fe93a3168c
due to it nwo enforcing an array the test has to be adjusted.
This doesn't seem like a BC break as the docblock did state array before

### DIFF
--- a/tests/Controller/MediaAdminControllerTest.php
+++ b/tests/Controller/MediaAdminControllerTest.php
@@ -94,7 +94,6 @@ class MediaAdminControllerTest extends TestCase
         $this->admin->checkAccess('create')->shouldBeCalled();
         $this->request->get('provider')->willReturn(true);
         $this->request->isMethod('get')->willReturn(true);
-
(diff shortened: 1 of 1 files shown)


kunicmarko20:
Lint errors should be fixed in a different PR, would you like to provide a PR? (I can do it if you don't have time)

esserj:
I'd like to try as there is another PR I ned to make for another bug, jsut came accros this one while running the tests before starting my changes, thats the php-cs-fixer i must use right?

kunicmarko20:
Yes, just run the fixer and everything should be fixed.

kunicmarko20:
@OskarStark let's merge this before the lint because this fixes the build and then we merge #1466

PR #1468, opened 2018-09-03
Fix broken build

greg0ire:

Fixes the build.

I've always wanted to do an octopus merge, thanks for providing me with the opportunity @esserj :heart: 

I think I don't need to reference your PRs for them to be closed when this is merged, because the hash of your commits is the same.


greg0ire:
:thinking: I could not merge because of the settings, only rebase, which made the hashes change, which means the PRs have to be closed manually. Thanks a lot @esserj !

PR #1480, opened 2018-10-11
Fix dailymotion provider regex

Phenix789:
I am targeting this branch, because I fix the regex use by the `DailyMotionProvider` to extract the video reference from the url.

No issue found for this

## Changelog

```markdown
### Fixed
 - Fix dailymotion regex to extract the video reference from the url.
```

## Subject

Update dailymotion regex to match all :
- `www.dailymotion.com/video/xxxxx`
- `http://www.dailymotion.com/video/xxxxx`
- `https://www.dailymotion.com/video/xxxxx`

So now the provider accept the video reference (like previously), but also url (like youtube provider can do) as `binaryContent`


OskarStark:
Nice PR, thank you for your contribution!

Can you please add a testcase for the urls and check travis?

Phenix789:
I update `DailyMotionProviderTest::testTransformWithUrl` to test all accepted `binaryContent` format.
I also update `MediaAdminControllerTest` because `CRUDController` call the method `all()` on form.

PR #1506, opened 2018-12-31
Undeprecated MediaManagerInterface

core23:
## Subject
The deprecation notice is wrong. You couldn't use `ManagerInterface` in any case.

## Changelog

```markdown
### Deprecated
- Undeprecated `MediaManagerInterface`
```


greg0ire:
> You couldn't use ManagerInterface in any case.

Can you please elaborate? Do you mean there are cases where consumer need to be sure both interfaces are implemented so that they can call methods on both interfaces or something?

core23:
The class extends `ManagerInterface` AND `PageableManagerInterface`, so you couldn't use the `ManagerInterface` in any case.

PR #1542, opened 2019-01-28
Fix default values for height/width in format configuration

Sharsie:
## Default height values are expected to be null, not false

As per the checks here: https://github.com/sonata-project/SonataMediaBundle/blob/061c00def148bdb695bdb9c595ec209c94dad80f/src/Resizer/SimpleResizer.php#L78-L88
and here: https://github.com/sonata-project/SonataMediaBundle/blob/061c00def148bdb695bdb9c595ec209c94dad80f/src/Resizer/SquareResizer.php#L88
The configuration of height (when not provided) should be null.

However the DependencyInjection builds configuration in such a way that providing no height (or width in case of custom formats) in formats configuration produces `false` instead as seen here: https://github.com/sonata-project/SonataMediaBundle/blob/08177acc4553678ba0ed03b2632981cc38e0b167/src/DependencyInjection/Compiler/AddProviderCompilerPass.php#L120
and here: https://github.com/sonata-project/SonataMediaBundle/blob/08177acc4553678ba0ed03b2632981cc38e0b167/src/DependencyInjection/Configuration.php#L69

and here: https://github.com/sonata-project/SonataMediaBundle/blob/08177acc4553678ba0ed03b2632981cc38e0b167/src/DependencyInjection/Configuration.php#L118-L119

The PR only changes the default values.

I am targeting this branch, because this is a hotfix.

Commit which broke the behaviour contains new CS rules with strict checks: https://github.com/sonata-project/SonataMediaBundle/commit/061c00def148bdb695bdb9c595ec209c94dad80f
There could possibly be other related issues, but I haven't come accross anything else

## Changelog

```markdown
### Fixed
- Strict checking of height/width sizes in formats configuration
```


Sharsie:
@kunicmarko20 Most of these seem to be strict comparisons (I think between master and 3.x branch?)
I think those were introduced by the commit I mentioned https://github.com/sonata-project/SonataMediaBundle/commit/061c00def148bdb695bdb9c595ec209c94dad80f

greg0ire:
Try rebasing to get the cs fixes then.
Also, please associate your email address with your Github account, or change the email in your commits to an address already associated with it.

Sharsie:
@greg0ire sorry, fixed the author and rebased 3.x
It's a mystery I haven't had those cs fixes already when I cloned the repository yesterday :\

Sharsie:
@greg0ire Fixed the commit message, thanks for the guide, got a few other hints from it as well :)
Dismissed the review again by the push, but hey...

greg0ire:
Great job!

PR #1544, opened 2019-01-29
Add support for <picture> element in media helper

### LINKED ISSUE #1543: Add <picture> element support
## Add <picture> element support

While the `` approach will do the trick many times, the `<picture>` element offers more control when images with different aspect ratios and/or content are required depending on the breakpoint.
It would be good to have builtin support for this in the media helper to allow simple use of image formats with sensible defaults for media queries, just like in the `` case.

cussack:
## Add support for <picture> element in media helper

This PR adds an additional configuration option `picture` to the media helper, that can act as an alternative to `srcset` and `sizes` and will trigger rendering of a `<picture>` element instead of an ``.

I am targeting this branch, because the PR offers an additional configuration option to the media helper. It is backwards compatible because media helper output remains unchanged if the new option is not used.

Closes #1543

## Changelog

```markdown
### Added
- Added configuration option for the media helper to generate a <picture> instead of <img>

### Fixed
- SimpleResizer::computeBox() will return the size of the resulting image instead of just the scaled image before the cropping step.
```

## To do
    - [X] Add tests

### DIFF
--- a/docs/reference/helpers.rst
+++ b/docs/reference/helpers.rst
@@ -96,10 +96,9 @@ If you need to specify media queries explicitly, do so with an object as follows
 
 .. code-block:: jinja
 
-    {% media media, 'large' with {'srcset': {'(max-width: 500px)': 'small', '(max-width: 1200px)': 'big', 'fallback': 'huge'}} %}
+    {% media media, 'large' with {'srcset': {'(max-width: 500px)': 'small', '(max-width: 1200px)': 'big'}} %}
 
-The ``fallback`` option determines which size is going to be rendered as ``<img>`` inside the ``<picture>`` element.
-If omitted, it defaults to ``'reference'``.
+The format parameter (``'large'`` above) determines which size is going to be rendered as ``<img>`` inside the ``<picture>`` element.
 
 Thumbnails for files
 --------------------
--- a/docs/reference/helpers.rst
+++ b/docs/reference/helpers.rst
@@ -85,6 +85,21 @@ helper. The option expects either a string or an array of formats.
 
     {% media media, 'large' with {'srcset': ['small', 'big']} %}
 
+To render the image as ``<picture>`` element instead of ``<img>``, pass a ``picture`` key instead of ``srcset`` above:
+
+.. code-block:: jinja
+
+    {% media media, 'large' with {'picture': ['small', 'big']} %}
+
+Media queries for ``<source>`` tags will default to a ``max-width`` equal to the image size.
+If you need to specify media queries explicitly, do so with an object as follows:
+
+.. code-block:: jinja
+
+    {% media media, 'large' with {'srcset': {'(max-width: 500px)': 'small', '(max-width: 1200px)': 'big', 'fallback': 'huge'}} %}
+
+The ``fallback`` option determines which size is going to be rendered as ``<img>`` inside the ``<picture>`` element.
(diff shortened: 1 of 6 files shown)


docs/reference/helpers.rst
```diff
@@ -85,6 +85,20 @@ helper. The option expects either a string or an array of formats.
 
     {% media media, 'large' with {'srcset': ['small', 'big']} %}
 
+To render the image as ``<picture>`` element instead of ``<img>``, pass a ``picture`` key instead of ``srcset`` above:
```
greg0ire: What should happen if both keys are provided? I think it should throw an exception.

cussack: Do we already have one that‘s suitable?

greg0ire: We can use a native one for this, since that exception is not meant to be caught. `\LogicException` would do IMO.

cussack:
@greg0ire @OskarStark thanks for picking up so quickly. If the build is still successful now, I would consider this PR ready for your review.

tests/Provider/ImageProviderTest.php
```diff
@@ -122,6 +122,9 @@ public function testHelperProperties()
                 $adminBox, // Third call
                 $largeBox, // Fourth call
                 $mediumBox,
+                $largeBox,
+                $largeBox, // Fifth call
```
OskarStark: LargeBox twice is correct?

src/Provider/ImageProvider.php
```diff
@@ -89,7 +93,23 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
...
-        if (MediaProviderInterface::FORMAT_ADMIN !== $format) {
+        if (isset($options['picture'])) {
+            $pictureParams = [];
+            foreach ($options['picture'] as $key => $pictureFormat) {
+                $formatName = $this->getFormatName($media, $pictureFormat);
+                $settings = $this->getFormat($formatName);
+                $src = $this->generatePublicUrl($media, $formatName);
+                $mediaQuery = \is_string($key)
+                    ? $key
+                    : '(max-width: '.$this->resizer->getBox($media, $settings)->getWidth().'px)';
```
OskarStark: Please use sprintf here

tests/Provider/ImageProviderTest.php
cussack: yes, processing order is format parameter followed by a loop over the picture array. For the test cases I added the sequence is large-medium-large, large-medium-large

src/Provider/ImageProvider.php
cussack: I changed it, thanks for the hint

greg0ire:
Thanks @cussack !

kunicmarko20:
Please squash it next time. 

Thanks @cussack!

cussack:
Thanks for your great work, I have to say that CMF/Sonata is the first CMS I'm truly happy with from a developer's standpoint. Keep it up!

greg0ire:
@kunicmarko20 oh damn didn't see there were 8 commits… sorry!

cussack:
who does the squash usually btw, contributor or maintainer?

kunicmarko20:
Both can do it, but if it is not set as a default option on github merge, maintainers usually forget.

OskarStark:
Thank you for your kind words @cussack 🙏🤗

PR #1546, opened 2019-01-31
Type hint all the thingz

greg0ire:
## Subject
I am targeting this branch, because this BC (right?).

### DIFF
--- a/src/Twig/Extension/MediaExtension.php
+++ b/src/Twig/Extension/MediaExtension.php
@@ -83,7 +83,7 @@ class MediaExtension extends \Twig_Extension implements \Twig_Extension_InitRunt
     {
         $media = $this->getMedia($media);
 
-        if (!$media) {
+        if (null === $media) {
             return '';
         }
 
@@ -115,7 +115,7 @@ class MediaExtension extends \Twig_Extension implements \Twig_Extension_InitRunt
     {
         $media = $this->getMedia($media);
 
-        if (!$media) {
+        if (null === $media) {
             return '';
         }
 
@@ -195,10 +195,8 @@ class MediaExtension extends \Twig_Extension implements \Twig_Extension_InitRunt
 
     /**
      * @param mixed $media
-     *
-     * @return MediaInterface|bool|null
      */
-    private function getMedia($media)
+    private function getMedia($media): ?MediaInterface
     {
         if (!$media instanceof MediaInterface && \strlen((string) $media) > 0) {
             $media = $this->mediaManager->findOneBy([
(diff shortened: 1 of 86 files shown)


src/Twig/Extension/MediaExtension.php
```diff
@@ -207,11 +205,11 @@ private function getMedia($media)
         }
 
         if (!$media instanceof MediaInterface) {
-            return false;
+            return null;
```
OskarStark: Hmm this could be a be a be break. Did you double check the usage in this class?

tests/Block/Breadcrumb/BreadcrumbTest.php
```diff
@@ -23,7 +23,7 @@
  */
 class BreadcrumbTest extends AbstractBlockServiceTestCase
 {
-    public function testBlockService()
+    public function testBlockService(): void
```
OskarStark: What if you extend this test class and don’t have the void return type when overriding this method? Is it ok?

tests/Controller/MediaControllerTest.php
```diff
@@ -138,29 +139,36 @@ public function testViewActionRendersView()
...
         $mediaManager = $this->prophesize(BaseEntityManager::class);
 
         $this->container->get('sonata.media.manager.media')->willReturn($mediaManager->reveal());
         $mediaManager->find($id)->willReturn($media);
     }
 
-    private function configureGetProvider($pool, $media, $provider)
-    {
+    private function configureGetProvider(
+        ObjectProphecy $pool,
```
OskarStark: You wouldnt add a rypehibt for Phpunit mock object here, why not type hinting the real class?

src/Twig/Extension/MediaExtension.php
greg0ire: yes

tests/Block/Breadcrumb/BreadcrumbTest.php
greg0ire: It's a test

tests/Controller/MediaControllerTest.php
greg0ire: Because it's not an instance of said class until you call `reveal()`

OskarStark: I know maybe it confuses more then it helps but from a technical POV it’s fine

greg0ire: Yeah I thought it was not great, but then thought other people would want to add a type hint and lose time finding out why it wasn't added. Plus reviewers would have asked why I did not put one :P Plus you get auto-complete in the method, so…

PR #1562, opened 2019-02-17
Fix incorrect quote

Seb33300:
## Subject

Fix incorrect quote in installation documentation config.


OskarStark:
Good catch 👍🏻👀

PR #1567, opened 2019-02-22
CategoryManager - wrong type hint

jerome-fix:
## CategoryManager - wrong FQDN for CategoryInterface

I am targeting this branch, because branch 3.x is currently broken.

## Changelog

```markdown
### Fixed
- return type for `CategoryManager::find()`
```

### DIFF
--- a/src/Model/CategoryManager.php
+++ b/src/Model/CategoryManager.php
@@ -13,8 +13,8 @@ declare(strict_types=1);
 
 namespace Sonata\MediaBundle\Model;
 
+use Sonata\ClassificationBundle\Model\CategoryInterface;
 use Sonata\ClassificationBundle\Model\CategoryManagerInterface as ManagerInterface;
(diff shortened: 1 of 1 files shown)


src/Model/CategoryManager.php
```diff
@@ -54,7 +54,7 @@ public function getRootCategories($loadChildren = true): iterable
     /**
      * {@inheritdoc}
      */
-    public function find($categoryId): iterable
+    public function find($categoryId):  ?CategoryInterface
```
greg0ire: Please remove the extra space and the build will be green.
```suggestion
    public function find($categoryId): ?CategoryInterface
```

jerome-fix: Sorry. I just push a new version

kunicmarko20:
thanks @jerome-fix !

nico-dangerous:
Can you tell when this fix will be released ? Thanks

kunicmarko20:
We requested a release, all I can say is soon.

jerome-fix:
Thanks !

PR #1583, opened 2019-04-05
fix(provider): fix loosing provider on multi providers context & max …

bmaziere:
…post size exceeded

## Subject

When a post is done but that it exceeds php max post size, post values seems to be empty. Then we loose the provider as it is a post parameter. With a multi providers context, the provider is lost so it has to be persistent. 

I am targeting this branch, because it's not a BC break i presume.

## Changelog

```markdown
### Fixed
Fix error 500 when max post size is exceeded on multi providers context
```

### DIFF
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -91,6 +91,12 @@ abstract class BaseMediaAdmin extends AbstractAdmin
             $this->getRequest()->query->set('provider', $provider);
         }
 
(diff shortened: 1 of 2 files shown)


mremi:
Hi, any news?

bmaziere:
FlintCI seems blocked for a while. anything i can do @OskarStark  ?

OskarStark:
Lets merge it, thank you @bmaziere 👍

PR #1585, opened 2019-04-23
Add option to lazy load images

### LINKED ISSUE #1584: Add option to lazy load images
## Feature Request

In many contexts it is useful to be able to lazy load images, but this bundle lacks this feature currently.
It would be good to have a JS-agnostic option to replace `src` and `srcset` attributes with an alternative that allows lazy loading.

cussack:
## Subject

Adds the `lazy` option to the media helper to allow moving `src` and `srcset` attributes to `data-src` and `data-srcset`, respectively.

I am targeting this branch, because the introduced option has no effect if omitted and thus is BC.

Closes #1584 

## Changelog

```markdown
### Added
- `lazy` option to media helper for images

```

### DIFF
--- a/docs/reference/helpers.rst
+++ b/docs/reference/helpers.rst
@@ -98,6 +98,16 @@ If you need to specify media queries explicitly, do so with an object as follows
 
 The format parameter (``'large'`` above) determines which size is going to be rendered as ``<img>`` inside the ``<picture>`` element.
 
+To make the ``<img>`` or ``<picture>`` lazy-loadable, add the parameter ``lazy: true``. ``src`` and ``srcset`` attributes
+will then become ``data-src`` and ``data-srcset``, respectively. You can define a loader or placeholder image through
+the ``src`` option if desired, otherwise ``src`` will default to ``#``. Please note that this will only prevent the image
+from loading, you will need to add code or a library that performs actual lazy-loading yourself.
+
+.. code-block:: jinja
+
+    {% media media, 'large' with {'lazy': true} %}
+
+
 Thumbnails for files
 --------------------
 
--- a/src/Provider/ImageProvider.php
+++ b/src/Provider/ImageProvider.php
@@ -71,6 +71,8 @@ class ImageProvider extends FileProvider
             throw new \LogicException("The 'srcset' and 'picture' options must not be used simultaneously.");
         }
 
+        $attrPrefix = isset($options['lazy']) && true === $options['lazy'] ? 'data-' : '';
+
         if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
             $box = $media->getBox();
         } else {
@@ -88,7 +90,8 @@ class ImageProvider extends FileProvider
         $params = [
             'alt' => $media->getName(),
             'title' => $media->getName(),
-            'src' => $this->generatePublicUrl($media, $format),
+            'src' => $options['src'] ?? '#',
+            $attrPrefix.'src' => $this->generatePublicUrl($media, $format),
             'width' => $mediaWidth,
             'height' => $box->getHeight(),
         ];
@@ -103,7 +106,7 @@ class ImageProvider extends FileProvider
                     ? $key
                     : sprintf('(max-width: %dpx)', $this->resizer->getBox($media, $settings)->getWidth());
 
-                $pictureParams['source'][] = ['media' => $mediaQuery, 'srcset' => $src];
+                $pictureParams['source'][] = ['media' => $mediaQuery, $attrPrefix.'srcset' => $src];
             }
 
             unset($options['picture']);
@@ -145,7 +148,10 @@ class ImageProvider extends FileProvider
                     $media->getBox()->getWidth()
                 );
 
-                $params['srcset'] = implode(', ', $srcSet);
+                $params[$attrPrefix.'srcset'] = implode(', ', $srcSet);
+            } elseif ($attrPrefix) {
+                $options[$attrPrefix.'srcset'] = $options['srcset'];
+                unset($options['srcset']);
             }
 
             $params['sizes'] = sprintf('(max-width: %1$dpx) 100vw, %1$dpx', $mediaWidth);
--- a/src/Provider/ImageProvider.php
+++ b/src/Provider/ImageProvider.php
@@ -149,7 +149,7 @@ class ImageProvider extends FileProvider
                 );
 
                 $params[$attrPrefix.'srcset'] = implode(', ', $srcSet);
-            } elseif ($attrPrefix) {
+            } elseif ($attrPrefix !== '') {
                 $options[$attrPrefix.'srcset'] = $options['srcset'];
                 unset($options['srcset']);
             }
--- a/tests/Provider/ImageProviderTest.php
+++ b/tests/Provider/ImageProviderTest.php
@@ -124,6 +124,12 @@ class ImageProviderTest extends AbstractProviderTest
                 $mediumBox,
                 $largeBox,
                 $largeBox, // Fifth call
+                $largeBox,
+                $largeBox, // Sixth call
+                $mediumBox,
+                $largeBox,
+                $largeBox, // Seventh call
+                $largeBox, // Eighth call
                 $mediumBox,
                 $largeBox
             ));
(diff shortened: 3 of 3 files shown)


OskarStark:
Please check travis, thank you!

kunicmarko20:
If I am correct, this will just prevent loading image? (because there is no src) The "lazy" part is missing, that should be added by the end user?

cussack:
@kunicmarko20 „lazy“ is implemented, „loading“ not because there‘s too many options out there

kunicmarko20:
I get it, there are too many options out there but that should then be explicitly mentioned in the docs.

src/Provider/ImageProvider.php
```diff
@@ -88,7 +90,7 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
         $params = [
             'alt' => $media->getName(),
             'title' => $media->getName(),
-            'src' => $this->generatePublicUrl($media, $format),
+            $attrPrefix . 'src' => $this->generatePublicUrl($media, $format),
```
core23: This will render all images without an `src` tag, which is mandatory!
https://www.w3.org/MarkUp/html3/img.html

cussack:
@OskarStark @kunicmarko20 @core23 I addressed your issues, please check.

src/Provider/ImageProvider.php
```diff
@@ -88,7 +90,8 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
         $params = [
             'alt' => $media->getName(),
             'title' => $media->getName(),
-            'src' => $this->generatePublicUrl($media, $format),
+            'src' => '#',
```
core23: Not sure if this is even better... Is this a valid `src` url?

cussack: Looks like it is. Paste this:

```
<!DOCTYPE html>
<html lang="en">
<head><title>test</title></head>
<body>
<img src="#" alt="">
</body>
</html>
```

here: https://validator.w3.org/#validate_by_input

cussack: @core23 what do you think?

cussack:
@core23 can you please provide some feedback? Thanks!

core23:
Sorry, but I'm still not a big fan of this PR. You add a `data-src` attribute instead of a `src` attribute. But his will not work as a valid lazy loading. There is no JavaScript code or anything else that will trigger the lazy loading. This needs to be done in the users application.

Either provide a working solution or add a hook, so you can customize this for your application.

cussack:
There’s three main reasons why I chose this approach.

First, there’s several possible ways to implement lazy loading in JS as well as several different use cases, so we shouldn’t make that choice here, this is a library intended to be reusable in different scenarios.
Second, all lazy loading implementations I used so far are looking for `data-` attributes, so I think it makes sense to make this the default behavior.
Third, specifically the media helper is a facade focused on ease of use, thus my choice for API and (default) behavior.

While I agree that a hook would be the most flexible solution (probably as a pluggable attribute mapper), I think it would be overkill at this point and should be implemented only with the next attribute mapping requirement, should it ever come across.
Maybe a good middle ground can be to add optional attribute mapping configuration and some hints for JS implementation in the docs.

I would like to hear some more opinions @OskarStark @kunicmarko20

docs/reference/helpers.rst
```diff
@@ -98,6 +98,16 @@ If you need to specify media queries explicitly, do so with an object as follows
 
 The format parameter (``'large'`` above) determines which size is going to be rendered as ``<img>`` inside the ``<picture>`` element.
 
+To make the ``<img>`` or ``<picture>`` lazy-loadable, add the parameter ``lazy: true``. ``src`` and ``srcset`` attributes
+will then become ``data-src`` and ``data-srcset``, respectively. You can define a loader or placeholder image through
+the ``src`` option if desired, otherwise ``src`` will default to ``#``. Please note that this will only prevent the image
+from loading, you will need to add code or a library that performs actual lazy-loading yourself.
```
greg0ire: Please comply with the sf docs standards:

> [Each line should break approximately after the first word that crosses the 72nd character (so most lines end up being 72-78 characters);](http://symfony.com/doc/current/contributing/documentation/standards.html)

src/Provider/ImageProvider.php
```diff
@@ -71,6 +71,8 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
             throw new \LogicException("The 'srcset' and 'picture' options must not be used simultaneously.");
         }
 
+        $attrPrefix = isset($options['lazy']) && true === $options['lazy'] ? 'data-' : '';
```
greg0ire: Shouldn't there be some kind of validation somewhere? IMO there should be an exception if `$options['lazy']` is not a boolean.

src/Provider/ImageProvider.php
```diff
@@ -145,7 +148,10 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
                     $media->getBox()->getWidth()
                 );
 
-                $params['srcset'] = implode(', ', $srcSet);
+                $params[$attrPrefix.'srcset'] = implode(', ', $srcSet);
+            } elseif ($attrPrefix) {
```
greg0ire: Let's be explicit about the case we are referring to
```suggestion
            } elseif ($attrPrefix !== '') {
```

tests/Provider/ImageProviderTest.php
```diff
@@ -189,6 +195,32 @@ public function testHelperProperties(): void
...
+        $this->assertSame('#', $properties['src']);
+        $this->assertArrayNotHasKey('srcset', $properties);
+
+        $properties = $provider->getHelperProperties($media, 'default_large', ['src' => 'scalar src', 'srcset' => 'scalar srcset', 'lazy' => true]);
+        $this->assertArrayHasKey('data-src', $properties);
+        $this->assertSame('scalar src', $properties['src']);
+        $this->assertSame('scalar srcset', $properties['data-srcset']);
+        $this->assertArrayNotHasKey('srcset', $properties);
+
+        $properties = $provider->getHelperProperties($media, 'default_large', ['picture' => ['default_medium', 'default_large'], 'class' => 'some-class', 'lazy' => true]);
```
greg0ire:
```suggestion
        $properties = $provider->getHelperProperties($media, 'default_large', [
            'picture' => ['default_medium', 'default_large'],
            'class' => 'some-class', 'lazy' => true
        ]);
```

src/Provider/ImageProvider.php
cussack: I was thinking of allowing an attribute map as well, something like 

```
{
  lazy: {
    src: data-src,
    srcset: data-srcset
  }
}
```

Is that something we should add in here?

cussack:
@greg0ire I will address your comments shortly.

What are your thoughts regarding @core23‘s major concerns? Should we add a JS implementation?
Adding a hook properly would involve a major refactoring IMO, `getHelperProprties()` would need to be changed into a configurable sequence of attribute mappers/transformers.

src/Provider/ImageProvider.php
greg0ire: > Perhaps in a different key though because it’s more generic

But then you would have different keys, which is not ideal I think. Do you think people really will want to customize this (I have no idea)?

cussack: Probably not, as I wrote earlier all implementations I came across use `data-` attributes. It would only allow to use implementations that don’t follow this or just rename attributes for totally different use cases

greg0ire:
@core23 is right when he says it's bad to "promise" lazy loading and then provide only half the implementation. Maybe allow customizing the attributes names without having `lazy` written anywhere in the code, and the provide a cookbook entry showing how to leverage this to implement lazy loading?

src/Provider/ImageProvider.php
greg0ire: Let's go for "totally different use cases", and change `lazy` into something else.

cussack:
I could totally live with that, but how would we allow the custom value for `src` for example? One option would be

```
{
  name: default-value
  attrMap: {
    name: new-name
  }
}
-> <img name="default-value" new-name="original-value">
```

But that would break if you want to customize srcset, for example. Another option:

```
{
  attrMap: {
    name1: {
      rename: new-name1,
      fallback: fallback-value
    },
    name2: new-name2
}
-> <img name1="fallback-value" new-name1="original-value1" new-name2="original-value2">
```

which is pretty verbose, not to say pretty ugly. Thoughts?

greg0ire:
Ugh indeed this looks ugly… this looks like a dead-end, I don't know what to propose… did you currently find a way to achieve your goal without any change to this lib?
Also, I can see that [it's possible to run javascript to prevent images from loading thanks to `<noscript>`](https://stackoverflow.com/a/20863227/353612), maybe do that instead, then parse the DOM in JS to find these `noscript` tags, and rewrite their content to another node, still in JS. That way, you don't have to change this lib, you just have to wrap your images into `noscript` tags.

cussack:
@greg0ire from a technical point of view I don't care too much if this PR will be merged or not, I will just override the ImageProvider and configure mine instead, it's also possible to just override the helper template and to do it there.

I just thought it would be good to have some lightweight out of the box support for this, because image lazy loading is a more or less important feature for all websites using images.
I will put this topic aside for the moment, perhaps we will continue discussion in #1584 at some point if there is some increased interest in this issue.

PR #1592, opened 2019-05-09
Fixed deletion of images with id greater than 1000

MaksFM:
Inside the $this->getReferenceImage() method, the id of the $media object is used, but at the time of the postRemove event, the $media object is detached from an EntityManager and its id is null. During the postRemove event, the $media->getId() method returns null, and the $this->getReferenceImage() method returns the wrong path and the file remains in the file system as garbage. My changes fix this.

## Subject

I am targeting this branch, because I use the last stable release in my project, and found an error in it, but my changes are backward compatible and fix it..

## Changelog

```markdown
### Fixed
Original image files with id greater than 1000 are not deleted from the file system.
```


MaksFM:
Demonstrates incorrect work of the getReferenceImage() method during the postRemove event. Details in the comments of the attached file. 

[tests.zip](https://github.com/sonata-project/SonataMediaBundle/files/3165988/tests.zip)

This is my first contribution, I don’t know how to provide the necessary tests more correctly.

MaksFM:
Good afternoon, I added tests for the code that fixes the described error.  Please give your comments if this solution is not suitable for the general repository.  Just then I will use in my project a private solution of the problem, not affecting the internal sources of the corresponding component.  Although, I still consider the most preferred bug fix in the main repository.

src/Provider/BaseProvider.php
```diff
@@ -63,6 +63,11 @@ abstract class BaseProvider implements MediaProviderInterface
      */
     protected $name;
 
+    /**
+     * @var array
+     */
+    protected $clones = [];
```
core23: Please make this private by default

MaksFM: Done.

src/Provider/BaseProvider.php
```diff
@@ -63,6 +63,11 @@ abstract class BaseProvider implements MediaProviderInterface
      */
     protected $name;
 
+    /**
+     * @var array
```
phansys:
```suggestion
     * @var MediaInterface[]
```

src/Provider/BaseProvider.php
```diff
@@ -191,6 +199,13 @@ public function preRemove(MediaInterface $media)
      */
     public function postRemove(MediaInterface $media)
     {
+        $hash = spl_object_hash($media);
+
+        if (array_key_exists($hash, $this->clones)) {
```
phansys:
```suggestion
        if (isset($this->clones[$hash])) {
```

tests/Provider/BaseProviderTest.php
```diff
@@ -151,9 +159,9 @@ public function getAbsolutePath(MediaInterface $media): void
     /**
      * {@inheritdoc}
      */
-    public function getReferenceImage(MediaInterface $media): void
+    public function getReferenceImage(MediaInterface $media)
```
phansys:
```suggestion
    public function getReferenceImage(MediaInterface $media): string
```

tests/Provider/BaseProviderThisVerOrBelowTest.php
```diff
@@ -0,0 +1,212 @@
...
+
+        return $provider;
+    }
+
+    public function testPostRemove(): void
+    {
+        $provider = $this->getProvider();
+        $media = new Media();
+        
+        //The ID is more than a thousand, the bug manifests itself under this condition.
```
core23: You can remove this comment, as the context of *the bug* is missing after merging this PR.

MaksFM: This test is more likely to demonstrate the error before correction and its absence after. In the resulting branch, it is preferable to delete it completely.

tests/Provider/BaseProviderThisVerOrBelowTest.php
```diff
@@ -0,0 +1,210 @@
...
+    public function getReferenceImage(MediaInterface $media): string
+    {
+        // A copy of the code from \Sonata\MediaBundle\Provider\FileProvider::getReferenceImage()
+        $path = sprintf('%s/%s',
+            $this->generatePath($media),
+            $media->getProviderReference()
+        );
+        
+        // Test code.
+        AbstractProviderTest::assertSame('/0001/02/1f981a048e7d8b671415d17e9633abc0059df394.png', $path);
```
phansys: `AbstractProviderTest` => `self`.
BTW, fixture methods should avoid the testing responsibility. The expected path should not be hardcoded, it should be provided by context (argument or so).

tests/Provider/BaseProviderTest.php
```diff
@@ -96,10 +96,37 @@ public function testMetadata(): void
...
+        $this->assertArrayNotHasKey($hash, $prop->getValue($provider));
+        $this->assertSame('/0001/02/1f981a048e7d8b671415d17e9633abc0059df394.png', $provider->prevReferenceImage);
+
+        $prop->setAccessible(false);
+    }
 }
 
 class TestProvider extends BaseProvider
 {
+    public $prevReferenceImage;
```
phansys:
```suggestion
    /**
     * @var string
     */
    public $prevReferenceImage;
```

core23:
FlintCI is failing. Can you please apply the cs fixes @MaksFM

phansys:
Could you please squash your commits?

core23:
Thanks @MaksFM

PR #1594, opened 2019-05-16
fix(exception): fix value of code exception to respect strict type

xXNoceboXx:
## Subject

In function `getMetadata` in class `BaseVideoProvider` there were an error to throw an exception (line: 212) because this `RuntimeException` receive `null` to second parameter but this class is strict type and this exception normally receive a long, 

## Changelog
```markdown
### Fixed
- Value of code to throw RuntimeException when url of video throw an exception
```


greg0ire:
The cs fixes were applied on 3.x, see #1593 , please rebase and the PR should be easier to read :)

xXNoceboXx:
@greg0ire  it's rebased ;)

greg0ire:
Looks way better, thanks!

PR #1601, opened 2019-05-28
[Bug]: GalleryMediaCollectionInterface references non existant Interface

th-lange:
## Subject
- The GalleryMediaCollectionInterface refereced the old
GalleryHasMediaInterface wich was migrated to GalleryItemInterface.
- The old reference was repalced with the new interface

## Why Master branch:
because: Branch `3.x` uses the old  GalleryHasMediaInterface
branch master replaced the old interface with the new one. The Interface: GalleryMediaCollectionInterface was forgotten in this migration

Closes #{put_issue_number_here}

## Changelog
On branch master
Changes to be committed:
- modified:   src/Model/GalleryMediaCollectionInterface.php

```markdown
### Changed
src/Model/GalleryMediaCollectionInterface.php

### Fixed
src/Model/GalleryMediaCollectionInterface.php
```

### DIFF
--- a/src/Model/GalleryMediaCollectionInterface.php
+++ b/src/Model/GalleryMediaCollectionInterface.php
@@ -19,7 +19,7 @@ namespace Sonata\MediaBundle\Model;
  */
 interface GalleryMediaCollectionInterface
 {
-    public function addGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia);
+    public function addGalleryHasMedia(GalleryItemInterface $galleryItem);
 
-    public function removeGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia);
+    public function removeGalleryHasMedia(GalleryItemInterface $galleryItem);
 }


src/Model/GalleryMediaCollectionInterface.php
```diff
@@ -19,7 +19,7 @@
  */
 interface GalleryMediaCollectionInterface
 {
-    public function addGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia);
+    public function addGalleryHasMedia(GalleryItemInterface $galleryHasMedia);
```
greg0ire: Shouldn't the parameter name be updated as well?

th-lange: idk. I just know that like that it is an error.
As I am not too much into SMB I considered keeping the change as small as possible.
As this would have repercussions to other Interfaces, Classes and the like... And as such break functionallity...

greg0ire: Changing a parameter name has no consequence whatsoever: https://3v4l.org/HBWVA

th-lange: Done. Sorry. Thought you spoke about the function name 
... which does not seem proper either

PR #1603, opened 2019-06-05
Modernize references to controllers

phansys:
## Subject

Modernize references to controllers in order to avoid deprecated syntaxes.

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Changed
- Updated "_controller" attribute for routes which were using deprecated syntax.
```


OskarStark:
Yeah, we can use `patch` in this case

kunicmarko20:
Thank you @phansys !

phansys:
> @OskarStark should we use patch for this instead? It fixes deprecations. @phansys IMO that's what your changelog should be about: the consequences, not the implementation details.

IMHO, the only change that could be perceived by the end user is the fact that `$route->getDefault('_controller')` is different, and it could lead to some edge case where the user is trusting on that. By the other hand, the deprecation notice is not something in which the user should trust.

PR #1616, opened 2019-09-01
Respect inherited "field_description.options.route.name" value at `list_image.html.twig`

phansys:
## Subject

Respect inherited "field_description.options.route.name" value at `list_image.html.twig`.

## Changelog

```markdown
### Fixed
- Respect "field_description.options.route.name" value at `list_image.html.twig` instead of using hardcoded "edit".
```


greg0ire:
Thanks @phansys !

PR #1618, opened 2019-09-01
Use Twig namespaced classes where possible

phansys:
## Subject

Use Twig namespaced classes where possible.

The non namespaced classes were left intact at public method signatures where they can't be replaced.

## Changelog

```markdown
### Fixed
- Using deprecated `\Twig_` classes without namespace.
```


greg0ire:
Thanks @phansys !

PR #1639, opened 2019-11-04
Extract FileProviderInterface from FileProvider

### LINKED ISSUE #1638: Extract FileProviderInterface from FileProvider
## Feature Request

The `FileProvider` currently extends `BaseProvider`, but exposes additional methods in its public API beyond those from `MediaProviderInterface`, such as `getAllowedExtensions` and `getAllowedMimeTypes`, which are not part of any interface.

This hampers decoration as no interface provides these methods. Instead, one must either check for an instance of `FileProvider`, forcing you to extend rather than decorate, or use `method_exists`, which is suboptimal as well.

Would it be possible to extract a `FileProviderInterface` that extends `MediaProviderInterface` and contains these additional methods?

tominventisbe:
## Subject

The `FileProvider` currently extends `BaseProvider`, but exposes additional methods in its public API beyond those from `MediaProviderInterface`, such as `getAllowedExtensions` and `getAllowedMimeTypes`, which are not part of any interface.

This hampers decoration as no interface provides these methods. Instead, one must either check for an instance of `FileProvider`, forcing you to extend rather than decorate, or use `method_exists`, which is suboptimal as well.

This extracts a new extending interface. 

This change should be backwards compatible with 3.x as only a new interface is added, but the contract for it was already implemented. Existing code targeting the `FileProvider` or extending it will remain operational.

Closes #1638.

## Changelog
```markdown
### Changed
Extract and implement FileProviderInterface from FileProvider to allow decoration (fixes #1638).

```

### DIFF
--- a/src/Provider/FileProvider.php
+++ b/src/Provider/FileProvider.php
@@ -38,7 +38,7 @@ use Symfony\Component\Validator\Constraints\NotNull;
 /**
  * @final since sonata-project/media-bundle 3.21.0
  */
-class FileProvider extends BaseProvider
+class FileProvider extends BaseProvider implements FileProviderInterface
 {
     protected $allowedExtensions;
 
--- a/src/Provider/FileProviderInterface.php
+++ b/src/Provider/FileProviderInterface.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Provider;
+
+interface FileProviderInterface extends MediaProviderInterface
+{
+    /**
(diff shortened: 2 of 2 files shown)


tominventisbe:
I'm not sure most of the CI failures are related. I think I fixed the styling errors for the new file by adding the `declare(strict_types=1);` directive.

core23:
Please rebase, this should fix the problems

tominventisbe:
Done!

greg0ire:
Please kindly squash your commits together. If you don't, we'll try to remember to do it for you but it's best if you save us this trouble.

How to do that?

1. `git rebase -i origin/3.x`, assuming `origin` is a git remote that points to this repository, and not your fork. If you're not sure what your remotes are, run `git remote -vvv`, there should be your fork and the holy/reference/base/origin/whatever-you-call-it repository.
2. A window will show up with many lines, replace `pick` with `fixup` on every line but the first one
3. Close your editor, git should do its magic, and you should end up with one commit
4. Use `git push --force` to overwrite what you already push. Don't forget the `--force` option otherwise git will try to merge both things together.

tominventisbe:
Also done.

PR #1641, opened 2019-11-11
Dropping support for symfony < 3.4

franmomu:
## Subject

This is the same thing as https://github.com/sonata-project/SonataAdminBundle/pull/5733, dropping support of Symfony < 3.4 and >= 4, < 4.2

## Changelog

```markdown
### Removed
- Support for Symfony < 3.4
- Support for Symfony >= 4, < 4.2
```

~There are some changes that weren't marked as deprecated:~

~- [Removing `getName` from Form](https://github.com/symfony/symfony/blob/3.4/UPGRADE-3.0.md#form)~


src/Controller/MediaController.php
```diff
@@ -53,22 +53,22 @@ public function getMedia($id)
      *
      * @return Response
      */
-    public function downloadAction($id, $format = MediaProviderInterface::FORMAT_REFERENCE)
+    public function downloadAction(Request $request, $id, $format = MediaProviderInterface::FORMAT_REFERENCE)
```
phansys: This class is not final yet, so the signature change are a BC break.

src/Form/Type/ApiMediaType.php
```diff
@@ -68,18 +67,6 @@ public function buildForm(FormBuilderInterface $builder, array $options)
...
     }
 
-    /**
-     * {@inheritdoc}
-     *
-     * NEXT_MAJOR: remove this method.
-     *
-     * @deprecated Remove it when bumping requirements to Symfony >=2.7
-     */
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
```
phansys: Removing a public method breaks BC.

src/DependencyInjection/Compiler/SecurityContextCompilerPass.php
```diff
@@ -1,49 +0,0 @@
...
- * SecurityContextCompilerPass.
- *
- * This compiler pass provides compatibility with Symfony < 2.6 security.context service
- * and 2.6+ security.authorization_checker service. This pass may be removed when support
- * for Symfony < 2.6 is dropped.
- */
-/**
- * @final since sonata-project/media-bundle 3.21.0
- */
-class SecurityContextCompilerPass implements CompilerPassInterface
```
core23: I'm not sure if you can remove this class. It's a BC break by definition

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -527,65 +480,6 @@ public function configureExtra(ContainerBuilder $container, array $config)
         }
     }
 
-    /**
-     * Add class to compile.
-     */
-    public function configureClassesToCompile()
```
core23: BC break?

src/Form/Type/ApiMediaType.php
franmomu: I'll take a look later in the other PR about public functions

src/Controller/MediaController.php
```diff
@@ -61,14 +61,16 @@ public function downloadAction($id, $format = MediaProviderInterface::FORMAT_REF
             throw new NotFoundHttpException(sprintf('unable to find the media with the id : %s', $id));
         }
 
-        if (!$this->get('sonata.media.pool')->getDownloadSecurity($media)->isGranted($media, $this->getCurrentRequest())) {
+        $request = $this->getCurrentRequest();
+
+        if (!$this->get('sonata.media.pool')->getDownloadSecurity($media)->isGranted($media, $request)) {
```
greg0ire: Why this change?

franmomu: Not necessary, the thing is that I first changed the signature of actions to add `Request $request`  and removed `getCurrentRequest` and then to respect BC I removed the parameter again and instead of changing back to `$this->getCurrentRequest()` everywhere I reused the variable. I can change it back.

greg0ire: Please do :pray:

src/DependencyInjection/Compiler/SecurityContextCompilerPass.php
```diff
@@ -20,10 +20,12 @@
...
  * This compiler pass provides compatibility with Symfony < 2.6 security.context service
- * and 2.6+ security.authorization_checker service. This pass may be removed when support
- * for Symfony < 2.6 is dropped.
- */
-/**
+ * and 2.6+ security.authorization_checker service.
+ *
+ * NEXT_MAJOR: Remove this class
+ *
+ * @deprecated
```
phansys:
```suggestion
 * @deprecated since sonata-project/media-bundle 3.x
```

tests/Controller/GalleryAdminControllerTest.php
```diff
@@ -133,26 +129,13 @@ private function configureSetFormTheme(FormView $formView, array $formTheme): vo
     {
         $twig = $this->prophesize(Environment::class);
 
-        // Remove this trick when bumping Symfony requirement to 3.4+
-        if (method_exists(DebugCommand::class, 'getLoaderPaths')) {
-            $rendererClass = FormRenderer::class;
-        } else {
-            $rendererClass = TwigRenderer::class;
-        }
+        $rendererClass = FormRenderer::class;
```
phansys: This assignment is not required anymore.

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -528,62 +481,10 @@ public function configureExtra(ContainerBuilder $container, array $config)
     }
 
     /**
-     * Add class to compile.
+     * NEXT_MAJOR: remove this method.
```
phansys: We should also trigger explicit deprecation messages (`E_USER_DEPRECATED`).

franmomu: I guess `@deprecated` is enough

phansys: I'm not sure about that, because the conversion of this annotation in a real error depends on the usage of the [Debug component](https://symfony.com/doc/current/components/debug.html#debugging-a-class-loader) (`DebugClassLoader`). So, if the users are not using this component, they will never be notified about the deprecation.

franmomu: I followed [the contributing doc](https://github.com/sonata-project/dev-kit/blob/master/project/CONTRIBUTING.md#deprecation), if so I guess it should be changed as well

greg0ire: I think it's fine to assume this component is in use.

phansys: In that case, we should check if we can elaborate a rule for these kind of scenarios where we are deprecating a method or a class. Maybe removing the `trigger_error()` calls.

PR #1646, opened 2019-11-29
Manual merge

greg0ire:


greg0ire:
Woops, the build was failing, we have been using Symfony's test class one more time. I don't have the time to fix it right now, but let's keep this in mind.

PR #1648, opened 2019-12-02
Steal Test fixture class from Filesystem component

greg0ire:
## Subject

The component archives have recently been modified so that they no longer
contain tests. We were relying on one of the fixtures that came with
thes tests. Copying the fixture to our project fixes the issue.
Allows using the component v 4.4

See https://github.com/symfony/filesystem/commit/f903e39aca6b4a4a2dbc473a1287b79b6c41cfb8#diff-fc723d30b02a4cca7a534518111c1a66


OskarStark:
Thanks @greg0ire

PR #1651, opened 2019-12-12
fix(thumbnail): add 'addResizer' method call only if the method exists

fancyweb:
## Subject

The ThumbnailCompilerPass add the "addResizer" method call on the thumbnail service
but this method is not on the ThumbnailInterface, so there is no guarantee it actually exists.
I do not take into account the fact that the service could be a child definition, as this is not the
case in "our" implementation. If someone alters it, it is his/her responsability.

I am targeting this branch, because it is a bug fix, backward compatible.

## Changelog

```markdown
### Fixed
Add the "addResizer" method call on the thumbnail service only if the method exists.
```

### DIFF
--- a/src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
+++ b/src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
@@ -32,6 +32,10 @@ class ThumbnailCompilerPass implements CompilerPassInterface
             'sonata.media.thumbnail.format'
         );
 
+        if (!\is_callable([$container->getParameterBag()->resolveValue($definition->getClass()), 'addResizer'])) {
+            return;
+        }
+
         $taggedServices = $container->findTaggedServiceIds(
             'sonata.media.resizer'
         );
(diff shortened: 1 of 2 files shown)


src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
```diff
@@ -32,13 +32,17 @@ public function process(ContainerBuilder $container)
             'sonata.media.thumbnail.format'
         );
 
+        if (!\is_callable([$container->getParameterBag()->resolveValue($definition->getClass()), $addResizerMethod = 'addResizer'])) {
```
core23: Please do not use assigns inside an if statement

src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
```diff
@@ -32,13 +32,18 @@ public function process(ContainerBuilder $container)
             'sonata.media.thumbnail.format'
         );
 
+        $addResizerMethod = 'addResizer';
```
core23: Is there a reason why you use a variable? Can we use a constant here?

fancyweb: No other reason than to avoid the potential mistake to check a different method than the one added later. Should I just remove this variable?

greg0ire:
Triggering a new build, it's supposed to be fixed since #1658

greg0ire:
Thanks @fancyweb !

PR #1653, opened 2019-12-12
Fix Media Helper (Twig): set picture key setting will ignore other parameters like Title, Alt etc.

spagr:
## Subject

When `picture` key is set via  Twig tag  
 `{% media media, 'large' with {'picture': ['small', 'big'], 'title':'My Title', 'alt':'My alt'} %}`  
  options like `alt`, `title`, etc. are ignored and used default from media Entity.

I am targeting this branch, because it is PATCH of this version.

## Changelog

```markdown
### Fixed
- Fix arrays union order of given `$options` and default `$params` in  `src/Provider/ImageProvider.php`
```

### DIFF
--- a/src/Provider/ImageProvider.php
+++ b/src/Provider/ImageProvider.php
@@ -103,7 +103,7 @@ class ImageProvider extends FileProvider
             }
 
             unset($options['picture']);
-            $pictureParams['img'] = $params + $options;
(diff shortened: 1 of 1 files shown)


greg0ire:
@spagr can you please provide unit tests for this?

PR #1647, opened 2019-12-01
Add missing default service argument

core23:
## Subject

This bug was reporting when calling `lint:container`.

## Changelog

```markdown
### Fixed
- Add missing default service argument for `AmazonMetadataBuilder`
```

PR #1666, opened 2020-01-07
fix bc break: restore sonata.media.metadata.amazon service defaults

fancyweb:
## Subject

This PR fixes passing the default values to the `sonata.media.metadata.amazon` service. Since https://github.com/sonata-project/SonataMediaBundle/pull/1647, those defaults are passed as the second argument instead (making them useless). That regression prevent files from being publicly accessible by default in our case.

I am targeting this branch, because it fixes a bc break.

## Changelog

```markdown
### Fixed
- Fixed passing the default values to the `sonata.media.metadata.amazon` service.
```

I would appreciate a fast release ❤️

### DIFF
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -384,7 +384,7 @@ class SonataMediaExtension extends Extension implements PrependExtensionInterfac
             ;
 
             $container->getDefinition('sonata.media.metadata.amazon')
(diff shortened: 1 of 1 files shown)


greg0ire:
Here you go : https://github.com/sonata-project/SonataMediaBundle/releases/tag/3.23.0 :)

fancyweb:
Thanks @greg0ire

PR #1679, opened 2020-02-02
Console commands returning status code for Symfony4.4 and 5.0

bigfoot90:
## Subject

As of Symfony 4.4 (and required for 5.0) console commands should return integer as exit code *(`0` if  all gone ok)*

## Changelog

```markdown

### Fixed
Fixed deprecations
- `Return value of "Sonata\MediaBundle\Command\CleanMediaCommand::execute()" should always be of the type int since Symfony 4.4, NULL returned.`
- `Return value of "Sonata\MediaBundle\Command\FixMediaContextCommand::execute()" should always be of the type int since Symfony 4.4, NULL returned.`
- `Return value of "Sonata\MediaBundle\Command\RemoveThumbsCommand::execute()" should always be of the type int since Symfony 4.4, NULL returned.`
and others similar.

```

### DIFF
--- a/src/Command/AddMassMediaCommand.php
+++ b/src/Command/AddMassMediaCommand.php
@@ -72,6 +72,8 @@ class AddMassMediaCommand extends BaseCommand
         }
 
         $output->writeln('Done!');
+
+        return 0;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


core23:
Please update the PR content and use the template, this wil a help a lot for us

greg0ire:
Thanks @bigfoot90 !

Issue #1681, opened 2020-02-03
Path generator using UUID VO


bigfoot90:
I'm using Ramsey's **UUID**, I found that `Sonata\MediaBundle\Generator\ODMGenerator` works great with UUID strings *(even if you are using ORM as persistence layer)*, but doesn't work with **VO** objects because of `strict_types` declaration.
**Please note** that class `Ramsey\Uuid\Uuid` implements magic method `__tostring()`.
Someway related to https://github.com/sonata-project/SonataMediaBundle/issues/1342

#### Sonata packages

```
$ composer show --latest 'sonata-project/*'
sonata-project/media-bundle        3.23.0 3.23.0 Symfony SonataMediaBundle
```

#### PHP version

```
$ php -v
PHP 7.4.1 (cli) (built: Jan 23 2020 12:40:18)
```

## Steps to reproduce
```php
class Media extends BaseMedia
{
    /**
     * @var Uuid
     */
    private $id;

    public function __construct()
    {
        $this->id = Uuid::uuid4();
    }

    public function getId(): Uuid
    {
        return $this->id;
    }
}
```

## Expected results
```php
$pathGenerator = new ODMGenerator();
$path = $pathGenerator->generatePath($media);
```
Should return a valid path *(like '/context/1234/56/')* from entity using UUID VO.

## Actual results

`substr() expects parameter 1 to be string, object given`

## Solution
Change the line https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Generator/ODMGenerator.php#L28 into
```php
$id = (string) $media->getId();
```

greg0ire:
Please make a PR with the proposed solution 👍

OskarStark:
I am against the casting but for a method check __toString() exists. Also didn’t we do something similar a few weeks ago in the sonata codebase somewhere? ORM bundle ?

greg0ire:
@OskarStark good memory! https://github.com/sonata-project/SonataDoctrineORMAdminBundle/pull/964

PR #1683, opened 2020-02-03
Generate path when using VO as UUID representation

bigfoot90:
## Subject
Class `Sonata\MediaBundle\Generator\ODMGenerator` can now be used *(instead of making a custom generator class)* in the case your domain is using VOs instead of UUID strings.

I am targeting this branch, because no BC breaks.

Closes https://github.com/sonata-project/SonataMediaBundle/issues/1681

## Changelog

```markdown
### Fixed
- Generating path when using VO representing UUID
```

### DIFF
--- a/src/Generator/ODMGenerator.php
+++ b/src/Generator/ODMGenerator.php
@@ -25,7 +25,7 @@ class ODMGenerator implements GeneratorInterface
      */
     public function generatePath(MediaInterface $media)
     {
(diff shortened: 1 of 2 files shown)


greg0ire:
Will squash-merge when the build passes.

bigfoot90:
What is blocking here?
Do I need to do something?

PR #1688, opened 2020-02-06
Prepare generators rename

bigfoot90:
## Subject

I think these names are more consistent with current implementations as they are not related to kind of persistence by to the type ID used by the model.

I am targeting this branch, it is supposed to be BC.

Reply to https://github.com/sonata-project/SonataMediaBundle/issues/1342#issuecomment-581243831

## Changelog

```markdown
### Changed
- Renamed class `Sonata\MediaBundle\Command\DefaultGenerator` into `Sonata\MediaBundle\Command\IdGenerator`
- Renamed class `Sonata\MediaBundle\Command\ODMGenerator` into `Sonata\MediaBundle\Command\UuidGenerator`
- Renamed class `Sonata\MediaBundle\Command\PHPCRGenerator` into `Sonata\MediaBundle\Command\PathGenerator`
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -1,6 +1,18 @@
 UPGRADE 3.x
 ===========
 
+UPGRADE FROM 3.23 to 3.24
+=========================
+
+### Generators rename
+
+Some classes have been renamed, the old types are just aliases for the new ones.
+You should replace all references to them in your code.
+
+- `Sonata\MediaBundle\Command\DefaultGenerator` is deprecated in favor of `Sonata\MediaBundle\Generator\IdGenerator`
+- `Sonata\MediaBundle\Command\ODMGenerator` is deprecated in favor of `Sonata\MediaBundle\Generator\UuidGenerator`
+- `Sonata\MediaBundle\Command\PHPCRGenerator` is deprecated in favor of `Sonata\MediaBundle\Generator\PathGenerator`
+
 UPGRADE FROM 3.6 to 3.7
 =======================
 
--- a/src/Generator/IdGenerator.php
+++ b/src/Generator/IdGenerator.php
@@ -1,0 +1,50 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Generator;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+/**
+ * @final since sonata-project/media-bundle 3.21.0
+ */
+class IdGenerator implements GeneratorInterface
+{
+    /**
+     * @var int
+     */
+    protected $firstLevel;
+
+    /**
+     * @var int
+     */
+    protected $secondLevel;
+
+    /**
+     * @param int $firstLevel
+     * @param int $secondLevel
+     */
+    public function __construct($firstLevel = 100000, $secondLevel = 1000)
+    {
+        $this->firstLevel = $firstLevel;
+        $this->secondLevel = $secondLevel;
+    }
+
+    public function generatePath(MediaInterface $media)
+    {
+        $rep_first_level = (int) ($media->getId() / $this->firstLevel);
+        $rep_second_level = (int) (($media->getId() - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
+
+        return sprintf('%s/%04s/%02s', $media->getContext(), $rep_first_level + 1, $rep_second_level + 1);
+    }
+}
--- a/src/Generator/PathGenerator.php
+++ b/src/Generator/PathGenerator.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Generator;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+/**
+ * @final since sonata-project/media-bundle 3.21.0
+ */
+class PathGenerator implements GeneratorInterface
+{
+    public function generatePath(MediaInterface $media)
+    {
+        $segments = preg_split('#/#', $media->getId(), -1, PREG_SPLIT_NO_EMPTY);
+
+        if (\count($segments) > 1) {
+            // remove last part from id
+            array_pop($segments);
+            $path = implode('/', $segments);
+        } else {
+            $path = '';
+        }
+
+        return $path ? sprintf('%s/%s', $media->getContext(), $path) : $media->getContext();
+    }
+}
--- a/src/Generator/UuidGenerator.php
+++ b/src/Generator/UuidGenerator.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Generator;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+class UuidGenerator implements GeneratorInterface
+{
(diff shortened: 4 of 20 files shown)


bigfoot90:
Sorry, I don't undestand why these commits 2f0e3c3 and d6b3806 are here in this PR, they should be already merged.

core23:
Can you do this on the stable branch? Introduce the new classes and deprecate the old one. This will improve the upgrade path to the next major release.

bigfoot90:
Rebased

bigfoot90:
Rebased

src/Generator/IdGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+namespace Sonata\MediaBundle\Generator;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+/**
+ * @final since sonata-project/media-bundle 3.21.0
+ */
+class IdGenerator implements GeneratorInterface
```
greg0ire: You can make it final rightaway, it's not a BC-break

src/Generator/IdGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * @var int
+     */
+    protected $secondLevel;
+
+    /**
+     * @param int $firstLevel
+     * @param int $secondLevel
+     */
+    public function __construct($firstLevel = 100000, $secondLevel = 1000)
```
greg0ire: You can use type declarations rightaway

OskarStark: And remove the phpdoc as it does not add any value

src/Generator/IdGenerator.php
bigfoot90: It's actually extended by the deprecated `DefaultGenerator` right now.
https://github.com/sonata-project/SonataMediaBundle/pull/1688/files#diff-fcd815cd554f9f3cf32406cc1f5b1831R21

core23: You can still make this final  by proxing the old methods

greg0ire: Or better, use a `class_alias`, like in this PR: https://github.com/doctrine/persistence/pull/71

OskarStark:
PR Header and title should be updated, as its no breaking change anymore!

src/Generator/IdGenerator.php
greg0ire: @bigfoot90 don't hesitate to ask me if you want some assistance / an example for this.

src/Generator/IdGenerator.php
bigfoot90: But then also `DefaultGenerator` class will be using typed declarations.
Isn't this a BC break?

greg0ire: Ah you are correct, if we use a class aliases, you can't do this. I hadn't that in mind when writing this comment.

greg0ire: even if you use extension, in fact.

greg0ire:
The build is failing because there are self deprecation notices. This can happen either because we are testing the legacy paths without marking corresponding tests with the `@group legacy` annotation, or because there are still places in `src` where we call the legacy paths.

tests/Generator/ODMGeneratorTest.php
```diff
@@ -17,6 +17,9 @@
 use Sonata\MediaBundle\Generator\ODMGenerator;
 use Sonata\MediaBundle\Tests\Entity\Media;
 
+/**
+ * @deprecated
```
greg0ire: a test is public API, it cannot be deprecated. Let's mark it as covering a legacy path.
```suggestion
 * @group legacy
```

tests/Generator/PHPCRGeneratorTest.php
```diff
@@ -17,6 +17,9 @@
 use Sonata\MediaBundle\Generator\PHPCRGenerator;
 use Sonata\MediaBundle\Tests\Entity\Media;
 
+/**
+ * @deprecated
```
greg0ire:
```suggestion
 * @group legacy
```

tests/Generator/DefaultGeneratorTest.php
```diff
@@ -17,6 +17,9 @@
 use Sonata\MediaBundle\Generator\DefaultGenerator;
 use Sonata\MediaBundle\Tests\Entity\Media;
 
+/**
+ * @deprecated
```
greg0ire:
```suggestion
 * @group legacy
```

greg0ire:
The build is failing with only one deprecation now:  1x: The Sonata\MediaBundle\Generator\DefaultGenerator class is deprecated since sonata/media-bundle 3.4 and will be removed in 4.0. Use \Sonata\MediaBundle\Generator\IdGenerator instead.

    1x in MediaControllerTest::testGetMediaAction from Sonata\MediaBundle\Tests\Functional

It means something in that test probably needs to be changed so that the `DefaultGenerator` is no longer called.

greg0ire:
I squashed into one commit, and added more changes (some tests needed to switch to `IdGenerator`). I think the build will pass. One thing left to do: add an upgrade note in `UPGRADE-3.x.md`. Please do so, amend and force push.

greg0ire:
It passes! Don't worry about the upgrade note, I'm taking care of it for you.

UPGRADE-3.x.md
```diff
@@ -1,6 +1,18 @@
...
 
+UPGRADE FROM 3.23 to 3.24
+=========================
+
+### Generators rename
+
+Some classes have been renamed, the old types are just aliases for the new ones.
+You should replace all references to them in your code.
+
+- `Sonata\MediaBundle\Command\DefaultGenerator` is deprecated in favor of `Sonata\MediaBundle\Command\IdGenerator`
```
core23:
```suggestion
- `Sonata\MediaBundle\Command\DefaultGenerator` is deprecated in favor of `Sonata\MediaBundle\Generator\IdGenerator`
```

Same for the other

src/Generator/IdGenerator.php
```diff
@@ -0,0 +1,53 @@
...
+    {
+        $this->firstLevel = $firstLevel;
+        $this->secondLevel = $secondLevel;
+    }
+
+    /**
+     * {@inheritdoc}
```
core23: Please remove this useless comment

src/Generator/PathGenerator.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+/**
+ * @final since sonata-project/media-bundle 3.21.0
+ */
+class PathGenerator implements GeneratorInterface
+{
+    /**
+     * {@inheritdoc}
```
core23: Please remove this comment

src/Generator/UuidGenerator.php
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+namespace Sonata\MediaBundle\Generator;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+
+class UuidGenerator implements GeneratorInterface
+{
+    /**
+     * {@inheritdoc}
```
core23: ... and this

tests/Generator/IdGeneratorTest.php
```diff
@@ -0,0 +1,41 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Tests\Generator;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\MediaBundle\Generator\IdGenerator;
+use Sonata\MediaBundle\Tests\Entity\Media;
+
+class IdGeneratorTest extends TestCase
```
core23:
```suggestion
final class IdGeneratorTest extends TestCase
```

tests/Generator/PathGeneratorTest.php
```diff
@@ -0,0 +1,38 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Tests\Generator;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\MediaBundle\Generator\PathGenerator;
+use Sonata\MediaBundle\Tests\Entity\Media;
+
+class PathGeneratorTest extends TestCase
```
core23:
```suggestion
final class PathGeneratorTest extends TestCase
```

tests/Generator/UuidGeneratorTest.php
```diff
@@ -0,0 +1,32 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Tests\Generator;
+
+use PHPUnit\Framework\TestCase;
+use Sonata\MediaBundle\Generator\UuidGenerator;
+use Sonata\MediaBundle\Tests\Entity\Media;
+
+class UuidGeneratorTest extends TestCase
```
core23:
```suggestion
final class UuidGeneratorTest extends TestCase
```

greg0ire:
Thanks @bigfoot90 !

PR #1699, opened 2020-02-13
Address block bundle deprecations

franmomu:

## Changelog

```markdown
### Fixed
- BlockBundle deprecations
```

I did this in a separate PR since I'm not sure about how to deal with the [`buildEditForm`](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Block/GalleryBlockService.php#L110) and [`pre*` methods](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Block/GalleryBlockService.php#L223) since I haven't found yet where they are used. I've seen the [`EditableBlockService`](https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/Block/Service/EditableBlockService.php), but not how to use it.


core23:
Please add a changelog @franmomu

greg0ire:
@core23 why is this minor?

core23:
> @core23 why is this minor?

Thought we use this for deprecation fixes. Feel free to change the label.

greg0ire:
We use that for new deprecations so that ppl getting a patch release do not get more deprecations, but for fixes, it's patch :)

PR #1706, opened 2020-02-24
Feature/digitalocean spaces

jessequinn:
## Subject

Allows for DigitalOcean Spaces to be used based on S3.

I am targeting this branch, because it is current.

I believe this addresses no issues at the moment other than the need for Spaces integration. Based on S3 filesystem. If new adaptor is made, currently in backlog, it will be easily integrated.

### Added
- Added DigitalOcean Spaces support

    - [ ] Update the tests
    - [ ] Update the documentation


src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -422,6 +422,54 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
 
+        // add the default configuration for the Spaces filesystem
+        if ($container->hasDefinition('sonata.media.adapter.filesystem.spaces') && isset($config['filesystem']['spaces'])) {
+            $container->getDefinition('sonata.media.adapter.filesystem.spaces')
+                ->replaceArgument(0, new Reference('sonata.media.adapter.service.spaces'))
+                ->replaceArgument(1, $config['filesystem']['spaces']['bucket'])
+                ->replaceArgument(2, ['create' => $config['filesystem']['spaces']['create'], 'region' => $config['filesystem']['spaces']['region'], 'directory' => $config['filesystem']['spaces']['directory'], 'ACL' => $config['filesystem']['spaces']['acl']])
+            ;
+
+            $container->getDefinition('sonata.media.metadata.amazon')
```
OskarStark: amazon?

jessequinn: from what i understand DO Spaces is s3-compliant. However, that is a good question if we need it at all. It may be required by Guafrette.

src/DependencyInjection/Configuration.php
```diff
@@ -251,6 +251,50 @@ private function addFilesystemSection(ArrayNodeDefinition $node): void
                             ->end()
                         ->end()
 
+                        ->arrayNode('spaces')
```
core23: Can you please try to extract some small methods. This will improve readability.

core23:
Can you please finish this @jessequinn ?

anacona16:
I was able to use Sonata Media with DigitalOcean Spaces using this approach #1774

jordisala1991:
This one should not be necessary anymore

PR #1716, opened 2020-03-16
Fix mongodb invalid xml mapping Error

### LINKED ISSUE #1712: Invalid mapping file BaseMedia.mongodb.xml
have installed Sonata Media Bundle ^3.23, when installing doctrine-mongodb-bundle ^4.1, i receive the below error:

In MappingException.php line 236 The mapping file vendor/sonata-project/media-bundle/src/Resources/config/doctrine/BaseMedia.mongodb.xml is invalid.

SylvanoTombo:
## Mongodb xml mapping error

Edit Mongodb xml mapping and change "FieldName" attribute to "field-name"

I am targeting this branch, because want to push this correction to the next release.

Closes #1712

## Changelog
Update invalid mongodb xml mapping for BaseGallery.mongodb.xml and BaseMedia.mongodb.xml files

### DIFF
--- a/src/Resources/config/doctrine/BaseGallery.mongodb.xml
+++ b/src/Resources/config/doctrine/BaseGallery.mongodb.xml
@@ -1,12 +1,12 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <doctrine-mongo-mapping xmlns="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping                         http://doctrine-project.org/schemas/odm/doctrine-mongo-mapping.xsd">
     <mapped-superclass name="Sonata\MediaBundle\Document\BaseGallery">
-        <field name="name" fieldName="name" type="string"/>
-        <field name="context" fieldName="context" type="string"/>
-        <field name="defaultFormat" fieldName="defaultFormat" type="string"/>
-        <field name="enabled" fieldName="enabled" type="boolean"/>
-        <field name="updatedAt" fieldName="updatedAt" type="date"/>
-        <field name="createdAt" fieldName="createdAt" type="date"/>
+        <field name="name" field-name="name" type="string"/>
+        <field name="context" field-name="context" type="string"/>
+        <field name="defaultFormat" field-name="defaultFormat" type="string"/>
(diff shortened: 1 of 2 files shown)


SylvanoTombo:
No change in the composer.json file..i only changed the mapping file.

malarzm:
Hey ODM maintainer here.

> No change in the composer.json file?

We've changed the `fieldName` to `field-name` in 2.0 of the ODM and we're also allowing to use both attributes in version 1.3. For version 1.2 and below the `fieldName` must be used as there's no compat layer.

VincentLanglet:
@core23 Since https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/blob/3.x/composer.json require 1.3, this should be ok.

greg0ire:
Please associate your email address with your Github account, or change the
email in your commits to an address already associated with it. If you do not
want to expose your personal email address, you may use
`$userName@users.noreply.github.com`, that way we can still reach you through Github.
Additionally, you will get credit for this contribution on your Github profile.

SylvanoTombo:
@core23, Sorry for requesting new review but i deleted unecessary commit.

VincentLanglet:
> AFAIK, this bundle should be able to work without "sonata-project/doctrine-mongodb-admin-bundle". So, if you need to ensure a given syntax is supported by "doctrine/mongodb-odm", that requirement must be explicitly declared in this package.

In the documentation I saw that `sonata-project/doctrine-mongodb-admin-bundle` is an optional dependency, like others SonataORM bundles. This file is certainly used only by the mongo bundle.

phansys:
The base model provided here (and its mapping) can be used by any project outside the Sonata ecosystem. IMO there should be at least an advise about the compatible versions of "doctrine/mongodb-odm".
I know declaring a conflict against `<1.3` may not be the best choice, since this package could be used in environments where "doctrine/mongodb-odm" is present but no necessarily associated with this bundle; but at code level I don't see a better alternative.
Maybe a documentation entry could be enough?

greg0ire:
A conflict is indeed how we usually handle that kind of optional dependency.

VincentLanglet:
@SylvanoTombo Can you add a conflict with `sonata-project/doctrine-mongodb-admin-bundle < 1.3` in the composer.json ?

franmomu:
Re-build

franmomu (approved):
This should be fine after https://github.com/sonata-project/SonataMediaBundle/pull/1818

core23 (approved):

phansys (approved):

phansys:
Thank you @SylvanoTombo!

PR #1717, opened 2020-03-21
Allow Twig 3

phansys:
## Subject
Allow Twig 3.

## Changelog

```markdown
### Added
- Support for "twig/twig:^3.0".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
         "symfony/twig-bridge": "^4.3",
         "symfony/twig-bundle": "^4.3",
         "symfony/validator": "^4.3",
-        "twig/twig": "^2.10"
+        "twig/twig": "^2.10 || ^3.0"
     },
     "conflict": {
         "doctrine/dbal": "<2.6.0",

### CODE REVIEW

**Comment by greg0ire** [comment]
Thanks @phansys !

**Comment by franmomu** [comment]
I'm not sure about this, shouldn't be tests changed to actually use `twig/twig: ^3` (which is not possible right now because of dependencies)? There could be hidden incompatibilities I guess.

**Comment by greg0ire** [comment]
Ow… are there dependencies still using Twig 2? If yes this is not great indeed.

> which is not possible right now because of dependencies

It's possible, and we do it in many jobs with inline aliases: `use twig/twig ^3.0 as 2`. But then our dependencies could crash because of that


```
sonata-project/admin-bundle      3.63.0  requires  twig/twig (^2.10)          
sonata-project/block-bundle      3.18.4  requires  twig/twig (^1.34 || ^2.0)  
sonata-project/core-bundle       3.18.0  requires  twig/twig (^1.34 || ^2.0)  
sonata-project/formatter-bundle  4.1.3   requires  twig/twig (^2.4)           
twig/extensions                  v1.5.4  requires  twig/twig (^1.27|^2.0)
```

phansys:
>Maybe we should revert?

Indeed! In fact yesterday after creating this PR I followed the one of the reason for this conflict to "sonata-project/formatter-bundle". Seems that I forgot to back here and inform about the situation.
Surely we should apply this later in `master`.
Sorry for the inconvenience.

PR #1719, opened 2020-03-22
Clean up tests

franmomu:
- Removed some deprecations
- Use Container instead of the mock to simplify

### DIFF
--- a/tests/App/config.yml
+++ b/tests/App/config.yml
@@ -6,8 +6,10 @@ framework:
         enabled: true
 
 twig:
+    exception_controller: null
     paths:
         - '%kernel.project_dir%/templates'
+    strict_variables: "%kernel.debug%"
 
 sonata_media:
     db_driver: doctrine_orm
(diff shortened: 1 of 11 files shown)


tests/App/config.yml
```diff
@@ -6,8 +6,10 @@ framework:
         enabled: true
 
 twig:
+    exception_controller: null
```
phansys:
```suggestion
    exception_controller: ~
```

franmomu: I used to use `~`, but I've seen sometimes it is not clear its meaning, so I think using `null` is more straightforward.

greg0ire: @phansys I'll let you merge if you agree, I don't have a strong opinion on this.

greg0ire: @franmomu do you mean the specification says it should be interpreted as null in some contexts, or just that not everybody know that this means `null`?

franmomu: > @franmomu do you mean the specification says it should be interpreted as null in some contexts, or just that not everybody know that this means `null`?

The second one, apparently looks like Symfony were changing `~` to `null` in the recipes https://github.com/symfony/recipes/pull/635

phansys: Let's merge then :+1:

phansys:
Thank you @franmomu!

PR #1720, opened 2020-03-22
Revert "Allow Twig 3"

greg0ire:
This reverts commit 7c12b071c337e1e32eb407fb6e4785d1968ee782.
We cannot effectively test with Twig 3 since many of our dependencies
are not compatible with it yet.
I'm not adding a changelog since #1717 is not released yet. Furthermore, I am going to remove the changelog and label from #1717


greg0ire:
Didn't notice, sorry!

phansys:
No problem at all! :smile:

PR #1723, opened 2020-03-23
Document how to inject $mediaManager in Symfony 4+ projects

maks-rafalko:
Document how to inject $mediaManager in Symfony 4+ projects, since after reading "Installation" and "Usage" articles, it's not clear how to get `$mediaManager` inside any class.

For example, there is a similar answer in SO: https://stackoverflow.com/a/58296157/2579875

This solution uses interface though rather than concrete implementation

### DIFF
--- a/docs/reference/usage.rst
+++ b/docs/reference/usage.rst
@@ -26,6 +26,19 @@ or this::
 
     $mediaManager->save($media, 'user', 'sonata.media.provider.youtube');
 
+``$mediaManager`` is an instance of ``Sonata\MediaBundle\Model\MediaManagerInterface``. In order to us it, inject it into constructor of a class::
+
+    public function __construct(MediaManagerInterface $mediaManager)
+    {
+        $this->mediaManager = $mediaManager;
+    }
+
+And create an alias so that Symfony understands what implementation to use::
+
+    // config/services.yaml
+
+    Sonata\MediaBundle\Model\MediaManagerInterface: '@sonata.media.manager.media'
+
 Retrieving metadata information
 -------------------------------
 


maks-rafalko:
Hm.. let me disagree.

1. I see the code of the old projects (symfony 2.*) where it worked out of the box (I think with compiler passes - don't know)

2. now (with Symfony 4.*) I *need to be aware of it* and do something that is not in the MediaBundle docs

3. Believe me, it took me some time to understand why this service is not available immediately after installation, and it will take literally (tens of) minutes to understand it for junior devs.

Don't we want to have the docs that just can be copy pasted to make everything work?

VincentLanglet:
When I see your documentation, I just an example of autowiring.
https://symfony.com/doc/current/service_container.html#injecting-services-config-into-a-service

First, it's symfony standard.
Second, if the dev does not use autowiring, the dev will be lost.

IMO, I prefer the dev to look at symfony doc if he wants to use a service.

maks-rafalko:
This PR is not just about DI and infection. It's about an Interface and its *many* implementations:

https://symfony.com/doc/current/service_container/autowiring.html#dealing-with-multiple-implementations-of-the-same-type

and this is not so clear for newcomers, especially when we have Media Bundle with its hidden services on top of Symfony

VincentLanglet:
You're making my point
- There is a symfony doc about this
- This works only if you use autowiring

The documentation of Sonata is made to explain how to use Sonata, not how to use Symfony.

maks-rafalko:
Ok, even if we forget about DI and Symfony, do you think it's clear here https://sonata-project.org/bundles/media/3-x/doc/reference/usage.html that `$mediaManager` is an instance of `Sonata\MediaBundle\Model\MediaManagerInterface`, but not `A\b\SomeClass`?

VincentLanglet:
> If you don't want to add DI-related things to the docs, at least be so kind to add a type of the variables that are used in the snippets. Nobody is a wizard here.

Such a PR would be accepted I think.

maks-rafalko:
See https://github.com/sonata-project/SonataMediaBundle/pull/1724

PR #1739, opened 2020-05-03
Fix missing translation of gallery context

core23:
## Subject

Without the change, `messages` would be used as default translation domain. This is wrong, because we always use `SonataMediaBundle` as domain when translating the context.

## Changelog

```markdown
### Fixed
- Fix missing translation of gallery context
```


greg0ire:
Thanks @core23 !

PR #1744, opened 2020-05-20
Make CoreBundle optional

wbloszyk:
## Subject

I am targeting this branch, because this changes are BC.

Part of https://github.com/sonata-project/dev-kit/issues/697

## Changelog

```markdown
### Removed
- remove SonataCoreBundle dependencies
```


src/Provider/MetadataInterface.php
```diff
@@ -13,27 +13,55 @@
...
- * NEXT_MAJOR: Remove CoreBundle dependency.
- */
-interface MetadataInterface extends \Sonata\CoreBundle\Model\MetadataInterface
-{
-    public function getTitle(): string;
+if (interface_exists('Sonata\CoreBundle\Model\MetadataInterface')) {
+    /**
+     * NEXT_MAJOR: Remove CoreBundle dependency.
+     */
+    interface MetadataInterface extends \Sonata\CoreBundle\Model\MetadataInterface
```
jordisala1991, line 20: What is the usage of this Metadata? According to CoreBundle you should extend the one on BlockBundle

wbloszyk: I don't know.

wbloszyk: Anyway decprecation message tell US we can use BlockBundle matsdata instead.

core23: > What is the usage of this Metadata? According to CoreBundle you should extend the one on BlockBundle

That's not right. There are different Metadatas: Block information metadata and media provider metadata.

jordisala1991: I was looking at this comment:  @deprecated since sonata-project/core-bundle 3.13.0, to be removed in 4.0. Use Sonata\BlockBundle\Meta\MetadataInterface instead.

But I dont really know what is the use for MetadataInterface

wbloszyk: i think this is another reason to create `sonata-project/contracts`

jordisala1991: Well, maybe we don't have to extend the block bundle one, I was just following the comments on the code. But it is correct that it doesn't have much sense

wbloszyk: Metadata is generated for each media by provider. It have sense for `MediaBlockService` but also make it more confusing.
For me we should keep MediaMetadata and Block Metadata seperate but based on `contracts/metadata`. This will keep desteny in naming and allow use metadata in other places.

src/Provider/MetadataInterface.php
```diff
@@ -13,27 +13,58 @@
 
 namespace Sonata\MediaBundle\Provider;
 
-/**
- * NEXT_MAJOR: Remove CoreBundle dependency.
- */
-interface MetadataInterface extends \Sonata\CoreBundle\Model\MetadataInterface
-{
-    public function getTitle(): string;
+use Sonata\BlockBundle\Meta\MetadataInterface as BlockMetadataInterface;
```
core23, line 16: This one can be simplified:

```php
if (interface_exists(CoreMetadataInterface::class)) {
   interface InternalMetadataInterface extends CoreMetadataInterface, BlockMetadataInterface {}
 else {
   interface InternalMetadataInterface extends BlockMetadataInterface {}
}

interface MetadataInterface extends InternalMetadataInterface {
// ...
}
```

jordisala1991: I suggested to bind this metadata to the block bundle one, but maybe it is not necessary as you pointed out.

Should we remove that extends?

core23: If we follow BC very strict, we could not even remove the `CoreMetadataInterface` extension :/

But I would risk that case. If the user checks somewhere against for a `CoreMetadataInterface`, they should also require the CoreBundle in their `composer.json` file.

composer.json
```diff
@@ -33,9 +33,11 @@
         "knplabs/gaufrette": "^0.8 || ^0.9",
         "kriswallsmith/buzz": "^0.15 || ^0.16",
         "psr/log": "^1.0",
-        "sonata-project/core-bundle": "^3.14",
+        "sonata-project/block-bundle": "^3.17",
```
jordisala1991, line 36: not sure about this one... this dependency was optional, and not it is required. Not sure we have to link the metadatainterface to the block bundle

wbloszyk: https://github.com/sonata-project/SonataBlockBundle/pull/506
https://github.com/sonata-project/SonataMediaBundle/commit/206b4cb26065fbd887bebc40a791aef6b56b15a9
https://github.com/sonata-project/SonataMediaBundle/commit/5cda60ead57473db7f31db1e3bcab3698f427c11

@core23 
Mybe you remeber something? Should we keep it?  

For me it is another reason to discouse about: https://github.com/sonata-project/dev-kit/issues/658. 

This will be awosome feature to create auto generating things like media put to block and display in site or site composer  (PageBundle) based on metadatas with some context.

src/Provider/MetadataInterface.php
```diff
@@ -13,16 +13,27 @@
 
 namespace Sonata\MediaBundle\Provider;
 
-/**
- * NEXT_MAJOR: Remove CoreBundle dependency.
- */
-interface MetadataInterface extends \Sonata\CoreBundle\Model\MetadataInterface
+use Sonata\BlockBundle\Meta\MetadataInterface as BlockMetadataInterface;
```
core23, line 16: This is wrong. Should be: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Provider/MetadataInterface.php

composer.json
core23: You're right. The dependency should be optional. There is no need for the block bundle, see: https://github.com/sonata-project/SonataMediaBundle/pull/1744#discussion_r437643302

core23:
Thanks @wbloszyk

PR #1748, opened 2020-06-05
Update custom service provider configuration documentation

hug963:
## Subject

I had an issue with my custom provider using the S3 filesystem.
The files were not publicly readable, which turned out to be because [this line](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Metadata/ProxyMetadataBuilder.php#L55) returned false and led to empty metadata.

This happened because my service was not public.

This updates the doc to a more modern yaml config and sets the service as public.

### DIFF
--- a/docs/reference/creating_a_provider_class.rst
+++ b/docs/reference/creating_a_provider_class.rst
@@ -253,29 +253,31 @@ Register the Class with the Service Container
 If you use the tag ``sonata.media.provider``, the provider service will be
 added to the provider pool.
 
-.. code-block:: xml
+.. code-block:: yaml
 
-    <!-- config/services.xml -->
+    # config/services.yaml
 
-    <service id="sonata.media.provider.vimeo" class="Sonata\MediaBundle\Provider\VimeoProvider">
-        <argument>sonata.media.provider.vimeo</argument>
-        <argument type="service" id="sonata.media.filesystem.local"/>
-        <argument type="service" id="sonata.media.cdn.server"/>
-        <argument type="service" id="sonata.media.generator.default"/>
-        <argument type="service" id="sonata.media.thumbnail.format"/>
-        <argument type="service" id="sonata.media.buzz.browser"/>
-        <argument type="service" id="sonata.media.metadata.proxy"/>
-        <call method="setTemplates">
-            <argument type="collection">
-                <argument key='helper_thumbnail'>@SonataMedia/Provider/thumbnail.html.twig</argument>
-                <argument key='helper_view'>@SonataMedia/Provider/view_vimeo.html.twig</argument>
-            </argument>
-        </call>
-        <call method="setResizer">
-            <argument type="service" id="sonata.media.resizer.simple"/>
-        </call>
-        <tag name="sonata.media.provider"/>
-    </service>
+    sonata.media.provider.vimeo:
+        class: Sonata\MediaBundle\Provider\VimeoProvider
+        arguments:
+            - sonata.media.provider.vimeo
+            - '@sonata.media.filesystem.local'
+            - '@sonata.media.cdn.server'
+            - '@sonata.media.generator.default'
+            - '@sonata.media.thumbnail.format'
+            - '@sonata.media.buzz.browser'
+            - '@sonata.media.metadata.proxy'
+        calls:
+            -
+                - setTemplates
+                - - helper_thumbnail: '@@SonataMedia/Provider/thumbnail.html.twig'
+                    helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig'
+            -
+                - setResizer
+                - ['@sonata.media.resizer.simple']
+        tags:
+            - { name: sonata.media.provider }
+        public: true
 
 The last important part is how the vimeo media should be displayed.
 


docs/reference/creating_a_provider_class.rst
```diff
@@ -253,29 +253,30 @@ Register the Class with the Service Container
...
+            - sonata.media.provider.vimeo
+            - '@sonata.media.filesystem.local'
+            - '@sonata.media.cdn.server'
+            - '@sonata.media.generator.default'
+            - '@sonata.media.thumbnail.format'
+            - '@sonata.media.buzz.browser'
+            - '@sonata.media.metadata.proxy'
+
+        calls:
+            - [setTemplates, [{ helper_thumbnail: '@@SonataMedia/Provider/thumbnail.html.twig', helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig' }]]
```
greg0ire, line 273: Can you please use a multiline equivalent here? Overall YAML is easier to read here, but this isn't, especially with the double `@`…

hug963: Hi,

Yaml is not really well suited for this kind of structure, this is the best I could come up with:

``` yaml
        calls:
        -
            - setTemplates
            -
                -
                    helper_thumbnail: @@SonataMedia/Provider/thumbnail.html.twig'
                    helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig'
        -
            - setResizer
            -
                - '@sonata.media.resizer.simple'
```

or more concise but even less understandable

```yaml
        calls:
        -
            - setTemplates
            - - helper_thumbnail: '@@SonataMedia/Provider/thumbnail.html.twig'
                helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig'
```

I think this looks a bit crazy. :)

Either we leave as is, or I just update the xml version.
But Symfony removed most docs for the xml configs, so it would be good to migrate to yaml anyway...

greg0ire: > I think this looks a bit crazy. :)

Maybe that's because you are not used to multiline syntax? The second one looks sensible to me. More than what you originally proposed anyway.

hug963: I updated the docs with the multi line syntax.

VincentLanglet:
Hi @greg0ire Is it ok for you ? :)

docs/reference/creating_a_provider_class.rst
```diff
@@ -253,29 +253,35 @@ Register the Class with the Service Container
...
+            - '@sonata.media.generator.default'
+            - '@sonata.media.thumbnail.format'
+            - '@sonata.media.buzz.browser'
+            - '@sonata.media.metadata.proxy'
+
+        calls:
+            -
+                - setTemplates
+                - - helper_thumbnail: '@@SonataMedia/Provider/thumbnail.html.twig'
+                    helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig'
```
phansys, line 276: Regardless the result is the same, maybe the "expanded" syntax is easier to understand. IMO, the explicit indentation provides a clearer example about the nodes and hierarchy. 
```suggestion
                - 
                    -
                        helper_thumbnail: '@@SonataMedia/Provider/thumbnail.html.twig'
                        helper_view: '@@SonataMedia/Provider/view_vimeo.html.twig'
```

docs/reference/creating_a_provider_class.rst
```diff
@@ -253,29 +253,35 @@ Register the Class with the Service Container
...
-        </call>
-        <tag name="sonata.media.provider"/>
-    </service>
+.. code-block:: yaml
+
+    # config/services.yaml
+
+    sonata.media.provider.vimeo:
+        class: Sonata\MediaBundle\Provider\VimeoProvider
+
```
phansys, line 262: We could remove these empty LFs.

docs/reference/creating_a_provider_class.rst
VincentLanglet: You should decide with @greg0ire 😄
See https://github.com/sonata-project/SonataMediaBundle/pull/1748#discussion_r436097655

greg0ire: I'm fine with either, the one that is not comfortable with that is OP (many devs have a phobia of linebreaks for some reason :stuck_out_tongue: )

VincentLanglet: Every new line should be payed to someone 😆

jordisala1991: I didnt understand, is the current code acceptable or there is a need to change something?

greg0ire: It's acceptable to me, but not to @phansys

phansys: That was a suggestion. I don't consider this as a blocker. If you agree this way we are fine, just dismiss my suggestion :+1:

docs/reference/creating_a_provider_class.rst
```diff
@@ -253,29 +253,35 @@ Register the Class with the Service Container
...
+        arguments:
+            - sonata.media.provider.vimeo
+            - '@sonata.media.filesystem.local'
```
VincentLanglet, line 265: Shouldn't the lines have the same format ?
```suggestion
            - '@sonata.media.provider.vimeo'
            - '@sonata.media.filesystem.local'
```

VincentLanglet: Oh my bad, the first argument was a string

VincentLanglet:
@hug963 Can you remove the empty LF as requested by @phansys ?
Then we'll merge this :)

hug963:
Hi @VincentLanglet , I removed the LFs.

jordisala1991:
Thank you @hug963

PR #1750, opened 2020-06-08
Document the media-orm-pack and media-odm-pack flex recipes

hug963:
## Subject

There is no mention in the docs fo the `media-orm-pack` and `media-odm-pack` flex recipes.
These are really practical and take care of the whole configuration automatically.


core23:
Thanks @hug963

PR #1756, opened 2020-06-20
Replace truncate filter with u filter

franmomu:
## Subject

This bundle was using `truncate` filter without explicitly require `twig/extensions`, with this PR it uses `u` filter.

~I added `twig/extra-bundle` to autoload the Twig extension and a fallback in case Flex is not used.~

Ref: https://github.com/sonata-project/SonataAdminBundle/pull/6132

I am targeting this branch, because these changes are BC.

## Changelog

```markdown
### Added
- Added `twig/string-extra` dependency.
### Changed
- Changed use of `truncate` filter with `u` filter.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
         "symfony/twig-bridge": "^4.3",
         "symfony/twig-bundle": "^4.3",
         "symfony/validator": "^4.3",
-        "twig/twig": "^2.10"
+        "twig/string-extra": "^3.0",
+        "twig/twig": "^2.12.1"
     },
     "conflict": {
         "doctrine/dbal": "<2.6.0",
@@ -99,7 +100,8 @@
         "sonata-project/notification-bundle": "If you want to generate thumbnails asynchronously.",
         "sonata-project/seo-bundle": "^2.1",
         "symfony/monolog-bundle": "If you want to log exceptions produced when dealing with images.",
-        "tilleuls/ckeditor-sonata-media-bundle": "^1.0"
+        "tilleuls/ckeditor-sonata-media-bundle": "^1.0",
+        "twig/extra-bundle": "Auto configures the Twig Intl extension"
     },
     "config": {
         "sort-packages": true
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -23,6 +23,7 @@ use Symfony\Component\DependencyInjection\Extension\PrependExtensionInterface;
 use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Twig\Extra\String\StringExtension;
 
 /**
  * @final since sonata-project/media-bundle 3.21.0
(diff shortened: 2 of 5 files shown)


wbloszyk:
@franmomu What with `"symfony/string": "^5.1"`? It is not require here like in AdminBundle?
https://github.com/sonata-project/SonataAdminBundle/pull/6132#discussion_r436359951

composer.json
```diff
@@ -57,7 +57,9 @@
         "symfony/twig-bridge": "^4.3",
         "symfony/twig-bundle": "^4.3",
         "symfony/validator": "^4.3",
-        "twig/twig": "^2.10"
+        "twig/extra-bundle": "^3.0",
```
wbloszyk, line 60:
```suggestion
        "symfony/string": "^5.1",
        "twig/extra-bundle": "^3.0",
```

wbloszyk: I think it is require like in https://github.com/sonata-project/SonataAdminBundle/pull/6132#discussion_r436359951

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -123,6 +124,10 @@ public function load(array $configs, ContainerBuilder $container)
                 ->replaceArgument(2, $sonataRoles);
         }
 
+        if (!isset($bundles['TwigExtraBundle'])) {
```
phansys, line 127: Instead of checking the bundle is registered, I think we could check something more specific.
By instance: https://github.com/sonata-project/SonataIntlBundle/pull/302/files#diff-0501290a6a6012a0743dd122a7d2ff14R122-R137.
I prefer that kind of check since the service can be overridden by users and the bundle availability doesn't guarantee the service integrity.
WDYT?

composer.json
phansys: We should try to be consistent with the decision about this dependency. See https://github.com/sonata-project/SonataIntlBundle/pull/302#discussion_r439797111.

franmomu: > I think it is require like in [sonata-project/SonataAdminBundle#6132 (comment)](https://github.com/sonata-project/SonataAdminBundle/pull/6132#discussion_r436359951)

Not in this case, here we just use the first parameter of the filter (`string|truncate(10)`), so any version of `symfony/string` is valid. In SonataAdmin we use the third one (`cut` parameter previously known as `preserve`) that it's only available in `5.1`.

src/DependencyInjection/SonataMediaExtension.php
franmomu: Sure, let me change it.

franmomu: So, should we require `twig/extra-bundle` or just suggest it?

phansys: Based on https://github.com/sonata-project/SonataMediaBundle/pull/1756#discussion_r443121733, I'd say we should suggest this dependency.

phansys:
Could you please update the PR description according the last change?.

jordisala1991:
Thank you @franmomu

wbloszyk:
@franmomu I think we make mistake here. `configureStringExtension()` must be rewrite to CompilerPass, otherwise someone can register here, as `StringExtension::class` and `twig.extensions.string` or other service. `twig.extension` will register twice some class and throw error:
```
Unable to register extension "Twig\Extra\String\StringExtension" as it is already registered.
```

PR #1758, opened 2020-06-26
Fix StringExtension registration

wbloszyk:
## Subject

This PR contains:
- move StringExtension registration to Compiler pass and check all `twig.extension` service to avoid duplication
- improve tests
- fix previus broken checking
```php
... !is_a($container->getDefinition('twig.extension.string')->getClass(), StringExtension::class)) ...
// it is like this:   !is_a('Twig\Extra\String\StringExtension', StringExtension::class)) ...
```

I am targeting this branch, because this change respect BC.

PATCH for non release #1756.

PR #1760, opened 2020-07-04
Fix TwigStringExtensionCompilerPass priority

wbloszyk:
## Subject

This PR fix case when TwigStringExtensionCompilerPass is build after TwigEnvironmentPass - then u filter is not registred.

I am targeting this branch, because it is PATCH.

Changelog is not require. It is fix for non release PR #1758 and #1756.


jordisala1991:
Marked as patch but no changelog is required if the other PRs are non released

PR #1767, opened 2020-07-16
Allow "friendsofsymfony/rest-bundle:^3.0"

wbloszyk:
## Subject

Allow "friendsofsymfony/rest-bundle:^3.0".

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Added
- Added support for "friendsofsymfony/rest-bundle:^3.0"
### Change
- support for deprecated "rest" routing type in favor for xml
```

PR #1779, opened 2020-08-03
Revert added type hints

wbloszyk:
## Subject

This PR revert type hints added in docblock which should not be added. They will be add corectly in other PR like https://github.com/sonata-project/SonataUserBundle/pull/1198.

I am targeting this branch, because it is FIX before release.

Changelog is not needed becouse this is fix for non release PR.

### DIFF
--- a/src/Controller/Api/GalleryController.php
+++ b/src/Controller/Api/GalleryController.php
@@ -129,7 +129,7 @@ class GalleryController
      *
      * @Rest\View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param int $id Gallery identifier
+     * @param $id
      *
      * @return GalleryInterface
      */
@@ -154,7 +154,7 @@ class GalleryController
      *
      * @Rest\View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param int $id Gallery identifier
+     * @param $id
      *
      * @return MediaInterface[]
      */
@@ -186,7 +186,7 @@ class GalleryController
      *
      * @Rest\View(serializerGroups={"sonata_api_read"}, serializerEnableMaxDepthChecks=true)
      *
-     * @param int $id Gallery identifier
+     * @param $id
      *
      * @return GalleryHasMediaInterface[]
      */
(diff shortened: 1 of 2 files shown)


wbloszyk:
@phansys @VincentLanglet @core23 @jordisala1991 @greg0ire Can you merge this PR? It is last commit require to release MediaBundle. It will be nice to have it before tomorrow work.

src/Controller/Api/GalleryController.php
```diff
@@ -339,8 +339,8 @@ public function putGalleryMediaGalleryhasmediaAction($galleryId, $mediaId, Reque
      *  }
      * )
      *
-     * @param int $galleryId A gallery identifier
-     * @param int $mediaId   A media identifier
+     * @param int $galleryId Gallery identifier
```
jordisala1991, line 342: remove int?

wbloszyk: I only revert change from https://github.com/sonata-project/SonataMediaBundle/pull/1767.

jordisala1991: I think we should at least be consistent on this file, are we leaving the int or removing it? And the same for the property description, If we are leaving it, dont remove it for some cases

wbloszyk: Some bundles must be release one by one to allow use `FOS/RestBundle` becouse of dependencies. Upgrade(and fix) RestApi consists from spets like add `fos/rest-bundle` v3 support,  `nelmio/api-doc-bundle` v3 support, add support for string identifier (like UUID), functional test. It is hard to do it in one PR(and shouldn't be do it). In this release we should add support for `fos/rest-bundle` v3 and do nothing else (like add, fix or remove other docblock). I mean this should be trait as two indepentend things. The goal will be support `RestBundle` and `ApiDocBundle` v2 and v3, and support for diffrence identifier(int, UUID, string ....). I know it is not good but keep it not working it not too.

wbloszyk:
Docblocks will be add in diffrence PR with some change. For now only integer (like ApiDoc annotation require) is allowed. Technicly identifier can be UUID or string and this will also be change. We should not release docblock type hints and change in in other next release.

wbloszyk:
@jordisala1991 Can we merge it?

jordisala1991:
Yes sure

PR #1781, opened 2020-08-10
Enable autowiring for "sonata.media.manager.media" and "sonata.media.manager.gallery" services

phansys:
## Subject
Enable autowiring for "sonata.media.manager.media" and "sonata.media.manager.gallery" services.

## Changelog

```markdown
### Added
- Added "Sonata\MediaBundle\Model\MediaManagerInterface" service alias in order to enable autowiring for "sonata.media.manager.media" service;
- Added "Sonata\MediaBundle\Model\GalleryManagerInterface" service alias in order to enable autowiring for "sonata.media.manager.gallery" service.
```

### DIFF
--- a/src/Resources/config/doctrine_mongodb.xml
+++ b/src/Resources/config/doctrine_mongodb.xml
@@ -10,10 +10,12 @@
             <argument>%sonata.media.media.class%</argument>
             <argument type="service" id="doctrine_mongodb"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\MediaManagerInterface" alias="sonata.media.manager.media" public="false"/>
         <service id="sonata.media.manager.gallery" class="%sonata.media.manager.gallery.class%" public="true">
             <argument>%sonata.media.gallery.class%</argument>
             <argument type="service" id="doctrine_mongodb"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\GalleryManagerInterface" alias="sonata.media.manager.gallery" public="false"/>
         <!-- Path generator servive -->
         <service id="sonata.media.generator.default" class="Sonata\MediaBundle\Generator\UuidGenerator">
 
--- a/src/Resources/config/doctrine_orm.xml
+++ b/src/Resources/config/doctrine_orm.xml
@@ -9,10 +9,12 @@
             <argument>%sonata.media.media.class%</argument>
             <argument type="service" id="doctrine"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\MediaManagerInterface" alias="sonata.media.manager.media" public="false"/>
         <service id="sonata.media.manager.gallery" class="%sonata.media.manager.gallery.class%" public="true">
             <argument>%sonata.media.gallery.class%</argument>
             <argument type="service" id="doctrine"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\GalleryManagerInterface" alias="sonata.media.manager.gallery" public="false"/>
         <!-- Path generator service -->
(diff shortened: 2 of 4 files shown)


jordisala1991:
https://github.com/symfony/symfony-docs/pull/9886#issuecomment-399927872

I don't really know if we should do this move.

phansys:
>I don't really know if we should do this move.

These aliases are intended to allow consumers to use autowiring, not to be used internally.

jordisala1991:
@greg0ire maybe you can help here, what is the conclusion, should we use what you mention here: https://github.com/symfony/symfony-docs/pull/9886#issuecomment-395235683 (which is what @phansys is doing here) or we should not add FQCN services (or alias)

greg0ire:
I think what @phansys did here is fine: we are not trying to use autowiring, but to let our users do so, right?

phansys:
>I think what @phansys did here is fine: we are not trying to use autowiring, but to let our users do so, right?

Yes, that is the purpose of this PR.

wbloszyk:
We should do it if people can't create own managers by this interfaces, otherwise this can create problems for this people when using autowiring. For now class FQNC should work, isn't it?

src/Resources/config/doctrine_mongodb.xml
```diff
@@ -10,10 +10,12 @@
             <argument>%sonata.media.media.class%</argument>
             <argument type="service" id="doctrine_mongodb"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\MediaManagerInterface" alias="sonata.media.manager.media"/>
```
core23, line 13: public=true ?

phansys: It's inheriting the visibility from the aliased service. Should I make the visibility explicit?

wbloszyk: It is interface alias and should not be public if IT work as private.

src/Resources/config/doctrine_mongodb.xml
```diff
@@ -10,10 +10,12 @@
             <argument>%sonata.media.media.class%</argument>
             <argument type="service" id="doctrine_mongodb"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\MediaManagerInterface" alias="sonata.media.manager.media" public="false"/>
```
wbloszyk, line 13: Public=false is not nesesery becouse it is default.

src/Resources/config/doctrine_mongodb.xml
```diff
@@ -10,10 +10,12 @@
...
         <service id="sonata.media.manager.gallery" class="%sonata.media.manager.gallery.class%" public="true">
             <argument>%sonata.media.gallery.class%</argument>
             <argument type="service" id="doctrine_mongodb"/>
         </service>
+        <service id="Sonata\MediaBundle\Model\GalleryManagerInterface" alias="sonata.media.manager.gallery" public="false"/>
```
wbloszyk, line 18: Too

src/Resources/config/doctrine_mongodb.xml
phansys: Did you read this comment? https://github.com/sonata-project/SonataMediaBundle/pull/1781#discussion_r468680310.
The alias inherits the visibility from the service. Did you experienced a different behavior?

wbloszyk: I always use private service with public alias. Can you look at `Session`? Using `SessionInterface` with autowiring is even forced.

PR #1801, opened 2020-09-02
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3

wbloszyk:
## Subject
Update conflict constraint for "nelmio/api-doc-bundle" in order to allow version 3.

I am targeting this branch, because these changes almost respect BC (with one, very little BC-break potential).

## Changelog

```markdown
### Added
- support for `nelmio/api-doc-bundle` >= 3.6
### Changed
- Updated docs in order to expose how to configure custom serialization paths under jms_serializer configuration node (jms_serializer.metadata.directories).
```

PR #1807, opened 2020-09-16
Add support for Twig 3

jorrit:
## Subject

It seems Twig 3 support can be enabled without other changes.

I am targeting this branch, because using Twig 3 is backwards compatible.

## Changelog

```markdown
### Added
- Added support for Twig 3.
```

TODO:

* [ ] Update nelmio/api-doc-bundle to ^3.0 (perhaps wait until #1801 is merged).
* [ ] Check if changes in the Twig/Node directory work in practice.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,7 @@
         "symfony/twig-bundle": "^4.4",
         "symfony/validator": "^4.4 || ^5.1",
         "twig/string-extra": "^3.0",
-        "twig/twig": "^2.12.1"
+        "twig/twig": "^2.12.1 || ^3.0"
     },
     "conflict": {
         "doctrine/dbal": "<2.6.0",
--- a/src/Twig/Extension/MediaExtension.php
+++ b/src/Twig/Extension/MediaExtension.php
@@ -21,12 +21,11 @@ use Sonata\MediaBundle\Twig\TokenParser\PathTokenParser;
 use Sonata\MediaBundle\Twig\TokenParser\ThumbnailTokenParser;
 use Twig\Environment;
 use Twig\Extension\AbstractExtension;
-use Twig\Extension\InitRuntimeInterface;
 
 /**
  * @final since sonata-project/media-bundle 3.21.0
  */
-class MediaExtension extends AbstractExtension implements InitRuntimeInterface
+class MediaExtension extends AbstractExtension
 {
     /**
      * @var Pool
@@ -43,11 +42,6 @@ class MediaExtension extends AbstractExtension implements InitRuntimeInterface
      */
     protected $mediaManager;
 
-    /**
-     * @var Environment
-     */
-    protected $environment;
-
     public function __construct(Pool $mediaService, ManagerInterface $mediaManager)
     {
         $this->mediaService = $mediaService;
@@ -63,11 +57,6 @@ class MediaExtension extends AbstractExtension implements InitRuntimeInterface
         ];
     }
 
-    public function initRuntime(Environment $environment)
-    {
-        $this->environment = $environment;
-    }
-
     /**
      * @param MediaInterface $media
      * @param string         $format
@@ -75,7 +64,7 @@ class MediaExtension extends AbstractExtension implements InitRuntimeInterface
      *
      * @return string
      */
-    public function media($media, $format, $options = [])
+    public function media(Environment $environment, $media, $format, $options = [])
     {
         $media = $this->getMedia($media);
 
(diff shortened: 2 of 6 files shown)


jorrit:
For some reason the tests fail locally because `\Sonata\Form\Validator\ErrorElement` is final and some tests in `FileProviderTest` try to mock it. I have no idea why it works in GitHub actions.

phansys:
`twig/twig:3` is not being installed: https://github.com/sonata-project/SonataMediaBundle/pull/1807/checks?check_run_id=1123657456#step:12:238.

franmomu:
By the way:
```bash
$ composer why-not twig/twig 3
sonata-project/formatter-bundle  4.2.0  requires  twig/twig (^2.12.1)
```

jorrit:
Sorry, what a rookie mistake. Tomorrow I'll see if the formatter bundle can be upgraded.

franmomu:
> Sorry, what a rookie mistake. Tomorrow I'll see if the formatter bundle can be upgraded.

It's not rookie at all, I didn't know about that command until I saw it to @greg0ire

jorrit:
FormatterBundle requires the MediaBundle so there is a circular reference of two bundles requiring twig 2. What's the best approach here? Update one, check the other and then go back and check the first again?

jordisala1991:
Do you know if the media-bundle has any dependency with formatter-bundle? I mean, does it have: `use Sonata\Formatter....;` or on services.xml?

jorrit:
@jordisala1991: no, there is no code reference. In fact, it seems the only place the Formatter Bundle is referenced is to load an additional configuration file called `formatter.xml`, which loads a Twig extension. This Twig extension is needed by the Formatter Bundle but has no dependency on it.

I have removed the dev dependency and Twig 3 is installed now. I have fixed a breakage, but it seems `nelmio/api-doc-bundle` has to be updated too. This is quite some work: https://github.com/nelmio/NelmioApiDocBundle/blob/master/UPGRADE-3.0.md

composer.json
```diff
@@ -99,7 +98,8 @@
         "sonata-project/seo-bundle": "^2.1",
         "symfony/monolog-bundle": "If you want to log exceptions produced when dealing with images.",
         "tilleuls/ckeditor-sonata-media-bundle": "^1.0",
-        "twig/extra-bundle": "Auto configures the Twig Intl extension"
+        "twig/extra-bundle": "Auto configures the Twig Intl extension",
+        "sonata-project/formatter-bundle": "^3.4 || ^4.0"
```
core23, line 102: Please add a textual description, no version information.

src/Twig/Extension/MediaExtension.php
```diff
@@ -21,12 +21,11 @@
 use Sonata\MediaBundle\Twig\TokenParser\ThumbnailTokenParser;
 use Twig\Environment;
 use Twig\Extension\AbstractExtension;
-use Twig\Extension\InitRuntimeInterface;
 
 /**
  * @final since sonata-project/media-bundle 3.21.0
  */
-class MediaExtension extends AbstractExtension implements InitRuntimeInterface
+class MediaExtension extends AbstractExtension
```
core23, line 28: BC Break alert

core23 (changes requested):

composer.json
jorrit: I left that in by mistake. I think it can be removed altogether. MediaBundle just provides some services to the Formatter Bundle, not the other way around.

src/Twig/Extension/MediaExtension.php
jorrit: Does this matter? I don't think anyone (should) rely on MediaExtension implements InitRuntimeInterface or not.

phansys: >Does this matter? I don't think anyone (should) rely on MediaExtension implements InitRuntimeInterface or not.

Think about the result if a user is declaring a method (argument or return) with the removed interface. Does this matter?

jorrit: `InitRuntimeInterface` is part of Twig 2 and removed in Twig 3. If the user refers to `InitRuntimeInterface` they can't upgrade to Twig 3.

phansys: AFAIK, we aren't dropping the support for `twig/twig:2` in this PR.

jorrit: That's true. The functionality of InitRuntimeInterface can be implemented diffently in a way that works both in Twig 2 and 3. InitRuntimeInterface is only used to inject the `Twig\Environment` and I've moved that to the classes in `Twig\Node`.

jordisala1991:
This PR should target master branch if it contains BC breaks.

jorrit:
> This PR should target master branch if it contains BC breaks.

Perhaps I can do something with `interface_exists()`? Anyway, #1801 needs to be finished first.

PR #1808, opened 2020-09-16
Remove unicode escape in json example

jorrit:
## Subject

The docs build fails, it seems the same issue as reported here: https://github.com/sphinx-doc/sphinx/issues/8210

Let's just remove the unicode escape because it doesn't matter.

I am targeting this branch, because it's a docs change.


jordisala1991 (approved):

jordisala1991:
Thank you @jorrit

Issue #1769, opened 2020-07-21
S3 filesystem should not have require on access key and secret key


Webonaute:
If you deploy on EC2 instance, you might already have permission to access internaly the s3 bucket. Maintaining access key and secret key is not require in that case.

### Environment
SonataMedia with s3 as filesystem

## Expected results
Not providing AWS_KEY and AWS_SECRET should still work on instance having proper IAM permission.

## Actual results

forced to create and maintain AWS_KEY and AWS_SECRET environment variable.
if you dont set accessKey and secretKey, it trigger error.

Replacing `src/DependencyInjection/Configuration.php`
```
 ->scalarNode('accessKey')->isRequired()->end()
 ->scalarNode('secretKey')->isRequired()->end()
```
By: 
```
 ->scalarNode('accessKey')->end()
 ->scalarNode('secretKey')->end()
```
might do the trick

phansys:
I think it's possible to make this change if we warn the user with a useful message in cases where the credentials are not defined or the connection isn't successful.
Do you mind to create a PR @Webonaute?

PR #1817, opened 2020-09-22
Add option to set additional S3 configuration settings

jorrit:
## Subject

I am using S3 compatible file storage (MinIO) in my environment. This requires that the URLs to it have the path style: `http://endpoint/bucket/filepath`. This is supported by the S3 SDK if the `use_path_style_endpoint` parameter is set.

Currently, setting additional options to the S3Client is not supported. This PR adds a new configuration node that lets the user set additional configuration values to be passed to the S3 client.

I am targeting this branch, because it's backwards compatible.

## Changelog

```markdown
### Added
- Added option to set additional S3 configuration settings
```

### DIFF
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -111,7 +111,9 @@ Full configuration options:
                 encryption:     aes256 # can be aes256 or not set
                 cache_control:  max-age=86400 # or any other
                 meta:
-                    key1:       value1 #any amount of metas(sent as x-amz-meta-key1 = value1)
+                    key1:       value1 # any amount of metas(sent as x-amz-meta-key1 = value1)
+                config:
+                    key1:       value1 # any amount of additional configuration settings to pass to the AWS SDK
 
             mogilefs:
                 hosts:      []
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -310,6 +310,12 @@ class Configuration implements ConfigurationInterface
                                     ->prototype('scalar')
                                     ->end()
                                 ->end()
+                                ->arrayNode('config')
+                                    ->info('Additional configuration settings to pass to the AWS S3 SDK')
+                                    ->useAttributeAsKey('name')
+                                    ->prototype('scalar')
+                                    ->end()
+                                ->end()
                             ->end()
                         ->end()
 
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -475,6 +475,10 @@ class SonataMediaExtension extends Extension implements PrependExtensionInterfac
                     ->setFactory([S3Client::class, 'factory']);
             }
 
+            if (!empty($config['filesystem']['s3']['config'])) {
+                $arguments = array_merge($arguments, $config['filesystem']['s3']['config']);
+            }
+
             $container->getDefinition('sonata.media.adapter.service.s3')
                 ->replaceArgument(0, $arguments);
         } else {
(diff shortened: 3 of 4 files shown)


jorrit:
Is there anything I can do to help this issue further?

VincentLanglet (commented):
Hi @jorrit. I think nobody saw the PR. I ping @sonata-project/contributors to review it.

When I read this, it seems that
```
s3:
    config:
        region: foo
```
overrides
```
s3:
    region: bar
```
Same for some others key.

IMHO, We should avoid having two way to do the same thing.

jorrit:
Do you have a suggestion? Perhaps I could read directly from the `s3:` level like this:

```yaml
s3:
  use_path_style_endpoint: true
  region: 'moon-1'
```

But then I would need to figure out how to combine validated keys with unvalidated keys.

VincentLanglet:
What are the key that you need ? Maybe we just need to add more validated keys.

jorrit:
The `S3Client` class accepts these variables in addition to the default variables, of which there are many more:

 * bucket_endpoint
 * use_arn_region
 * use_accelerate_endpoint
 * use_dual_stack_endpoint
 * use_path_style_endpoint

The goal of this PR was to allow the (power) user to set any of these parameters and not just serve my specific use case.

VincentLanglet:
> The `S3Client` class accepts these variables in addition to the default variables, of which there are many more:
> 
> * bucket_endpoint
> * use_arn_region
> * use_accelerate_endpoint
> * use_dual_stack_endpoint
> * use_path_style_endpoint
> 
> The goal of this PR was to allow the (power) user to set any of these parameters and not just serve my specific use case.

What about a PR to allow the config
```
s3:
     bucket:
     accessKey:
     secretKey:
     create:         false
     region:         s3.amazonaws.com # change if not using US Standard region
     version:        2006-03-01 # change according the API version you are using
     storage:        standard # can be one of: standard or reduced
     acl:            public # can be one of: public, private, open, auth_read, owner_read, owner_full_control
     encryption:     aes256 # can be aes256 or not set
     cache_control:  max-age=86400 # or any other
     bucket_endpoint: NEW
     use_arn_region: NEW
     use_accelerate_endpoint: NEW
     use_dual_stack_endpoint: NEW
     use_path_style_endpoint: NEW
     meta:
         value1         
```
?

phansys:
This PR is related to #1769.
Regarding the options, please note that the bucket endpoint can be already configured through the `endpoint` node:
https://github.com/sonata-project/SonataMediaBundle/blob/c91ba02f1f9277987052f488b97d457689f6582c/src/DependencyInjection/Configuration.php#L273

jorrit:
> This PR is related to #1769.
> Regarding the options, please note that the bucket endpoint can be already configured through the `endpoint` node:
> https://github.com/sonata-project/SonataMediaBundle/blob/c91ba02f1f9277987052f488b97d457689f6582c/src/DependencyInjection/Configuration.php#L273

The AWS SDK will rewrite the host of the endpoint from `http://endpoint` to `http://bucketname.endpoint` unless `use_path_style_endpoint` is true. That is what got me here.

> What about a PR to allow the config
 
There are many more config options that are universal for all AWS SDK clients that I didn't list. I wished to avoid copying the list of possible options from the SDK to our Configuration.php.

I could read the possible options directly using `S3Client::getArguments()`. This would be quite ingenious I think, but would make Configuration.php dynamic depending on whether the SDK is installed.

VincentLanglet:
> > What about a PR to allow the config
> 
> There are many more config options that are universal for all AWS SDK clients that I didn't list. I wished to avoid copying the list of possible options from the SDK to our Configuration.php.

Sure I understand that. But that would allow to validate all the config value (for instance, depending on the SDK version), and provide specific error message.

Since we started to enumerate some config options (like region) we may keep doing it the same way...

But maybe other @sonata-project/contributors have a different opinion on this.

greg0ire:
What about asking the user for the id of a service that is an instance of `S3Client`? That way they can configure it however they see fit, with the most appropriate bundle for that (`aws/aws-sdk-php-symfony`)

VincentLanglet:
Hi @jorrit ; are you still interested by this ?

jorrit:
No, I've solved this issue in a different way.

PR #1821, opened 2020-09-23
Make `sonata.media.admin.media.manager` an alias of `sonata.admin.manager.orm` / `sonata.admin.manager.doctrine_mongodb` 

jorrit:
## Subject

Since sonata-project/doctrine-orm-admin-bundle 3.22 the constructor of `Sonata\DoctrineORMAdminBundle\Model\ModelManager` throws a deprecation warning when the second argument is not provided. The file `doctrine_orm_admin.xml` of the media bundle registers this class as the service `sonata.media.admin.media.manager` without this argument. I think this service can be made an alias of `sonata.admin.manager.orm`. This makes sure the parameters are correct regardless of the orm bundle version.

Maybe I've overlooked something that would make this solution invalid, but it seems this solution also makes sure there is just one instance of the `ModelManager`, which seems a plus.

I am targeting this branch, because it is a backwards compatible change.

## Changelog

```markdown
### Changed
- When using Doctrine ORM or MongoDB the service `sonata.media.admin.media.manager` is now an alias of `sonata.admin.manager.orm` or `sonata.admin.manager.doctrine_mongodb` instead of a separate service implemented by the same class.

### Fixed
- Deprecation notice in `Sonata\DoctrineORMAdminBundle\Model\ModelManager` and `Sonata\DoctrineMongoDBAdminBundle\Model\ModelManager`.
```

### DIFF
--- a/src/Resources/config/doctrine_mongodb_admin.xml
+++ b/src/Resources/config/doctrine_mongodb_admin.xml
@@ -40,9 +40,7 @@
                 </argument>
             </call>
         </service>
-        <service id="sonata.media.admin.media.manager" class="Sonata\DoctrineMongoDBAdminBundle\Model\ModelManager" public="true">
-            <argument type="service" id="doctrine_mongodb"/>
-        </service>
+        <service id="sonata.media.admin.media.manager" alias="sonata.admin.manager.doctrine_mongodb" public="true"/>
         <service id="sonata.media.admin.gallery" class="%sonata.media.admin.gallery.class%" public="true">
             <tag name="sonata.admin" manager_type="doctrine_mongodb" group="%sonata.media.admin.groupname%" label_catalogue="%sonata.media.admin.gallery.translation_domain%" label="gallery" label_translator_strategy="sonata.admin.label.strategy.underscore" icon="%sonata.media.admin.groupicon%"/>
             <argument/>
--- a/src/Resources/config/doctrine_orm_admin.xml
+++ b/src/Resources/config/doctrine_orm_admin.xml
@@ -41,9 +41,7 @@
                 </argument>
             </call>
         </service>
-        <service id="sonata.media.admin.media.manager" class="Sonata\DoctrineORMAdminBundle\Model\ModelManager" public="true">
-            <argument type="service" id="doctrine"/>
-        </service>
+        <service id="sonata.media.admin.media.manager" alias="sonata.admin.manager.orm" public="true"/>
         <service id="sonata.media.admin.gallery" class="%sonata.media.admin.gallery.class%" public="true">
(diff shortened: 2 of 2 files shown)


franmomu (dismissed):
Seems like a fine solution 👍

phansys:
I think this is a good proposal, I'd just like to have a clear idea about how both services are currently registered, in order to avoid any potential edge case where the aliased service could not be found.
Could you please provide a little explanation @jorrit?

jorrit:
The service `sonata.admin.manager.orm` is created unconditionally by the SonataDoctrineORMAdmin bundle (see https://github.com/sonata-project/SonataDoctrineORMAdminBundle/blob/3.x/src/Resources/config/doctrine_orm.xml).
It is set to the same class as `sonata.media.admin.media.manager` but depending on the version of the ORM bundle it will have the right amount of arguments.

phansys:
This service was unconditionally available before this change. Now, it aliases a service which might not exist if `doctrine/orm` is not installed.
IMO, if we decide to make this change, we should take the same action for all the available managers.

jorrit:
The class these two services are currently referring to is `Sonata\DoctrineORMAdminBundle\Model\ModelManager` so right now these can only exist when SonataDoctrineORMAdminBundle is installed. Nothing will change there.

Furthermore, sonata_orm_admin.xml will only be loaded when the `sonata_media.db_driver` setting is set to `doctrine_orm`.

phansys:
I'm referring to these sibling services:

The `phpcr` implementation seems to be following the same alias approach:
https://github.com/sonata-project/SonataMediaBundle/blob/f18c04b7c9cd09a526ab8f37ee2726dff81ffdeb/src/Resources/config/doctrine_phpcr_admin.xml#L21

But not `mongodb`:
https://github.com/sonata-project/SonataMediaBundle/blob/f18c04b7c9cd09a526ab8f37ee2726dff81ffdeb/src/Resources/config/doctrine_mongodb_admin.xml#L43

jorrit:
I hadn't looked at the other persistence bundles. It's interesting that PHPCR already uses an alias. This confirms that it works. I changed mongodb too, which seems to have the same deprecation situation (see https://github.com/sonata-project/SonataDoctrineMongoDBAdminBundle/commit/b382c816171e6891f055cbcbe7e0a21f6c49aa28)

franmomu (approved):

phansys (approved):
Excellent!
I think we could remove the "sonata.media.admin.media.manager" definition if its aliased service doesn't exist (which is the case for custom managers).

jorrit:
@phansys What situation would that be? Wouldn't `sonata.admin.manager.orm` point to the custom manager also?

phansys:
> @phansys What situation would that be? Wouldn't `sonata.admin.manager.orm` point to the custom manager also?

We are currently supporting 3 built-in managers: `orm`, `doctrine_mongodb` and `doctrine_phpcr`. I was thinking that also custom managers were allowed and inferred from the `manager_type` option, but this is not the case:
https://github.com/sonata-project/SonataMediaBundle/blob/f18c04b7c9cd09a526ab8f37ee2726dff81ffdeb/src/DependencyInjection/Configuration.php#L45-L52
So, forget about my last comment.

phansys:
Thank you @jorrit!

PR #1832, opened 2020-10-06
Improve orm XML mapping configuration

jorrit:
## Subject

The *.orm.xml files contained some errors:

 * The xsi attributes were not setup properly, preventing validation in an XML editor.
 * The default values for three columns were configured using the `default` attribute. This attribute does not exist. A default value must be configured using the `<options>` subelement.

I have removed the default values altogether and not replaced them with the right configuration code, because this introduces schema changes. Apparently the bundle works fine without the defaults.

I am targeting this branch, because the changes are backwards compatible.

## Changelog

```markdown
### Fixed
- Fixed validity of *.orm.xml mapping files.
```


phansys (approved):

greg0ire (approved):

greg0ire:
Thanks @jorrit !

PR #1478, opened 2018-10-03
Removed `BaseProxyExtension` dependency

core23:
I am targeting this branch, because this is a patch.

Closes sonata-project/SonataFormatterBundle#367

## Changelog

```markdown
### Fixed
- Restore compatible for `sonata-project/formatter-bundle` 4

```

## Subject

Moved all stuff from `Sonata\FormatterBundle\Extension\BaseProxyExtension` to this bundle.

PR #1834, opened 2020-10-07
Allow symfony/validator and symfony/finder ^5.1

jorrit:
## Subject

I wish to use symfony/validator ^5.1 in my project and Sonatas Media and Formatter bundle are preventing this. I have removed the development dependency on the formatter bundle in this change because otherwise it's impossible to update shared dependencies.

I also updated symfony/finder to ^5.1.

I am targeting this branch, because it is a backwards compatible change.

## Changelog

```markdown
### Added
- Support for symfony/validator ^5.1
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "symfony/dependency-injection": "^4.4",
         "symfony/event-dispatcher": "^4.4",
         "symfony/filesystem": "^4.4 || ^5.1",
-        "symfony/finder": "^4.4",
+        "symfony/finder": "^4.4 || ^5.1",
         "symfony/form": "^4.4",
         "symfony/framework-bundle": "^4.4",
         "symfony/http-foundation": "^4.4",
@@ -55,7 +55,7 @@
         "symfony/translation": "^4.4",
         "symfony/twig-bridge": "^4.4",
         "symfony/twig-bundle": "^4.4",
-        "symfony/validator": "^4.4",
+        "symfony/validator": "^4.4 || ^5.1",
         "twig/string-extra": "^3.0",
         "twig/twig": "^2.12.1"
     },


VincentLanglet:
How can we remove `sonata-project/formatter-bundle` without any consequences ? Is there other useless dependencies ?

jordisala1991:
Well the dependency exists: https://github.com/sonata-project/SonataMediaBundle/blob/53576fda7887ced1e5268f9f39e7f86aad3bb35e/src/DependencyInjection/SonataMediaExtension.php#L75

It would be nice to break the circular dependency at some point, but that would require at least to configure a variant for testing

jordisala1991:
It is interesting... the only class defined on the formatter.xml is a class that we removed the dependency on formatterBundle a long ago: https://github.com/sonata-project/SonataMediaBundle/pull/1478

Maybe we should check if somewhere else the bundle is used and if not just remove the if and let the class load without conditions? wdyt @VincentLanglet

VincentLanglet:
> Maybe we should check if somewhere else the bundle is used and if not just remove the if and let the class load without conditions? wdyt @VincentLanglet

I agree.
But then we should move the declaration to this file https://github.com/sonata-project/SonataMediaBundle/blob/53576fda7887ced1e5268f9f39e7f86aad3bb35e/src/Resources/config/twig.xml instead

jorrit:
> But then we should move the declaration to this file [`53576fd`/src/Resources/config/twig.xml](https://github.com/sonata-project/SonataMediaBundle/blob/53576fda7887ced1e5268f9f39e7f86aad3bb35e/src/Resources/config/twig.xml) instead

See #1836.

VincentLanglet (approved):

jordisala1991 (approved):

jordisala1991:
Thank you @jorrit

PR #1664, opened 2020-01-03
Deprecate Guzzle with psr http-client

core23:
I am targetting this branch, because this adds a new implementation without removing the existing one. This is a rework of https://github.com/sonata-project/SonataMediaBundle/pull/1181, but targets the psr http-client.

## Changelog

``` markdown
### Added
- Added support for `psr/http-client` in `BaseVideoProvider`

### Deprecated
 - Deprecate `Guzzle` and `Buzz` usage in `BaseVideoProvider`
```
## Subject

Added support for an abstract http client without removing the existing old guzzle implementation.

PR #1854, opened 2020-11-04
Amazon mimetype bugfix

michealmouner:
## Subject

this happened on 3.28 and should be reverted (the fix will be BC).

Fix Amazon guess MimeType on amazon s3 
actually the `$filename` at this point is `amazon s3 path` and not the `tmp path` for the file 
So, any mime guesser will fail because file not found on the machine

related to [#1664] Deprecate Guzzle and Buzz usage in BaseVideoProvider (@core23)

## Changelog

```markdown
### Changed
- update AmazonMetadataBuilder::getContentType 

```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,7 @@
         "jms/serializer-bundle": "^2.3.1 || ^3.0",
         "knplabs/gaufrette": "^0.8 || ^0.9",
         "kriswallsmith/buzz": "^0.15 || ^0.16",
+        "league/mime-type-detection": "^1.5",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,6 @@
         "jms/serializer-bundle": "^2.3.1 || ^3.0",
         "knplabs/gaufrette": "^0.8 || ^0.9",
         "kriswallsmith/buzz": "^0.15 || ^0.16",
-        "league/mime-type-detection": "^1.5",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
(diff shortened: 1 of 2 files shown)


phansys:
AFAIK, the previous behavior was intended to guess the MIME type based on the file contents, not just inferring it from the file extension.
Please, note that even if there is no choice to guess the MIME type from the file contents, the current implementation also provides the extension based inference through the `getMimeTypes()` method.

michealmouner:
as u said we don't have local file anymore at this step
and also, we checked the content type in earlier steps using  SymfonyMimeType (when the file was exist locally)
So, guessing the type from extension at this step will not be problem

note: `league Flysystem bundle` is using the same idea to guess the type when uploading to S3

phansys:
Then, I don't understand why we need other dependency to perform a task that is [already covered](https://symfony.com/doc/current/components/mime.html#mime-types-utilities).

michealmouner:
> Then, I don't understand why we need other dependency to perform a task that is [already covered](https://symfony.com/doc/current/components/mime.html#mime-types-utilities).

Done

jorrit:
Can I help with this issue? Because I am experiencing this bug also.

core23 (approved):

phansys (commented):
Thank you for your first contribution @michealmouner!
Please, be aware of the following concerns:
* The changelog must be updated, because with the current changes, there is no new dependency;
* The changelog entries should not be written in first person (by instance, instead of "I added" use "Added");
* A test covering this fix would be great in order to avoid regressions.

michealmouner:
> Can I help with this issue? Because I am experiencing this bug also.

sure 👍

phansys:
@michealmouner, @jorrit; do you have any news about this?

jorrit:
My improvements are at #1866.

PR #1857, opened 2020-11-09
Add missing italian translations

gremo:

## Changelog

```markdown
### Changed
- Translation keys for `gallery` and `Gallery`, inspired by the DE translation file, 
  for correct translations of menu key and the form group.
```

### DIFF
--- a/src/Resources/translations/SonataMediaBundle.it.xliff
+++ b/src/Resources/translations/SonataMediaBundle.it.xliff
@@ -11,8 +11,12 @@
                 <target>Media</target>
             </trans-unit>
             <trans-unit id="gallery">
-                <source>Gallery</source>
+                <source>gallery</source>
                 <target>Gallerie</target>
+            </trans-unit>
+            <trans-unit id="Gallery">
+                <source>Gallery</source>
+                <target>Galleria</target>
             </trans-unit>
             <trans-unit id="options">
                 <source>Options</source>


src/Resources/translations/SonataMediaBundle.it.xliff
```diff
@@ -11,9 +11,13 @@
                 <target>Media</target>
             </trans-unit>
             <trans-unit id="gallery">
-                <source>Gallery</source>
+                <source>gallery</source>
                 <target>Gallerie</target>
             </trans-unit>
+            <trans-unit id="Gallery">
+                <source>Gallery</source>
```
phansys, line 18:
```suggestion
```

phansys (changes requested):

src/Resources/translations/SonataMediaBundle.it.xliff
gremo: Hi @phansys , this is on purpose. We still need both `gallery` and `Gallery` keys, because how the translation are arranged. See for example here: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Resources/translations/SonataMediaBundle.de.xliff#L570

phansys: Why should we have different translation keys between languages? Even worst, when this key isn't present in the catalog for the [base language](https://github.com/sonata-project/SonataMediaBundle/blob/ec6bff7f8f38ee79ec65c5f9113ba0514088e251/src/Resources/translations/SonataMediaBundle.en.xliff#L13-L16).
IMO, if there is a missing key, we should add it for all the available languages.

gremo: I agree with you @phansys but this is an "old" problem. Even the naming of form groups are quite wrong, compared to other sonata bundles.

I can add the missing key for the base language and my language, but as you can see there is a "case mismatch" which require a more complex solution (form groups renaming), a lot of translations to change without knowing the language itself and a breaking change.

VincentLanglet: Hi @gremo ! Could you link us the place these translations (gallery and Gallery) are used ?

There is the same thing with 
```
<trans-unit id="media_">
    <source>Media</source>
    <target>Medien</target>
</trans-unit>
```

Either it's an error with translation (it should always be `Gallery` or always `gallery`) and then we remove the useless one and update the other, either it's a missing translation and we should add it
- at least for the base language
- with always the same id (you used `Gallery` but in DE `gallery_` is used)

gremo: Hi @VincentLanglet, sure!

[Here is the admin label (lower cased `gallery`)](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Resources/config/doctrine_orm_admin.xml#L46) while here is the [form group label (upper cased `Gallery`)](https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Admin/GalleryAdmin.php#L88).

In my PR i preferred to use `Gallery` as identifier because `gallery_` (with underscore at the end) isn't so meaningful. 

However, I agree with you, we should stick and conform all languages. What do you suggest?

VincentLanglet: If the quick fix is preferred:
- Since all id are lowercase, I would go with `gallery_`.
- I would add the `gallery_` key just next to `gallery` one. (So it means moving the DE ones)

But the best fix would be to change
```
$formMapper
            ->with('Gallery', ['class' => 'col-md-9'])->end()
            ->with('Options', ['class' => 'col-md-3'])->end()
        ;
```
with a translation key
```
$formMapper
            ->with('form_group.gallery', ['class' => 'col-md-9'])->end()
            ->with('form_group.options', ['class' => 'col-md-3'])->end()
        ;
```
And updating the translations (By changing `Options` to `form_group.options` and adding/updating `form_group.gallery`)

Don't you think ?

gremo: @VincentLanglet OK for me. Yes, best fix would be renaming the group, but it's such a huge break... don't think it would be a good idea to do it in 3.x. If you are OK with underscored version, I'll go for it, also for the german language.

VincentLanglet: > @VincentLanglet OK for me. Yes, best fix would be renaming the group, but it's such a huge break... don't think it would be a good idea to do it in 3.x.

Why do you consider this as a break ? As long as we update the translations files to rename the key, it will show the same code as before, no ?

If it's BC-break, it could be great to add these comment
```
$formMapper
// NEXT_MAJOR: Change Gallery key to `form_group.gallery` and update translations files.
->with('Gallery', ['class' => 'col-md-9'])->end()
// NEXT_MAJOR: Change Options key to `form_group.options` and update translations files.
->with('Options', ['class' => 'col-md-3'])->end()
;
```
This way it will be done in next major.

> If you are OK with underscored version, I'll go for it, also for the german language.

And doing this for the base language (en) could be great

gremo: @VincentLanglet one could extend the `GalleryAdmin`, adding new fields to the Gallery or Options groups. Doing this, renaming the groups, will break other people work. I thinks it's not something that one would expect from a minor or a patch version, am I wrong?

I'll create a new PR because this is misleading now.

VincentLanglet: > @VincentLanglet one could extend the `GalleryAdmin`, adding new fields to the `Gallery` or `Options` groups. Doing this, renaming the groups, will break other people work, at least will break the admin page layout. I thinks it's not something that one would expect from a minor or a patch version, am I wrong?

Oh, you're right.

gremo:
Closed in favour of #1863

src/Resources/translations/SonataMediaBundle.it.xliff
gremo: Hi @VincentLanglet , done, check my new PR, thanks :)

PR #1859, opened 2020-11-13
Replace is_callable call with ReflectionClass

franmomu:
In PHP8, is_callable fails when checking for a non-static method with a classname. 

See https://github.com/php/php-src/blob/f5be0e5110ff816d2ce168c154cb3575d08e3cad/UPGRADING#L30

I haven't check other packages how have they solved this, maybe there is another way.

I guess the proper way to solve it (for next major version) is to create an interface with that method and check if the class implements that interface.

### DIFF
--- a/src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
+++ b/src/DependencyInjection/Compiler/ThumbnailCompilerPass.php
@@ -32,7 +32,9 @@ class ThumbnailCompilerPass implements CompilerPassInterface
             'sonata.media.thumbnail.format'
         );
 
-        if (!\is_callable([$container->getParameterBag()->resolveValue($definition->getClass()), 'addResizer'])) {
+        $reflectionClass = new \ReflectionClass($container->getParameterBag()->resolveValue($definition->getClass()));
+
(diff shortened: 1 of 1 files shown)


VincentLanglet:
The check is not really enough, since the signature of the method must support
```
$definition->addMethodCall('addResizer', [$id, new Reference($id)]);
```

It could be great to introduce an Interface indeed.

Do you want to add them here ? If not, could you at least add a NEXT_MAJOR comment ?

jordisala1991:
> The check is not really enough, since the signature of the method must support
> 
> ```
> $definition->addMethodCall('addResizer', [$id, new Reference($id)]);
> ```
> 
> It could be great to introduce an Interface indeed.
> 
> Do you want to add them here ? If not, could you at least add a comment/issue ?

We should add an interface, but it should not block this PR
https://github.com/sonata-project/SonataMediaBundle/issues/1861

jordisala1991 (approved):

OskarStark:
I am for adding an interface, but merging this PR first.

@franmomu open for a PR introducing an interface? At least an issue to keep that in mind?

OskarStark (approved):

OskarStark:
Nevermind, #1861 is already created b @jordisala1991 👍🏻

PR #1866, opened 2020-11-18
Fix guessing mime type in AmazonMetadataBuilder

jorrit:
## Subject

`AmazonMetadataBuilder->getContentType()` should not try to access the `$filename` parameter as a local file because it has been uploaded already to S3.

I am targeting this branch, because it's fixing a regression.

Closes #1854.

## Changelog

```markdown
### Fixed
- Guessing the content type of a file stored on Amazon S3.
```

### DIFF
--- a/src/Metadata/AmazonMetadataBuilder.php
+++ b/src/Metadata/AmazonMetadataBuilder.php
@@ -114,12 +114,16 @@ class AmazonMetadataBuilder implements MetadataBuilderInterface
     /**
      * Gets the correct content-type.
      *
-     * @param string $filename
+     * @param string $filename path to the file inside the S3 bucket
      *
      * @return array
+     * @phpstan-return array{contentType: string}
      */
     protected function getContentType($filename)
     {
-        return ['contentType' => $this->mimeTypes->guessMimeType($filename)];
+        $ext = pathinfo($filename, PATHINFO_EXTENSION);
+        $mimeTypes = $this->mimeTypes->getMimeTypes($ext);
+
+        return ['contentType' => current($mimeTypes)];
     }
 }
--- a/tests/Metadata/AmazonMetadataBuilderTest.php
+++ b/tests/Metadata/AmazonMetadataBuilderTest.php
@@ -26,8 +26,10 @@ final class AmazonMetadataBuilderTest extends TestCase
      */
     public function testAmazon(array $settings, array $expected): void
     {
-        $mimeTypes = $this->createMock(MimeTypesInterface::class);
-        $mimeTypes->method('guessMimeType')->willReturn($expected['contentType']);
+        $mimeTypes = $this->createStub(MimeTypesInterface::class);
+        $mimeTypes->method('getMimeTypes')->willReturnCallback(static function (string $ext): array {
+            return 'png' === $ext ? ['image/png'] : [];
+        });
 
         $media = $this->createStub(MediaInterface::class);
         $filename = '/test/folder/testfile.png';


src/Metadata/AmazonMetadataBuilder.php
```diff
@@ -114,12 +114,15 @@ protected function getDefaultMetadata()
     /**
      * Gets the correct content-type.
      *
-     * @param string $filename
+     * @param string $filename path to the file inside the s3 bucket
      *
      * @return array
```
core23, line 119: Bonuspoints if you can add more specific (phpstan) phpdoc here.

core23 (approved):

src/Metadata/AmazonMetadataBuilder.php
jorrit: I would love those bonus points, but what exactly do you mean? Replacing `@return array` with something more specific? Details about why $filename isn't present when this function is called?

VincentLanglet: I think he wants something like
```
@phpstan-return array<string, string>
```
```
@phpstan-return array{'contentType': string}
```

src/Metadata/AmazonMetadataBuilder.php
```diff
@@ -114,12 +114,15 @@ protected function getDefaultMetadata()
     /**
      * Gets the correct content-type.
      *
-     * @param string $filename
+     * @param string $filename path to the file inside the s3 bucket
```
phansys, line 117:
```suggestion
     * @param string $filename path to the file inside the S3 bucket
```

tests/Metadata/AmazonMetadataBuilderTest.php
```diff
@@ -27,7 +27,9 @@ final class AmazonMetadataBuilderTest extends TestCase
     public function testAmazon(array $settings, array $expected): void
     {
         $mimeTypes = $this->createMock(MimeTypesInterface::class);
-        $mimeTypes->method('guessMimeType')->willReturn($expected['contentType']);
+        $mimeTypes->method('getMimeTypes')->willReturnCallback(static function ($ext) {
```
phansys, line 30:
```suggestion
        $mimeTypes->method('getMimeTypes')->willReturnCallback(static function (string $ext): array {
```

phansys (changes requested):

tests/Metadata/AmazonMetadataBuilderTest.php
```diff
@@ -27,7 +27,9 @@ final class AmazonMetadataBuilderTest extends TestCase
     public function testAmazon(array $settings, array $expected): void
     {
         $mimeTypes = $this->createMock(MimeTypesInterface::class);
```
phansys, line 29:
```suggestion
        $mimeTypes = $this->createStub(MimeTypesInterface::class);
```

VincentLanglet (approved):

core23 (approved):

src/Metadata/AmazonMetadataBuilder.php
core23: You don't need the quotes around `contentType`, but it shoud work

VincentLanglet (approved):

core23 (approved):

michealmouner (approved):
thank you

phansys (approved):

VincentLanglet:
Thanks ! @jorrit

PR #1889, opened 2021-01-04
Datetimeinterface

### LINKED ISSUE #1710: DateTimes should be type-hinted as DateTimeInterface
## Feature Request

Currently, properties containing a date (or a datetime) are type-hintend in phpdoc with `DateTime`.
Since Doctrine also allows to map properties with `DateTimeImmutable`, such properties should be type-hinted with `DateTimeInterface`.

See for example https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Model/Media.php#L103

garak:
## Subject

As detailed in related issue, this PR replace type hints and return type of DateTime with broader DateTimeInterface, allowing to use DateTimeImmutable.

I am targeting this branch, because change is BC.

Closes #1710

## Changelog

Replaced DateTime type hints and return types with DateTimeInterface 

```markdown
### Changed
- Replaced DateTime type hints and return types with DateTimeInterface 
```

Added some checks in testing, to skip tests when not appliable.
Notice that this can be fixed in some other ways (e.g. by forcing some requires in dev). Anyway, I couldn't run my tests without such changes.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -51,6 +51,11 @@ are required when aws/aws-sdk-php 3.x is installed.
 
 The `Guzzle` and `Buzz` dependencies are deprecated and will be replaced with the abstract `http-client` interface, so you can choose your preferred client implementation. You should adapt to the new `BaseVideoProvider::__construct()` signature.
 
+## Model classes use `\DateTimeInterface`
+
+If you overrided some date-related methods (`setUpdatedAt`, `setCreatedAt`, `setCdnFlushAt`) in model classes (Gallery, Media, GalleryHasMedia),
+you need to change type hint of argument to `\DateTimeInterface`.
+
 UPGRADE FROM 3.25 to 3.26
 =========================
 
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -51,11 +51,6 @@ are required when aws/aws-sdk-php 3.x is installed.
 
 The `Guzzle` and `Buzz` dependencies are deprecated and will be replaced with the abstract `http-client` interface, so you can choose your preferred client implementation. You should adapt to the new `BaseVideoProvider::__construct()` signature.
 
-## Model classes use `\DateTimeInterface`
-
-If you overrided some date-related methods (`setUpdatedAt`, `setCreatedAt`, `setCdnFlushAt`) in model classes (Gallery, Media, GalleryHasMedia),
-you need to change type hint of argument to `\DateTimeInterface`.
-
 UPGRADE FROM 3.25 to 3.26
 =========================
 
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -49,6 +49,9 @@ If you have implemented a custom model, you must adapt the signature of the foll
  * `GalleryHasMediaInterface::getId`
  * `GalleryInterface::getId`
 
+If you overrided some date-related methods (`setUpdatedAt`, `setCreatedAt`, `setCdnFlushAt`) in model classes (Gallery, Media, GalleryHasMedia),
+you need to change type hint of argument to `\DateTimeInterface`.
+
 ## Renamed GalleryHasMedia to GalleryItem
 
 All Actions, Controllers, Interfaces and anything related to this is renamed accordingly.
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -49,8 +49,8 @@ If you have implemented a custom model, you must adapt the signature of the foll
  * `GalleryHasMediaInterface::getId`
  * `GalleryInterface::getId`
 
-If you overrided some date-related methods (`setUpdatedAt`, `setCreatedAt`, `setCdnFlushAt`) in model classes (Gallery, Media, GalleryHasMedia),
-you need to change type hint of argument to `\DateTimeInterface`.
+If you have overridden some date-related methods (`setUpdatedAt()`, `setCreatedAt()`, `setCdnFlushAt()`) in model classes (`Gallery`, `Media`, `GalleryHasMedia`),
+you need to change the arguments' type declarations to `\DateTimeInterface`.
 
 ## Renamed GalleryHasMedia to GalleryItem
 
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -35,12 +35,12 @@ abstract class Gallery implements GalleryInterface
     protected $enabled;
 
     /**
-     * @var \DateTime|null
+     * @var \DateTimeInterface|null
      */
     protected $updatedAt;
 
     /**
-     * @var \DateTime|null
+     * @var \DateTimeInterface|null
      */
     protected $createdAt;
 
@@ -79,7 +79,7 @@ abstract class Gallery implements GalleryInterface
         return $this->enabled;
     }
 
-    public function setUpdatedAt(?\DateTime $updatedAt = null): void
+    public function setUpdatedAt(?\DateTimeInterface $updatedAt = null): void
     {
         $this->updatedAt = $updatedAt;
     }
@@ -89,7 +89,7 @@ abstract class Gallery implements GalleryInterface
         return $this->updatedAt;
     }
 
-    public function setCreatedAt(?\DateTime $createdAt = null): void
+    public function setCreatedAt(?\DateTimeInterface $createdAt = null): void
     {
         $this->createdAt = $createdAt;
     }
--- a/tests/CDN/PantherPortalTest.php
+++ b/tests/CDN/PantherPortalTest.php
@@ -18,6 +18,9 @@ use Sonata\MediaBundle\CDN\PantherPortal;
 
 class PantherPortalTest extends TestCase
 {
+    /**
+     * @requires extension soap
+     */
     public function testPortal(): void
     {
         $client = $this->createMock(ClientSpy::class);
@@ -35,6 +38,9 @@ class PantherPortalTest extends TestCase
         $panther->flushPaths([$path]);
     }
 
+    /**
+     * @requires extension soap
+     */
     public function testException(): void
     {
         $this->expectException(\RuntimeException::class);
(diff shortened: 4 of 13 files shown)


VincentLanglet:
There is something wrong with your branch, it has a lot of commit

garak:
> There is something wrong with your branch, it has a lot of commit

I noticed that, and tried to rebase my branch on 3.x, but no luck.
Any suggestions?

VincentLanglet:
> Any suggestions?

```
git reset --hard origin/3.x
git cherry-pick 4df4fb9
```

tests/CDN/PantherPortalTest.php
```diff
@@ -20,6 +20,10 @@ class PantherPortalTest extends TestCase
 {
     public function testPortal(): void
```
OskarStark, line 21: Symfony uses @require annotation can we use this too? 

Also this should be fixed in a separate PR

tests/Form/Type/MediaTypeTest.php
```diff
@@ -67,6 +67,10 @@ public function testMissingFormOptions(): void
 
     public function testMissingFormContextOption(): void
     {
+        if (!\function_exists('gd_info')) {
```
OskarStark, line 70: Same here for the @require annotation

OskarStark (changes requested):

tests/CDN/PantherPortalTest.php
garak: Never heard about it, can you point me to an example?

OskarStark: Sure, I am currently on a phone 📱, will have a look later

src/Model/Gallery.php
```diff
@@ -82,7 +82,7 @@ public function getEnabled()
         return $this->enabled;
     }
 
-    public function setUpdatedAt(?\DateTime $updatedAt = null)
+    public function setUpdatedAt(?\DateTimeInterface $updatedAt = null)
```
OskarStark, line 85: This is a BC break isn't it? 🤔

garak: Why?

VincentLanglet: I think it is, but it feel so bad to delay this PR to a major version...
Maybe an upgrade note could be enough

VincentLanglet: @garak In case someone was overriding the function.

```
class A {
    public function foo(\DatetimeInterface $a) {}
}

class B extends A {
    public function foo(\Datetime $a) {}
}
```
is not valid

tests/CDN/PantherPortalTest.php
garak: never mind, found it (and added it)

OskarStark: Does it work?

tests/CDN/PantherPortalTest.php
```diff
@@ -18,6 +18,9 @@
 
 class PantherPortalTest extends TestCase
 {
+    /**
+     * @requires function \SoapClient::__construct
```
OskarStark, line 22:
```suggestion
     * @requires extension soap
```

tests/CDN/PantherPortalTest.php
```diff
@@ -35,6 +38,9 @@ public function testPortal(): void
         $panther->flushPaths([$path]);
     }
 
+    /**
+     * @requires function \SoapClient::__construct
```
OskarStark, line 42:
```suggestion
     * @requires extension soap
```

tests/CDN/PantherPortalTest.php
```diff
@@ -50,10 +56,12 @@ public function testException(): void
     }
 }
 
-class ClientSpy extends \SoapClient
```
OskarStark, line 53: I don't understand why we have a class, if we mock it anyway, can't we directly mock it or better create an anonymous class:
```php
        $client = new class() implements \SoapClient {
            public function flush(): string
            {
                return 'hello';
            }
        };
```

WDYT @VincentLanglet @franmomu ?

tests/CDN/PantherPortalTest.php
```diff
@@ -50,10 +56,12 @@ public function testException(): void
...
-{
-    public function flush(): string
+if (class_exists(\SoapClient::class)) {
+    class ClientSpy extends \SoapClient
```
OskarStark, line 60: Can we remove the if class exists check if we move it to an extra file? 🤔

tests/Form/Type/MediaTypeTest.php
```diff
@@ -123,6 +126,9 @@ public function testInvalidFormProviderOption(): void
         ]);
     }
 
+    /**
+     * @requires extension gd_info
```
OskarStark, line 130: IMO the extension is `gd`:
```suggestion
     * @requires extension gd
```

tests/Functional/Controller/TruncateControllerTest.php
```diff
@@ -21,6 +21,9 @@
 
 final class TruncateControllerTest extends TestCase
 {
+    /**
+     * @requires extension gd_info
```
OskarStark, line 25: IMO the extension is `gd`:
```suggestion
     * @requires extension gd
```

src/Model/Gallery.php
OskarStark: But we can remove the typehint now and trigger an exception like:

/**
 * NEXT_MAJOR: Add the typehint back.
 */
public function setUpdatedAt(/* ?\DateTimeInterface */$updatedAt = null)
{
    trigger deprecation here to warn the user it needs to be an DateTimeInterface?
    // ...
}

garak: @OskarStark but how can you check for argument? You can't check for interfaces

VincentLanglet (changes requested):
Can you add a line in https://github.com/sonata-project/SonataMediaBundle/blob/3.x/UPGRADE-3.x.md#upgrade-from-3x-to-3x to explain about the little BC-break ?

UPGRADE-3.x.md
```diff
@@ -43,6 +43,10 @@ are required when aws/aws-sdk-php 3.x is installed.
 
 The `Guzzle` and `Buzz` dependencies are deprecated and will be replaced with the abstract `http-client` interface, so you can choose your preferred client implementation. You should adapt to the new `BaseVideoProvider::__construct()` signature.
 
+## Model classes use `\DateTimeInterface`
+
+If you overrided `setUpdatedAt` method in model classes (Gallery, Media, GalleryHasMedia), you need to change type hint of argument to `\DateTimeInterface`.
```
VincentLanglet, line 48: There is also `setCdnFlushAt` and `setCreatedAt`

VincentLanglet (approved):

src/Model/Gallery.php
franmomu: What do you mean @garak? IMO we should not do this (as it is right now, introducing a BC break) in the stable branch, these models are meant to be extended by the user.

VincentLanglet:
```
class A {
    public function foo($a) {}
}

class B extends A {
    public function foo(\Datetime $a) {}
}
```
Is not more valid IMHO @OskarStark 

> IMO we should not do this (as it is right now, introducing a BC break) in the stable branch, these models are meant to be extended by the user.

Model are meant to be extended but these setters shouldn't be a lot

garak: @franmomu I think we can safely change since likely no one is overriding such methods.
I can't see a feasible way to check for argument, in the case we remove the type hint.

core23: This change will break my project(s)!

> Model are meant to be extended but these setters shouldn't be a lot

For a greenfield project, you're right, This model classes were designed for extensions, (e.g. the entity class will do this).

core23 (dismissed):

src/Model/Gallery.php
OskarStark: It is, but since 7.4: https://3v4l.org/iCa85

VincentLanglet: So should we do this PR on master ? :/

core23: Yes, that's the only way :(

VincentLanglet:
Can this PR be done on master ? :)

VincentLanglet:
@jordisala1991 recently worked on the version 4 of this bundle.
@garak Do you have time to make this PR on master ?

I've added the 4.0 milestone in order to not forget to take this BC-break.

garak:
> @garak Do you have time to make this PR on master ?

Done

OskarStark (approved):

VincentLanglet:
There is some conflicts

garak:
> There is some conflicts

I just rebased

UPGRADE-3.x.md
```diff
@@ -51,6 +51,11 @@ are required when aws/aws-sdk-php 3.x is installed.
 
 The `Guzzle` and `Buzz` dependencies are deprecated and will be replaced with the abstract `http-client` interface, so you can choose your preferred client implementation. You should adapt to the new `BaseVideoProvider::__construct()` signature.
 
+## Model classes use `\DateTimeInterface`
```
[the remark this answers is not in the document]
garak: Moved to UGRADE-4.0md

VincentLanglet (approved):

UPGRADE-4.0.md
```diff
@@ -49,6 +49,9 @@ If you have implemented a custom model, you must adapt the signature of the foll
  * `GalleryHasMediaInterface::getId`
  * `GalleryInterface::getId`
 
+If you overrided some date-related methods (`setUpdatedAt`, `setCreatedAt`, `setCdnFlushAt`) in model classes (Gallery, Media, GalleryHasMedia),
+you need to change type hint of argument to `\DateTimeInterface`.
```
phansys, line 53:
```suggestion
If you have overridden some date-related methods (`setUpdatedAt()`, `setCreatedAt()`, `setCdnFlushAt()`) in model classes (`Gallery`, `Media`, `GalleryHasMedia`),
you need to change the arguments' type declarations to `\DateTimeInterface`.
```

phansys (approved):

jordisala1991 (approved):

jordisala1991:
Thank you @garak

PR #1910, opened 2021-02-10
Add support for doctrine/persistence 2

VincentLanglet:
## Subject

Since https://github.com/sonata-project/SonataNotificationBundle/releases/tag/3.11.0, we can use doctrine/persistence 2 in this bundle ; that's why tests are failing. I fixed this.


greg0ire:
I don't think a changelog is required here: you fixed the test suite.

greg0ire (approved):

phansys (approved):

PR #1933, opened 2021-03-12
Bump aws/aws-sdk-php < 3.0

wbloszyk:
## Subject

I am targeting this branch, because this change respect BC.

## Changelog

```markdown
### Removed
- Removed support for `aws/aws-sdk-php` < 3.0
```


src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -420,23 +405,6 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
-            // @todo: Remove the following conditional block when support for aws/aws-sdk-php < 3.0 is dropped.
-            if (isset($config['filesystem']['s3']['sdk_version'])) {
-                if (3 === $config['filesystem']['s3']['sdk_version'] && !class_exists(Sdk::class)) {
-                    throw new \UnexpectedValueException(
-                        'The configuration "sonata_media.filesystem.s3.sdk_version" can not contain the value 3 since'.
-                        ' the installed version of aws/aws-sdk-php is not 3.x.'
-                    );
-                }
-
-                if (2 === $config['filesystem']['s3']['sdk_version'] && class_exists(Sdk::class)) {
```
phansys, line 432: IMO, you should throw an exception if the users pass `2` as value for this config, since it will be not allowed anymore.
In this way, I think a conflict against `aws/aws-sdk-php; < 3.0 || >=4.0` will be required.

phansys (changes requested):

src/DependencyInjection/SonataMediaExtension.php
wbloszyk: I move it from master. After, conflict MUST be updated. 

IMO `sdk_version` should be ignored. If i understand corretly it will be autodetected. Anyway I will check it again when I find some time for it.

phansys: We can't ignore a configuration, unless there is other configuration explicitly producing the ignoring rule.
IMO, allowing to set a configuration that can not be processed and omitting it silently without any advice brings a bad DX.

wbloszyk: `sdk_version` is validlidated in configuration. I add test for it. Also I update composer.

composer.json
```diff
@@ -63,6 +63,7 @@
         "twig/twig": "^2.12.1"
     },
     "conflict": {
+        "aws/aws-sdk-php": "<3.0",
```
phansys, line 66: We don't know what changes the 4.x release can introduce. When it is released, we can remove the conflict if there are no breaking changes.
```suggestion
        "aws/aws-sdk-php": "<3.0 || >= 4.0",
```

phansys (changes requested):

phansys (approved):

wbloszyk:
Can we merge it? @sonata-project/contributors

VincentLanglet (approved):

PR #1930, opened 2021-03-10
Update constraints in order to allow some "symfony/*:^5.2" deps

phansys:
## Subject
Update constraints in order to allow some "symfony/*:^5.2" deps.

I am targeting this branch, because these changes respect BC.

## Changelog

```markdown
### Added
- Support for "symfony/templating:^5.2";
- Support for "symfony/http-foundation:^5.2";
- Support for "symfony/dependency-injection:^5.2";
- Support for "symfony/config:^5.2";
- Support for "symfony/routing:^5.2".
```

The support for "sonata-project/datagrid-bundle:2.x" was dropped because it is not allowed by "sonata-project/notification-bundle":
```
composer why-not sonata-project/datagrid-bundle:2.5.0
sonata-project/notification-bundle  3.x-dev  requires  sonata-project/datagrid-bundle (^3.0)  
```

## To do

- [x] Check if we can allow more "symfony/*:^5.2" dependencies;
- [x] Use a stable constraint for "sonata-project/notification-bundle";
- [x] Resolve the exception `ArgumentCountError` from "nelmio/api-doc-bundle" (see #1801, https://github.com/nelmio/NelmioApiDocBundle/pull/1457#issuecomment-796620190).

PR #1943, opened 2021-03-24
Declare conflict against nelmio/api-doc-bundle <2.13.5 || >=4.0

phansys:
## Subject
Declare conflict against nelmio/api-doc-bundle <2.13.5 || >=4.0.

Follow up of #1930.

## Changelog

```markdown
### Added
- Conflict against nelmio/api-doc-bundle <2.13.5 || >=4.0.
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -70,6 +70,7 @@
         "friendsofsymfony/rest-bundle": "<2.6",
         "jms/serializer": "<0.13",
         "liip/imagine-bundle": "<1.9",
+        "nelmio/api-doc-bundle": "<2.13.5 || >=4.0",
         "rackspace/php-opencloud": "<1.6",
         "sonata-project/block-bundle": "<3.17 || >=4.0",
         "sonata-project/classification-bundle": "<3.0 || >= 5.0",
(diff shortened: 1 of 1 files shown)


jordisala1991 (approved):

wbloszyk (approved):

phansys:
Seems that the branch protection rules are referencing some non existing jobs.

wbloszyk:
> Seems that the branch protection rules are referencing some non existing jobs.

https://github.com/sonata-project/SonataMediaBundle/pull/1944/files

jordisala1991 (approved):

Webonaute:
Why declaring conflict instead of fixing the conflict? it prevent from upgrading our project.

VincentLanglet:
> Why declaring conflict instead of fixing the conflict? it prevent from upgrading our project.

Declaring the conflict is about telling the truth about the current situation to avoid people upgrading the nelmio api and getting stuck with a non working project.

Fixing the conflict
- does not take the same time than adding the conflict
- is maybe not possible without introducing BC-break.

Feel free to provide a PR ! :)

wbloszyk:
> Why declaring conflict instead of fixing the conflict? it prevent from upgrading our project.

Diffrence between `nelmio/api-doc-bundle` v2, v3 and v4 are huge. Also they are using annotations which force us to provide solution for each version in seperate files. It will be hard to load with dotrine and contribiuting. 

In Sonata4 we want to move API in seperate bundle(s). Then users will be ale to use Sonata solution, third part solution or PWN solution for provide API.

VincentLanglet:
Btw @phansys I'm not sure it works with the V3: https://github.com/sonata-project/SonataMediaBundle/pull/1958

PR #1955, opened 2021-04-19
Avoid to generate the same caller reference twice for a CloudFront invalidation batch request

phansys:
## Subject
Avoid to generate the same caller reference twice for a CloudFront invalidation batch request, since the CDN don't allow the submitted paths to be flushed if this reference was previously used.
See https://docs.aws.amazon.com/cloudfront/latest/APIReference/API_InvalidationBatch.html#cloudfront-Type-InvalidationBatch-CallerReference.

## Changelog

```markdown
### Fixed
- CDN invalidation from CloudFront when submitting paths that were previously invalidated.
```

### DIFF
--- a/src/CDN/CloudFront.php
+++ b/src/CDN/CloudFront.php
@@ -189,7 +189,7 @@ class CloudFront implements CDNInterface
                     'Quantity' => \count($normalizedPaths),
                     'Items' => $normalizedPaths,
                 ],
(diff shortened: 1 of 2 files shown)


VincentLanglet (approved):
Shouldn't we use something more random ? Like uniqid

n4huel (approved):

jordisala1991 (approved):

jordisala1991:
Thank you @phansys

PR #1967, opened 2021-06-06
Add phpstan and psalm

jordisala1991:
## Subject

I am targeting this branch, because this will be hard to apply and mantain on 3.x.


psalm-baseline.xml
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="4.7.3@38c452ae584467e939d55377aaf83b5a26f19dd1">
+  <file src="src/Controller/Api/GalleryController.php">
+    <MissingDependency occurrences="3">
```
VincentLanglet, line 4: Shouldn't we add the dependency ?

jordisala1991: We have the Rest bundle, but it is complaining because the rest bundle requires sensio framework extra bundle and it does not include it. I don't know how, but on notification bundle it works and doesn't complain with this error.

psalm-baseline.xml
```diff
@@ -0,0 +1,26 @@
...
+      <code>Rest\View|FormInterface</code>
+    </MissingDependency>
+  </file>
+  <file src="src/Controller/Api/MediaController.php">
+    <MissingDependency occurrences="2">
+      <code>Rest\View</code>
+      <code>Rest\View|FormInterface</code>
+    </MissingDependency>
+  </file>
+  <file src="src/Listener/ODM/MediaEventSubscriber.php">
```
jordisala1991, line 16: For this other error, should I install doctrien/mongodb-odm?

VincentLanglet: Why do we have `doctrine/mongodb-odm` if we're supposed to use `sonata-project/doctrine-orm-admin-bundle` with this bundle ?

Anyway, if it fix the error, I would add the dev dependency yes. It would check our conflict is the right one.

phpstan.neon.dist
```diff
@@ -0,0 +1,19 @@
+includes:
+    - phpstan-baseline.neon
+
+parameters:
+    level: 0
+    bootstrapFiles:
+        - vendor/bin/.phpunit/phpunit/vendor/autoload.php
```
VincentLanglet, line 7: With this, there is something to add to the composer.json

https://github.com/sonata-project/SonataAdminBundle/blob/master/composer.json#L117-L124

But maybe we should start doing: 
https://github.com/sonata-project/dev-kit/issues/1559

composer.json
```diff
@@ -91,6 +91,7 @@
         "matthiasnoback/symfony-config-test": "^4.2",
         "matthiasnoback/symfony-dependency-injection-test": "^4.0",
         "nelmio/api-doc-bundle": "^2.13.5",
+        "phpstan/phpstan": "^0.12.84",
```
VincentLanglet, line 94: We should add the phpstan extension installer and phpstan/phpstan-phpunit.

composer.json
```diff
@@ -100,7 +101,8 @@
         "symfony/browser-kit": "^4.4 || ^5.2",
         "symfony/http-client": "^4.4 || ^5.2",
         "symfony/phpunit-bridge": "^5.2",
-        "symfony/security-csrf": "^4.4 || ^5.2"
+        "symfony/security-csrf": "^4.4 || ^5.2",
+        "vimeo/psalm": "^4.7.2"
```
VincentLanglet, line 105: We should also add the plugin for phpunit.

psalm-baseline.xml
jordisala1991: This bundle integrates with phpcr, mongodb, and orm, but we are only adding the sonata-project orm bundle as require-dev. (And it is still missing doctrine/orm)

phpstan.neon.dist
```diff
@@ -0,0 +1,19 @@
...
+    paths:
+        - src
+        - tests
+    excludePaths:
+        - src/Controller/Api/*
```
VincentLanglet, line 12: It would be better to avoid adding those 3 paths to the exclude path.

phpstan.neon.dist
jordisala1991: Agree on the dev-kit issue, but I would like not to block this one until a decision is made. So for now I just added the composer thing

psalm-baseline.xml
```diff
@@ -0,0 +1,96 @@
...
+    <!-- NEXT_MAJOR: Remove the PHPCR issues -->
+    <file src="src/Admin/PHPCR/GalleryAdmin.php">
+        <InvalidExtendClass occurrences="1">
+            <code>BaseGalleryAdmin</code>
+        </InvalidExtendClass>
+        <MethodSignatureMismatch occurrences="1">
+            <code>protected function configureDatagridFilters(DatagridMapper $filter)</code>
+        </MethodSignatureMismatch>
+    </file>
+    <!-- NEXT_MAJOR: Remove this issue -->
```
VincentLanglet, line 12: What's the issue ? How can you be sure all those can be remove in the next major ?

jordisala1991: Because I started this work on master, but I saw dev-kit requires all branches to be compatible with psalm, so I had to downgrade.

composer.json
jordisala1991: With the current level, does it provide something ? Nothing fails without it, but I can add it, yes.

VincentLanglet: It adds extra rules.

phpstan.neon.dist
jordisala1991: Solved

psalm-baseline.xml
jordisala1991: The issue with those classes is that on those version of blockbundle and seo bundle the parent clases are marked with `@final`. On next seo-bundle / block-bundle this is solved because parent classes extend other clases that are not final.

composer.json
jordisala1991: Solved

composer.json
jordisala1991: Solved

psalm-baseline.xml
jordisala1991: Solved

psalm-baseline.xml
jordisala1991: Added

VincentLanglet (approved):

jordisala1991:
Dev-kit PR: https://github.com/sonata-project/dev-kit/pull/1560

franmomu (approved):

jordisala1991:
I will do a merge to master asap

PR #1968, opened 2021-06-06
Move xml config to php

### LINKED ISSUE #1941: Migrate XML to PHP configuration
## Feature Request

Migrate all configs under: https://github.com/sonata-project/SonataMediaBundle/tree/3.x/src/Resources/config

This tool can help: https://github.com/symplify/config-transformer

Ref: https://github.com/sonata-project/SonataMediaBundle/pull/1940#discussion_r597608037

jordisala1991:
## Subject

I am targeting this branch, because this is the less problematic branch to do this change, and we are close to next major.

Closes #1941.

## Changelog

```markdown
### Changed
- Move from XML configuration to PHP
```


VincentLanglet (dismissed):

src/Resources/config/seo_block.php
```diff
@@ -0,0 +1,50 @@
...
+    // Use "service" function for creating references to services when dropping support for Symfony 4.4
+    $services = $containerConfigurator->services();
+
+    $services->set('sonata.media.block.breadcrumb_view', GalleryViewBreadcrumbBlockService::class)
+        ->tag('sonata.breadcrumb')
+        ->tag('sonata.block', ['context' => 'breadcrumb'])
+        ->args([
+            new ReferenceConfigurator('twig'),
+            new ReferenceConfigurator('knp_menu.menu_provider'),
+            new ReferenceConfigurator('knp_menu.factory'),
```
jordisala1991, line 30: Here we have the dependency with knp-menu (only when using seo-bundle too).

VincentLanglet: Does this fix need to be Backport to 3.X ?

src/Resources/config/api_controllers.php
```diff
@@ -0,0 +1,41 @@
...
+
+    $services->set('sonata.media.controller.api.media', MediaController::class)
+        ->args([
+            new ReferenceConfigurator('sonata.media.manager.media'),
+            new ReferenceConfigurator('sonata.media.pool'),
+            new ReferenceConfigurator('form.factory'),
+        ]);
+
+    $services->alias(GalleryController::class, 'sonata.media.controller.api.gallery');
+    $services->alias(MediaController::class, 'sonata.media.controller.api.media');
```
franmomu, line 40: In the xml these aliases are public, if these aliases are just to make the controllers public, I think we can just make `sonata.media.controller.api.gallery` and `sonata.media.controller.api.media` public and do not use alias.

src/Resources/config/seo_block.php
jordisala1991: This is kind of a fix, because in the real world, people using sonata most likely will add this bundle already because of admin-bundle requiring it, but IMO it should be explicit on composer.json. So it can be backported to 3.x, but since it is not critical, I would prefer if someone else can do it, I would like to focus on 4.0

src/Resources/config/api_controllers.php
jordisala1991: Those alias are used in the routing, I made them public again. Agree on the removing of this aliases an using directly the service, but I would prefer to do such change on another PR and focus here on the migration of xml to php with same definitions.

src/Resources/config/api_form_doctrine_mongodb.php
```diff
@@ -0,0 +1,64 @@
...
+            'sonata_api_write',
+        ]);
+
+    $services->set('sonata.media.api.form.type.media', ApiMediaType::class)
+        ->tag('form.type', ['alias' => 'sonata_media_api_form_media'])
+        ->args([
+            new ReferenceConfigurator('sonata.media.pool'),
+            '%sonata.media.admin.media.entity%',
+        ])
+        // NEXT_MAJOR: make symfony/monolog-bundle a require dependency and remove nullOnInvalid
```
jordisala1991, line 42: This is a comment that were on previous files. To be solved on next PRs.

VincentLanglet (approved):

franmomu (approved):

VincentLanglet:
Thanks

PR #1972, opened 2021-06-06
Improve phpstan to level 1

jordisala1991:
## Subject

I am targeting this branch, because master will be released soon and it is easier.


src/Model/Gallery.php
```diff
@@ -50,7 +50,8 @@ abstract class Gallery implements GalleryInterface
     protected $defaultFormat = MediaProviderInterface::FORMAT_REFERENCE;
 
     /**
-     * @var GalleryItemInterface[]|Collection
+     * @var Collection
+     * @phpstan-var Collection<GalleryItemInterface>
```
VincentLanglet, line 54:
```suggestion
     * @phpstan-var Collection<array-key, GalleryItemInterface>
```

And @phpstan-var is not really needed, phpstorm understand this. `Foo<bar>`

src/Model/GalleryInterface.php
```diff
@@ -100,12 +102,14 @@ public function setDefaultFormat($defaultFormat);
     public function getDefaultFormat();
 
     /**
-     * @param array $galleryItems
+     * @param Collection $galleryItems
+     * @phpstan-param Collection<GalleryItemInterface> $galleryItems
```
VincentLanglet, line 106: same

src/Model/GalleryInterface.php
```diff
@@ -100,12 +102,14 @@ public function setDefaultFormat($defaultFormat);
...
      */
     public function setGalleryItems($galleryItems);
 
     /**
-     * @return GalleryItemInterface[]
+     * @return Collection
+     * @phpstan-return Collection<GalleryItemInterface>
```
VincentLanglet, line 112: same

tests/Provider/BaseProviderTest.php
```diff
@@ -16,19 +16,16 @@
...
-use Sonata\MediaBundle\Provider\MediaProviderInterface;
+use Sonata\MediaBundle\Tests\App\Provider\TestProvider;
 use Sonata\MediaBundle\Tests\Entity\Media;
 use Sonata\MediaBundle\Thumbnail\ThumbnailInterface;
-use Symfony\Component\Form\FormBuilderInterface;
 
 class BaseProviderTest extends AbstractProviderTest
 {
-    public function getProvider(): MediaProviderInterface
+    public function getProvider(): TestProvider
```
jordisala1991, line 28: How show I deal with this? this method is abstract and returns an interface, but the concrete classes have more methods and phpstan complains, but if I do this psalm complains on higher levels, and php 7.3 complains too.

VincentLanglet: Which method do you need more than the one in MediaProviderInterface ?

Quick solution would be to keep the `MediaProviderInterface` return type and add `@return TestProvider`, it solve the php 7.3 error. But psalm will complain on higher level I think.

Right solution would be to make the `AbstractProviderTest` using a template
`@phpstan-template P of MediaProviderInterface` and
```
/** @phpstan-return P */
public function getProvider(): MediaProviderInterface;
```

jordisala1991: I will go for the template solution, seems to be the most accurate

jordisala1991: Done

src/Model/GalleryInterface.php
jordisala1991: Fixed

src/Model/Gallery.php
jordisala1991: Fixed

src/Model/GalleryInterface.php
jordisala1991: Fixed

VincentLanglet (approved):

phansys (approved):

VincentLanglet:
Thanks

PR #1973, opened 2021-06-07
Compatibility with Symfony 5 and Twig 3

jordisala1991:
## Subject

I am targeting this branch, because this contains bc breaks.

Closes #1927 #1897 #1932

## Changelog

```markdown
### Added
- Added compatibility with Symfony 5
- Added compatibility with Twig 3
```

PR #1975, opened 2021-06-09
Increase Psalm and PHPStan levels

jordisala1991:
## Subject

I am targeting this branch, because this causes a lot of BC breaks.

## Changelog

```markdown
### Removed
- Injecting directly the container on almost all cases
```


VincentLanglet:
Not sure it's wanted but you took all the work from https://github.com/sonata-project/SonataMediaBundle/pull/1973 is it still a draft ?

jordisala1991:
Yes, it is still a draft, requires the other PR merged first.

jordisala1991:
@VincentLanglet since you are doing a lot of work increasing phpstan on the admin bundle, I have a few questions on how to solve issues here.

src/Command/SyncThumbsCommand.php
```diff
@@ -160,6 +161,7 @@ public function execute(InputInterface $input, OutputInterface $output): int
             }
 
             //clear entity manager for saving memory
+            \assert($this->mediaManager instanceof BaseManager);
```
jordisala1991, line 164: This is correct or better directly type BaseManager instead of ManagerInterface (the manager interface does not have the getObjectManager method

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -401,10 +401,10 @@ private function configureAdapters(ContainerBuilder $container, array $config):
     {
         foreach (['gd', 'imagick', 'gmagick'] as $adapter) {
             if ($container->hasParameter('sonata.media.adapter.image.'.$adapter.'.class')) {
-                $container->register(
-                    'sonata.media.adapter.image.'.$adapter,
-                    $container->getParameter('sonata.media.adapter.image.'.$adapter.'.class')
-                );
+                $adapterClass = $container->getParameter('sonata.media.adapter.image.'.$adapter.'.class');
+                \assert(\is_string($adapterClass));
```
jordisala1991, line 405: better to cast this as string or assert is_string?

src/Listener/ORM/MediaEventSubscriber.php
```diff
@@ -44,22 +45,24 @@ public function onClear(): void
...
     {
-        $media = $args->getEntity();
+        $media = $args->getObject();
 
         if (!$media instanceof MediaInterface) {
-            return $media;
+            throw new \RuntimeException('There is no media on the persistence event.');
         }
 
         if ($this->container->has('sonata.media.manager.category') && !$media->getCategory()) {
```
jordisala1991, line 68: How could I fix this error? this method is on the Media model, but not on the MediaInterface (this was done when splitting media-bundle and classification-bundle to avoid having to install classification when only using media.

src/Twig/Extension/MediaExtension.php
```diff
@@ -61,9 +61,9 @@ public function getTokenParsers()
     }
 
     /**
-     * @param MediaInterface $media
-     * @param string         $format
-     * @param array          $options
+     * @param MediaInterface|int|string $media
```
jordisala1991, line 64: this could receive the MediaInterface directly or an identifier, it is correct to type it with int|string?

jordisala1991:
And this one: https://github.com/sonata-project/SonataMediaBundle/pull/1975/files#file-src-block-galleryblockservice-php-L121 will we problematic. I don't know how it is suposed to work.

VincentLanglet:
> And this one: https://github.com/sonata-project/SonataMediaBundle/pull/1975/files#file-src-block-galleryblockservice-php-L121 will we problematic. I don't know how it is suposed to work.

This method was deprecated in 3.x and remove in 4.0.

There is now a FieldDescriptionFactory::create method.

jordisala1991:
> > And this one: https://github.com/sonata-project/SonataMediaBundle/pull/1975/files#file-src-block-galleryblockservice-php-L121 will we problematic. I don't know how it is suposed to work.
> 
> This method was deprecated in 3.x and remove in 4.0.
> 
> There is now a FieldDescriptionFactory::create method.

Nice, then maybe the problem will be with the line 126:

```
src/Block/MediaBlockService.php#L216
Call to an undefined method Sonata\BlockBundle\Form\Mapper\FormMapper::getAdmin().
```

src/Listener/ORM/MediaEventSubscriber.php
VincentLanglet: You could have a check `method_exists` ; better would be a check `instanceOf SomeInterface`.

This split seems weird because I see `use Sonata\ClassificationBundle\Model\CategoryInterface;` in the `class Media`.
If I click on this, I end in the test/Fixtures...

src/DependencyInjection/SonataMediaExtension.php
VincentLanglet: In SonataAdmin I used assert, since IMHO it's always a string according to the config.

src/Command/SyncThumbsCommand.php
VincentLanglet: This is not correct.

Since the constructor use MediaManagerInterface, I could implement MediaManagerInterface without extending the BaseManager. This is a problem I had with multiple class from the AdminBundle:
An abstract class which implement an interface should not have more public method than the interface.

I solve this in SonataAdmin by deprecating a loooot of method in abstract classes, or adding some in the interface.

Here it worst than my case because it's in another bundle, so I would 
- Add an issue to the doctrine-extensions. In next-major either the method should not be public anymore, either it should be added to interface.
- Change the type in the construct to BaseManager or stop using this method. If you cannot stop using it means we have to add it to the interface.

Another way, BC, would be to create a new interface in doctrine-extensions which extends the ManagerInterface and add this method. The BaseManager will implements this new interface and you'll be able to use it in the constructor.

VincentLanglet: Looking at the way you use this method, I would say it would be nice to not add getObjectManager to the interface.

Maybe you could add some `clear` method instead.
Or a ClearableManagerInterface could be created and here you would do
```
if ($this->mediaManager instanceof ClearableManagerInterface) {
     $this->mediaManager->clear();
}
```
This way it would work for both clearable and not clearable managers.

It would be interested to see if some others sonata projects use directly the `getObjectManager` method.

src/Twig/Extension/MediaExtension.php
VincentLanglet: `MediaInterface|int|string` Seems correct to me

src/Command/SyncThumbsCommand.php
```diff
@@ -161,8 +161,9 @@ public function execute(InputInterface $input, OutputInterface $output): int
             }
 
             //clear entity manager for saving memory
-            \assert($this->mediaManager instanceof BaseManager);
-            $this->mediaManager->getObjectManager()->clear();
+            if ($this->mediaManager instanceof ClearableManagerInterface) {
```
VincentLanglet, line 164: I assume you plan to make the PR on doctrine-extensions. Ping me when done ;)

src/Block/MediaBlockService.php
```diff
@@ -209,10 +204,10 @@ protected function getFormatChoices(?MediaInterface $media = null): array
...
         // simulate an association ...
-        $fieldDescription = $this->getMediaAdmin()->getModelManager()->getNewFieldDescriptionInstance($this->mediaAdmin->getClass(), 'media', [
+        $fieldDescription = $this->fieldDescriptionFactory->create($this->mediaAdmin->getClass(), 'media', [
             'translation_domain' => 'SonataMediaBundle',
         ]);
-        $fieldDescription->setAssociationAdmin($this->getMediaAdmin());
+        $fieldDescription->setAssociationAdmin($this->mediaAdmin);
         $fieldDescription->setAdmin($form->getAdmin());
         $fieldDescription->setOption('edit', 'list');
         $fieldDescription->setAssociationMapping([
```
VincentLanglet, line 213: This method was removed. This is now in the construct of the fieldDescription.
And for instance the DoctrineORMAdmin fieldDescriptionFactory->create method is doing
```
[$metadata, $propertyName, $parentAssociationMappings] = $this->getParentMetadataForProperty($class, $name);

        return new FieldDescription(
            $name,
            $options,
            $metadata->fieldMappings[$propertyName] ?? [],
            $metadata->associationMappings[$propertyName] ?? [],
            $parentAssociationMappings,
            $propertyName
        );
```

So I would say this call can be removed, and it's the fieldDescriptionFactory to correctly set the association mapping.

src/Block/MediaBlockService.php
```diff
@@ -209,10 +204,10 @@ protected function getFormatChoices(?MediaInterface $media = null): array
     protected function getMediaBuilder(FormMapper $form): FormBuilderInterface
     {
         // simulate an association ...
-        $fieldDescription = $this->getMediaAdmin()->getModelManager()->getNewFieldDescriptionInstance($this->mediaAdmin->getClass(), 'media', [
+        $fieldDescription = $this->fieldDescriptionFactory->create($this->mediaAdmin->getClass(), 'media', [
             'translation_domain' => 'SonataMediaBundle',
         ]);
-        $fieldDescription->setAssociationAdmin($this->getMediaAdmin());
+        $fieldDescription->setAssociationAdmin($this->mediaAdmin);
         $fieldDescription->setAdmin($form->getAdmin());
```
VincentLanglet, line 211: The AdminBundle FormMapper has this method, but not the blockBundle one...

What I don't understand is that I don't find class which implements the FormMapper interface of BlockBundle.

src/Block/MediaBlockService.php
jordisala1991: Removed

src/Command/SyncThumbsCommand.php
jordisala1991: Yes!

src/Command/SyncThumbsCommand.php
jordisala1991: Done

src/Resources/config/block.php
```diff
@@ -27,34 +27,42 @@
...
     $services->set('sonata.media.block.feature_media', FeatureMediaBlockService::class)
         ->public()
         ->tag('sonata.block')
         ->args([
             new ReferenceConfigurator('twig'),
-            new ReferenceConfigurator('service_container'),
+            new ReferenceConfigurator('sonata.media.pool'),
+            new ReferenceConfigurator('sonata.media.admin.media'),
             new ReferenceConfigurator('sonata.media.manager.media'),
+            new ReferenceConfigurator('sonata.admin.field_description_factory.orm'),
```
jordisala1991, line 44: Ideally I would love to not couple with orm here, how should I do it? if Odm is installed I should use the field_description_factory from odm.

One idea could be to add an alias here to point to the correct field_description_factory
Other idea could be to name them equally on the persistence bundle (not sure if that is a good idea at all).

Maybe there are other ideas, wdyt @VincentLanglet ?

VincentLanglet: I don't know a lot with config. Maybe @franmomu does.
I don't think using the same name is the best cause you could use both persistence bundle.
I think alias is made for this. The same way you inject an interface and not a class implementing the interface in the construct.

src/Block/MediaBlockService.php
```diff
@@ -209,21 +203,17 @@ protected function getFormatChoices(?MediaInterface $media = null): array
     protected function getMediaBuilder(FormMapper $form): FormBuilderInterface
     {
         // simulate an association ...
-        $fieldDescription = $this->getMediaAdmin()->getModelManager()->getNewFieldDescriptionInstance($this->mediaAdmin->getClass(), 'media', [
+        $fieldDescription = $this->fieldDescriptionFactory->create($this->mediaAdmin->getClass(), 'media', [
```
franmomu, line 206:
```suggestion
        $fieldDescription = $this->mediaAdmin->createFieldDescription('media', [
```
There is a `AdminInterface::createFieldDescription()` method, so there is no need of injecting `FieldDescriptionFactory`.

jordisala1991: oh that's better

jordisala1991: This methods also calls the setAdmin, so I can remove that line

src/Block/MediaBlockService.php
jordisala1991: I think this wouldn't be a problem using the admin createFieldDescription

src/Resources/config/block.php
jordisala1991: This is not a problem now.

src/Block/MediaBlockService.php
jordisala1991: Done.

jordisala1991:
Not sure what to do with this:

```
 ------ -------------------------------------------------------------------- 
  Line   src/Filesystem/Replicate.php                                        
 ------ -------------------------------------------------------------------- 
  196    Call to an undefined method Gaufrette\Adapter::createFile().        
  201    Call to an undefined method Gaufrette\Adapter::createFileStream().  
  206    Call to an undefined method Gaufrette\Adapter::listDirectory().     
 ------ -------------------------------------------------------------------- 
```

Maybe remove those methods? I don't know a lot about Gaufrette, we have passed for several 0.x releases and maybe it was something from the past, or we have something that is broken.

jordisala1991:
Once the other 2 Prs are merged, I will prepare this one

VincentLanglet:
> Not sure what to do with this:
> 
> ```
>  ------ -------------------------------------------------------------------- 
>   Line   src/Filesystem/Replicate.php                                        
>  ------ -------------------------------------------------------------------- 
>   196    Call to an undefined method Gaufrette\Adapter::createFile().        
>   201    Call to an undefined method Gaufrette\Adapter::createFileStream().  
>   206    Call to an undefined method Gaufrette\Adapter::listDirectory().     
>  ------ -------------------------------------------------------------------- 
> ```
> 
> Maybe remove those methods? I don't know a lot about Gaufrette, we have passed for several 0.x releases and maybe it was something from the past, or we have something that is broken.

This was removed here: https://github.com/KnpLabs/Gaufrette/pull/98/files#diff-2a51af5b23ee4563c0778fc99b99112ce0bda19c2f6b0777ea1ffb90049abce5

Let's remove them

jordisala1991:
> > Not sure what to do with this:
> > ```
> >  ------ -------------------------------------------------------------------- 
> >   Line   src/Filesystem/Replicate.php                                        
> >  ------ -------------------------------------------------------------------- 
> >   196    Call to an undefined method Gaufrette\Adapter::createFile().        
> >   201    Call to an undefined method Gaufrette\Adapter::createFileStream().  
> >   206    Call to an undefined method Gaufrette\Adapter::listDirectory().     
> >  ------ -------------------------------------------------------------------- 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe remove those methods? I don't know a lot about Gaufrette, we have passed for several 0.x releases and maybe it was something from the past, or we have something that is broken.
> 
> This was removed here: https://github.com/KnpLabs/Gaufrette/pull/98/files#diff-2a51af5b23ee4563c0778fc99b99112ce0bda19c2f6b0777ea1ffb90049abce5
> 
> Let's remove them

Thanks for finding the PR, I was having trouble finding it. Based on that I have removed the listDirectories, but the other ones are keep with a new implementation, wdyt? I think this is how it is suposed to work.

src/Listener/ORM/MediaEventSubscriber.php
```diff
@@ -79,7 +79,7 @@ protected function getMedia(LifecycleEventArgs $args)
             throw new \RuntimeException('There is no media on the persistence event.');
         }
 
-        if (null !== $this->categoryManager && !$media->getCategory()) {
+        if (null !== $this->categoryManager && method_exists($media, 'getCategory') && method_exists($media, 'setCategory') && !$media->getCategory()) {
```
VincentLanglet, line 82: Should we create an issue about this ?

jordisala1991: Yes!

jordisala1991: Done!

src/Command/SyncThumbsCommand.php
jordisala1991: PR done, needs to be merged and released. Resolving this conversation.

jordisala1991:
This is almost ready, only missing the PR on doctrine-extensions.

src/Listener/ORM/MediaEventSubscriber.php
jordisala1991: I just opened an issue for this, adding the method_exists in the meanwhile.

composer.json
```diff
@@ -36,7 +36,7 @@
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
         "psr/log": "^1.0",
-        "sonata-project/doctrine-extensions": "^1.10.1",
+        "sonata-project/doctrine-extensions": "^1.13@dev",
```
VincentLanglet, line 39: You can now use 1.13

src/Admin/BaseMediaAdmin.php
```diff
@@ -118,7 +122,13 @@ public function alterNewInstance(object $object): void
...
                 $category = $this->categoryManager->find($categoryId);
 
-                if ($category && $category->getContext()->getId() === $context) {
+                if (!$category) {
+                    return;
+                }
+
+                $categoryContext = $category->getContext();
+
+                if ($categoryContext && method_exists($categoryContext, 'getId') && $categoryContext->getId() === $context) {
```
VincentLanglet, line 131: Can't https://github.com/sonata-project/SonataMediaBundle/issues/1978#issuecomment-861252460 could solve all the method_exists checks ?

jordisala1991: It depends on how Classification has implemented their interfaces, at some point we will have to decide what to do here, but there are bundles that contains the getId methods on the interface and other not.

composer.json
jordisala1991: Done

src/Admin/BaseMediaAdmin.php
jordisala1991: Solved

jordisala1991:
I have added the missing files to the Fixtures folder, replacing the old models and adding entities.
Now we are testing with latest classification bundle files without having to include it.

I have also prepared a new PR to integrate with the real classification-bundle.

The order should be:

1. Review this one taking into account that the Fixtures/ClassificationBundle will be removed soon.
2. Merge this one
3. Merge the PR on classification-bundle (this can come before or after, doesn't matter).
4. Prepare the removal of Fixtures/ClassificationBundle on: https://github.com/sonata-project/SonataMediaBundle/pull/1989

src/Model/MediaInterface.php
```diff
@@ -380,4 +381,8 @@ public function getGalleryItems();
      * @return string
      */
     public function getPreviousProviderReference();
+
+    public function getCategory(): ?CategoryInterface;
```
VincentLanglet, line 385: Should we have another Interface `MediaWithCategoryInterface` with these two methods instead ?

If you don't use the ClassificationnBundle you need MediaInterface, if you use the ClassificationBundle, you'll need the new Interface.

jordisala1991: I was thinking the same, but you will normally extend the src/Entity/BaseMedia.php, so you will always get the one with get/set Category.

We can add that interface, but the src/Entity will implement both.

VincentLanglet: I would have said that if someone wants the media without the classification, he could have implement the interface so 2 interfaces seems better than 1.

But...
Whoah, this code https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Model/Media.php#L17 mean that the Media entity cannot be used without the ClassificationBundle, isn't it ?
I don't understand how the "split" could work if the class is used.

If we use the Classification classes in the "root code" of the MediaBundle, I don't see other solution than adding the ClassificationBundle in the require section (and not only the require-dev). Or I totally misunderstand how this bundle works.

jordisala1991: This bundle works without the ClassificationBundle (I am using it on my projects).

The thing is: If you don't install classification bundle, the relation with media and Category is not created: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/DependencyInjection/SonataMediaExtension.php#L627-L638

Then, you just don't use the setCategory and getCategory, and that's all. The fields are not displayed on the admin pages.

VincentLanglet (approved):

src/Controller/Api/GalleryController.php
```diff
@@ -216,7 +217,10 @@ public function getGalleryMediasAction($id)
 
         $media = [];
         foreach ($galleryItems as $galleryItem) {
-            $media[] = $galleryItem->getMedia();
+            $itemMedia = $galleryItem->getMedia();
```
phansys, line 220: Maybe we could use `$medium` instead of `$itemMedia` here.

src/Extra/Pixlr.php
```diff
@@ -77,15 +83,24 @@ final class Pixlr
      * @param string $referrer
```
phansys, line 83: Could not we declare these types and remove the docblocks?

src/Filesystem/Replicate.php
```diff
@@ -193,17 +195,24 @@ public function getAdapterClassNames()
 
     public function createFile($key, Filesystem $filesystem)
     {
-        return $this->primary->createFile($key, $filesystem);
-    }
+        if ($this->primary instanceof FileFactory) {
+            return $this->primary->createFile($key, $filesystem);
+        } elseif ($this->secondary instanceof FileFactory) {
+            return $this->secondary->createFile($key, $filesystem);
+        }
```
phansys, line 202:
```suggestion
        }
        
        if ($this->secondary instanceof FileFactory) {
            return $this->secondary->createFile($key, $filesystem);
        }
```
```suggestion
        } elseif ($this->secondary instanceof FileFactory) {
            return $this->secondary->createFile($key, $filesystem);
        }
```

src/Filesystem/Replicate.php
```diff
@@ -193,17 +195,24 @@ public function getAdapterClassNames()
...
 
-    public function listDirectory($directory = '')
+    public function createStream($key)
     {
-        return $this->primary->listDirectory($directory);
+        if ($this->primary instanceof StreamFactory) {
+            return $this->primary->createStream($key);
+        } elseif ($this->secondary instanceof StreamFactory) {
+            return $this->secondary->createStream($key);
+        }
```
phansys, line 213:
```suggestion
        }
        
        if ($this->secondary instanceof StreamFactory) {
            return $this->secondary->createStream($key);
        }
```
```suggestion
        } elseif ($this->secondary instanceof StreamFactory) {
            return $this->secondary->createStream($key);
        }
```

src/Listener/ODM/MediaEventSubscriber.php
```diff
@@ -31,18 +33,26 @@ public function getSubscribedEvents()
...
     }
 
-    protected function getMedia(EventArgs $args)
+    protected function getMedia(LifecycleEventArgs $args)
     {
-        return $args->getDocument();
+        $media = $args->getObject();
+
+        if (!$media instanceof MediaInterface) {
+            throw new \RuntimeException('There is no media on the persistence event.');
```
phansys, line 53: I'd say `\InvalidArgumentException` or `\LogicException` are more accurate in this case.

src/Command/CleanMediaCommand.php
```diff
@@ -32,7 +31,7 @@ final class CleanMediaCommand extends Command
     protected static $defaultDescription = 'Find orphaned files in media upload directory';
 
     /**
-     * @var MediaProviderInterface[]|null
+     * @var string[]|null
```
phansys, line 34: Are these strings service ids or FQCNs? Just wondering if we could use `class-string`.

phansys (changes requested):

src/Extra/Pixlr.php
jordisala1991: Yes, I plan to do that on another Pr , when phpstan level is strong enough to first tell if the phpdoc is correct.

src/Command/CleanMediaCommand.php
jordisala1991: They are strings, `$provider->getName()`

src/Listener/ODM/MediaEventSubscriber.php
jordisala1991: Fixed

src/Controller/Api/GalleryController.php
jordisala1991: Fixed

jordisala1991:
@franmomu @VincentLanglet the dev-master of sonata-admin:

```
3) Sonata\MediaBundle\Tests\Controller\MediaAdminControllerTest::testCreateActionToSelectProvider
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: You have requested a non-existent service "sonata.admin.request.fetcher".
```

I think it is because of the last PR merged on master which changed the CRUDController

VincentLanglet:
This is a new services subscribed. You have to add it to the container in the test ; like I did for classification bundle

jordisala1991:
Right!, Fixed

VincentLanglet (dismissed):

src/Filesystem/Replicate.php
```diff
@@ -193,17 +198,28 @@ public function getAdapterClassNames()
...
 
-    public function createFileStream($key, Filesystem $filesystem)
-    {
-        return $this->primary->createFileStream($key, $filesystem);
+        if ($this->secondary instanceof FileFactory) {
+            return $this->secondary->createFile($key, $filesystem);
+        }
+
+        throw new \RuntimeException(sprintf('None of the adapters implements: %s', FileFactory::class));
```
phansys, line 209:
```suggestion
        throw new \LogicException(sprintf('None of the adapters implements %s.', FileFactory::class));
```
Even, I'd say that this exception could be of type `\TypeError`, since the invalid flow is caused by a value not respecting the expected type.

src/Filesystem/Replicate.php
```diff
@@ -193,17 +198,28 @@ public function getAdapterClassNames()
...
-        return $this->primary->listDirectory($directory);
+        if ($this->primary instanceof StreamFactory) {
+            return $this->primary->createStream($key);
+        }
+
+        if ($this->secondary instanceof StreamFactory) {
+            return $this->secondary->createStream($key);
+        }
+
+        throw new \RuntimeException(sprintf('None of the adapters implements: %s', StreamFactory::class));
```
phansys, line 222:
```suggestion
        throw new \LogicException(sprintf('None of the adapters implements %s.', StreamFactory::class));
```

phansys (commented):
I left some comments about the exception types.
It's not a blocker to me, just sharing some concerns.

phansys (approved):

PR #1979, opened 2021-06-13
Avoid defining services on the Extension class

jordisala1991:
## Subject

## Changelog

```markdown
### Changed
- Change to use integer for simple and square resizers mode
### Removed
- Remove usage of `.class` parameters on Container
- Remove defining services on Extension class
- Remove `ImagineCompatibleResizerTrait`
```


VincentLanglet (commented):
An upgrade note is needed isnt it ?

jordisala1991:
Yes, we should add an upgrade note. Should I mention the class parameter removal, the string to int change and the removal of the trait?

VincentLanglet:
> Yes, we should add an upgrade note. Should I mention the class parameter removal, the string to int change and the removal of the trait?

I don't know for the class and the trait, I had the string to int change, because it requires the user to update his config.

VincentLanglet (approved):

PR #1983, opened 2021-06-13
Improve SonataMedia Adapters

jordisala1991:
## Subject

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,20 @@
+parameters:
+    ignoreErrors:
+        -
+            # TODO: Remove this integration, the last version of this dependency is too old to be installed
+            message: "#^Class OpenCloud\\\\ObjectSource\\\\Service not found\\.$#"
+            count: 1
+            path: src/Resources/config/gaufrette.php
+
+        -
+            # TODO: Remove this integration, the last version of this dependency is too old to be installed
+            message: "#^Class OpenCloud\\\\OpenStack not found\\.$#"
+            count: 1
+            path: src/Resources/config/gaufrette.php
+
+        -
+            # TODO: Remove this integration, the last version of this dependency is too old to be installed
+            message: "#^Class OpenCloud\\\\Rackspace not found\\.$#"
+            count: 1
+            path: src/Resources/config/gaufrette.php
+
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -291,11 +291,11 @@ final class SonataMediaExtension extends Extension implements PrependExtensionIn
 
             $container->getDefinition('sonata.media.metadata.amazon')
                 ->replaceArgument(0, [
-                        'acl' => $config['filesystem']['s3']['acl'],
-                        'storage' => $config['filesystem']['s3']['storage'],
-                        'encryption' => $config['filesystem']['s3']['encryption'],
-                        'meta' => $config['filesystem']['s3']['meta'],
-                        'cache_control' => $config['filesystem']['s3']['cache_control'],
+                    'acl' => $config['filesystem']['s3']['acl'],
+                    'storage' => $config['filesystem']['s3']['storage'],
+                    'encryption' => $config['filesystem']['s3']['encryption'],
+                    'meta' => $config['filesystem']['s3']['meta'],
+                    'cache_control' => $config['filesystem']['s3']['cache_control'],
                 ]);
 
             $arguments = [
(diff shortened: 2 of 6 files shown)


src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -291,11 +291,11 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
 
             $container->getDefinition('sonata.media.metadata.amazon')
                 ->replaceArgument(0, [
-                        'acl' => $config['filesystem']['s3']['acl'],
-                        'storage' => $config['filesystem']['s3']['storage'],
-                        'encryption' => $config['filesystem']['s3']['encryption'],
-                        'meta' => $config['filesystem']['s3']['meta'],
-                        'cache_control' => $config['filesystem']['s3']['cache_control'],
+                    'acl' => $config['filesystem']['s3']['acl'],
```
VincentLanglet, line 294: Why is this not caught by the linter ? Is there a rule for this ?

phpstan-baseline.neon
```diff
@@ -0,0 +1,20 @@
+parameters:
+    ignoreErrors:
+        -
+            # TODO: Remove this integration, the last version of this dependency is too old to be installed
```
VincentLanglet, line 4: Should we create an issue about this ?

src/Resources/config/gaufrette.php
```diff
@@ -16,6 +16,9 @@
 use Gaufrette\Adapter\Ftp;
 use Gaufrette\Adapter\OpenCloud;
 use Gaufrette\Filesystem;
+use OpenCloud\ObjectSource\Service;
+use OpenCloud\OpenStack;
+use OpenCloud\Rackspace;
```
jordisala1991, line 21: This integration should probably be removed: https://github.com/rackspace/php-opencloud/blob/working/composer.json#L24-L29

Last release 2016... and I don't think it works... it forces the install of https://packagist.org/packages/guzzle/guzzle#v3.8.1

the 3.8.1 , and not the 3.9.1 because on 3.8.1 the symfony/event-dispatcher was >=2.1. On 3.9 they moved to: symfony/event-dispatcher: ~2.1

At the end, it probably does not work. I would prefer leaving this as is, and open another PR to remove this integration. IMO if we decide to not remove it, we shouldn't add it to composer.json, even in require-dev this could cause a lot of trouble, it is a really outdated code and dependencies (guzzle/guzzle is deprecated too)

Probably all the gaufrette could be removed in favour of flysystem, but that will be a lot of work probably... not going to land for 4.0.

jordisala1991: wdyt? @VincentLanglet

phpstan-baseline.neon
jordisala1991: You got too fast, read my other comment.

src/Resources/config/gaufrette.php
jordisala1991: More info, this is the dependency: https://github.com/rackspace/php-opencloud

And it is deprecated too, in favour of this one: https://github.com/php-opencloud/openstack

A little bit better but the gauffrete adapter does not work with that one, only with the other.

VincentLanglet: https://github.com/php-opencloud/openstack is recommended in the readme of https://github.com/rackspace/php-opencloud.
So https://github.com/rackspace/php-opencloud can be removed indeed.

src/DependencyInjection/SonataMediaExtension.php
jordisala1991: Not sure, I think this is not the first time I see something like this in Sonata, copy pastes bad indented.

src/Resources/config/gaufrette.php
jordisala1991: So, IMO:

1. This PR fixes an issue with having parameters for just class names and also not showing what are our dependencies because adding classes as strings instead of ::class, so it should be merged.
2. Open issue to talk about gauffrete (maybe for 5.0)
3. Open issue to remove OpenStack
4. Deprecated OpenStack on 3.x
5. Remove openStack on master

is that right?

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -330,35 +330,29 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
             $container->removeDefinition('sonata.media.filesystem.replicate');
         }
 
-        if ($container->hasDefinition('sonata.media.adapter.filesystem.mogilefs') && isset($config['filesystem']['mogilefs'])) {
```
jordisala1991, line 333: This was deprecated and remove on previous PRs.

VincentLanglet (approved):

phansys (approved):

PR #1984, opened 2021-06-13
Remove parameters from sonata admins

jordisala1991:
## Subject

## Changelog

```markdown
### Removed
- All container parameters that were used to customize classes.
```

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -3,11 +3,10 @@ UPGRADE FROM 3.x to 4.0
 
 ### Container parameters
 
-  We have been using container parameters on SonataMediaBundle 3.x as an undocumented extension
-  points for things like classes, and configurations.
+  We have been using container parameters as undocumented extension points for things like classes and configurations.
 
   In SonataMediaBundle 4.0 those are completely removed and we encourage you to use the default
-  dependency injection override to change services configuration if you need to.
+  dependency injection override to change the default values for the removed service configurations if you need to.
 
   If you need to change something that you believe it should be handled somehow in configuration,
   please open an issue and we will discuss it.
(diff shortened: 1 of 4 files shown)


VincentLanglet (dismissed):
Should we add this to the upgrade note ?

phansys (commented):
I am not up to date with these changes.
Is there a proposed replacement for these configuration points?
Do we have documentation regarding this?

jordisala1991:
This extension points are not documented and there is not proposed replacement. They are dangerous because no one is controlling what you can define on this parameters. On config you get nice validation and restrictions.

The idea is to override the service definition if you want to change something or open an issue if you need some extension point, to discuss if it is really worth it or it can be done in another way.

jordisala1991:
Added the UPGRADE note.

VincentLanglet (dismissed):

UPGRADE-4.0.md
```diff
@@ -1,6 +1,17 @@
 UPGRADE FROM 3.x to 4.0
 =======================
 
+### Container parameters
+
+  We have been using container parameters on SonataMediaBundle 3.x as an undocumented extension
+  points for things like classes, and configurations.
```
phansys, line 7:
```suggestion
  We have been using container parameters as undocumented extension points for things like classes and configurations.
```

UPGRADE-4.0.md
```diff
@@ -1,6 +1,17 @@
...
+
+  In SonataMediaBundle 4.0 those are completely removed and we encourage you to use the default
+  dependency injection override to change services configuration if you need to.
```
phansys, line 10:
```suggestion
  dependency injection override to change the default values for the removed service configurations if you need to.
```

phansys (dismissed):

VincentLanglet (approved):

phansys (approved):

phansys:
Thank you @jordisala1991.

PR #1988, opened 2021-06-15
Increase Psalm

jordisala1991:
## Subject

I am targeting this branch, because this is BC break

### DIFF
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="6" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" autoloader="vendor/bin/.phpunit/phpunit/vendor/autoload.php" errorBaseline="psalm-baseline.xml">
+<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="4" findUnusedPsalmSuppress="true" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" autoloader="vendor/bin/.phpunit/phpunit/vendor/autoload.php" errorBaseline="psalm-baseline.xml">
     <projectFiles>
         <directory name="src"/>
         <directory name="tests"/>
--- a/src/CDN/CloudFrontVersion3.php
+++ b/src/CDN/CloudFrontVersion3.php
@@ -116,13 +116,23 @@ final class CloudFrontVersion3 implements CDNInterface
                     'CallerReference' => $this->getCallerReference(),
                 ],
             ]);
-            $status = $result->get('Invalidation')['Status'];
 
-            if (false === array_search($status, self::AVAILABLE_STATUSES, true)) {
-                throw new \RuntimeException(sprintf('Unable to determine the flush status from the given response: "%s".', $status));
+            $invalidation = $result->get('Invalidation');
+            \assert(\is_array($invalidation));
+
+            $status = $invalidation['Status'] ?? null;
+
+            if (null === $status || !\in_array($status, self::AVAILABLE_STATUSES, true)) {
+                throw new \RuntimeException('Unable to determine the flush status from the given response.');
             }
 
-            return $result->get('Invalidation')['Id'];
+            $id = $invalidation['Id'] ?? null;
+
+            if (null === $id) {
+                throw new \RuntimeException('Unable to determine the flush id from the given response.');
+            }
+
+            return $id;
         } catch (CloudFrontException $ex) {
             throw new \RuntimeException(sprintf('Unable to flush paths "%s".', implode('", "', $paths)), 0, $ex);
         }
--- a/src/Provider/FileProvider.php
+++ b/src/Provider/FileProvider.php
@@ -246,13 +246,15 @@ class FileProvider extends BaseProvider implements FileProviderInterface
             }, 200, $headers);
         }
 
-        if (!$this->getFilesystem()->getAdapter() instanceof Local) {
-            throw new \RuntimeException('Cannot use X-Sendfile or X-Accel-Redirect with non \Sonata\MediaBundle\Filesystem\Local');
+        $adapter = $this->getFilesystem()->getAdapter();
+
+        if (!$adapter instanceof Local) {
+            throw new \RuntimeException(sprintf('Cannot use X-Sendfile or X-Accel-Redirect with non %s.', Local::class));
         }
 
         $filename = sprintf(
             '%s/%s',
-            $this->getFilesystem()->getAdapter()->getDirectory(),
+            $adapter->getDirectory(),
             $this->generatePrivateUrl($media, $format)
         );
 
--- a/tests/Controller/GalleryAdminControllerTest.php
+++ b/tests/Controller/GalleryAdminControllerTest.php
@@ -61,6 +61,11 @@ class GalleryAdminControllerTest extends TestCase
      */
     private $twig;
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L64-L68
+     */
     protected function setUp(): void
     {
         $this->container = new Container();
--- a/tests/Provider/FakeHttpWrapper.php
+++ b/tests/Provider/FakeHttpWrapper.php
@@ -97,7 +97,7 @@ class FakeHttpWrapper
         $file = __DIR__.'/../Fixtures/'.self::$ref[$path];
 
         if (!is_file($file)) {
-            var_dump('unable to retrieve the file : '.$file);
+            throw new \RuntimeException(sprintf('Unable to retrieve the file "%s".', $file));
         }
 
         $this->fp = fopen($file, $mode);
--- a/tests/Twig/Extension/MediaExtensionTest.php
+++ b/tests/Twig/Extension/MediaExtensionTest.php
@@ -133,6 +133,11 @@ class MediaExtensionTest extends TestCase
         return $this->template;
     }
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/twigphp/Twig/blob/3.x/src/TemplateWrapper.php#L24-L30
+     */
     public function getTemplateWrapper(): TemplateWrapper
     {
         if (null === $this->templateWrapper) {
--- a/tests/Validator/Constraints/ImageUploadDimensionValidatorTest.php
+++ b/tests/Validator/Constraints/ImageUploadDimensionValidatorTest.php
@@ -86,6 +86,11 @@ final class ImageUploadDimensionValidatorTest extends ConstraintValidatorTestCas
         $this->assertNoViolation();
     }
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Validator/Test/ConstraintValidatorTestCase.php#L298-L301
+     */
     public function testWithTooSmallImage(): void
     {
         $media = $this->mockMedia();
--- a/src/Command/CleanMediaCommand.php
+++ b/src/Command/CleanMediaCommand.php
@@ -68,7 +68,7 @@ final class CleanMediaCommand extends Command
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
-        $dryRun = (bool) $input->getOption('dry-run');
+        $dryRun = $input->getOption('dry-run');
         $verbose = $output->getVerbosity() >= OutputInterface::VERBOSITY_VERBOSE;
 
         $finder = Finder::create();
--- a/src/Controller/GalleryController.php
+++ b/src/Controller/GalleryController.php
@@ -42,9 +42,11 @@ final class GalleryController extends AbstractController
     }
 
     /**
+     * @param int|string $id
+     *
      * @throws NotFoundHttpException
      */
-    public function viewAction(string $id): Response
+    public function viewAction($id): Response
     {
         $gallery = $this->galleryManager->findOneBy([
             'id' => $id,
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -27,6 +27,11 @@ final class Configuration implements ConfigurationInterface
 {
     private const DB_DRIVERS = ['doctrine_orm', 'doctrine_mongodb', 'no_driver'];
 
+    /**
+     * @psalm-suppress PossiblyNullReference, PossiblyUndefinedMethod
+     *
+     * @see https://github.com/psalm/psalm-plugin-symfony/issues/174
+     */
     public function getConfigTreeBuilder()
     {
         $treeBuilder = new TreeBuilder('sonata_media');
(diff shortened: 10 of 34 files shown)


src/Command/FixMediaContextCommand.php
```diff
@@ -65,7 +65,7 @@ public function execute(InputInterface $input, OutputInterface $output): int
         }
 
         foreach ($this->mediaPool->getContexts() as $context => $contextAttrs) {
-            /** @var ContextInterface $defaultContext */
+            /** @var ContextInterface|null $defaultContext */
```
VincentLanglet, line 68: Instead of phpdoc, a PR should be done to https://github.com/sonata-project/SonataClassificationBundle/blob/3.x/src/Model/ContextManagerInterface.php#L19 in order to extend the template https://github.com/sonata-project/sonata-doctrine-extensions/blob/1.x/src/Model/ManagerInterface.php#L23

src/CDN/CloudFrontVersion3.php
```diff
@@ -118,13 +118,23 @@ public function flushPaths(array $paths): string
...
-            $status = $result->get('Invalidation')['Status'];
 
-            if (false === array_search($status, self::AVAILABLE_STATUSES, true)) {
-                throw new \RuntimeException(sprintf('Unable to determine the flush status from the given response: "%s".', $status));
+            $invalidation = $result->get('Invalidation');
+            \assert(\is_array($invalidation));
+
+            $status = $invalidation['Status'] ?? null;
+
+            if (null === $status || false === array_search($status, self::AVAILABLE_STATUSES, true)) {
```
phansys, line 127:
```suggestion
            if (null === $status || !in_array($status, self::AVAILABLE_STATUSES, true)) {
```

tests/Provider/FakeHttpWrapper.php
```diff
@@ -94,7 +94,7 @@ public function stream_open($path, $mode, $options, &$opened_path): bool
         $file = __DIR__.'/../Fixtures/'.self::$ref[$path];
 
         if (!is_file($file)) {
-            var_dump('unable to retrieve the file : '.$file);
+            throw new \RuntimeException('unable to retrieve the file : '.$file);
```
phansys, line 97:
```suggestion
            throw new \RuntimeException(sprintf('Unable to retrieve the file "%s".', $file));
```

src/Provider/FileProvider.php
```diff
@@ -241,13 +241,15 @@ public function getDownloadResponse(MediaInterface $media, $format, $mode, array
             }, 200, $headers);
         }
 
-        if (!$this->getFilesystem()->getAdapter() instanceof Local) {
-            throw new \RuntimeException('Cannot use X-Sendfile or X-Accel-Redirect with non \Sonata\MediaBundle\Filesystem\Local');
+        $adapter = $this->getFilesystem()->getAdapter();
+
+        if (!$adapter instanceof Local) {
+            throw new \RuntimeException(sprintf('Cannot use X-Sendfile or X-Accel-Redirect with non %s', Local::class));
```
phansys, line 247:
```suggestion
            throw new \RuntimeException(sprintf('Cannot use X-Sendfile or X-Accel-Redirect with non %s.', Local::class));
```

phansys (dismissed):

src/CDN/CloudFrontVersion3.php
```diff
@@ -118,13 +118,23 @@ public function flushPaths(array $paths): string
...
-            $status = $result->get('Invalidation')['Status'];
 
-            if (false === array_search($status, self::AVAILABLE_STATUSES, true)) {
-                throw new \RuntimeException(sprintf('Unable to determine the flush status from the given response: "%s".', $status));
+            $invalidation = $result->get('Invalidation');
+            \assert(\is_array($invalidation));
+
+            $status = $invalidation['Status'] ?? null;
+
+            if (null === $status || !in_array($status, self::AVAILABLE_STATUSES, true)) {
```
phansys, line 127: Leverage the compiler optimizations and make PHP-CS-Fixer happy.
```suggestion
            if (null === $status || !\in_array($status, self::AVAILABLE_STATUSES, true)) {
```

phansys (dismissed):

src/Model/GalleryInterface.php
```diff
@@ -42,57 +42,51 @@ public function getContext();
     /**
      * @param string $context
      *
-     * @return void
+     * @return static
```
franmomu, line 45: why change to fluent interface?

psalm.xml
```diff
@@ -11,4 +11,9 @@
         <pluginClass class="Psalm\SymfonyPsalmPlugin\Plugin"/>
         <pluginClass class="Psalm\PhpUnitPlugin\Plugin"/>
     </plugins>
+    <issueHandlers>
+        <InternalClass errorLevel="suppress"/>
+        <InternalMethod errorLevel="suppress"/>
+        <InternalProperty errorLevel="suppress"/>
```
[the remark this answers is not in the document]
jordisala1991: I was adding those psalm supress directly on the files, but then I realized how it was done on AdminBundle and would like a consistency across Sonata, so I followed the same path: https://github.com/sonata-project/SonataAdminBundle/blob/master/psalm.xml#L15-L19

Basically the supress is needed for the twig classes. some parser class from twig is internal.

src/Model/GalleryInterface.php
jordisala1991: I had to change some of the Models phpdoc to make phpstan happy, then I realize we weren't using fluent interface for entities. I tried to be consisent on how the maker-bundle generates entities, because that will be how most of our users will expect the entities to be.

Do you prefer doing this change on a separate PR or you don't like fluent interfaces?

src/Command/FixMediaContextCommand.php
jordisala1991: Removed this line

jordisala1991:
Needs https://github.com/sonata-project/SonataClassificationBundle/pull/681 or similar first.

jordisala1991:
This requires more work on classification-bundle.

src/Command/FixMediaContextCommand.php
```diff
@@ -65,7 +64,6 @@ public function execute(InputInterface $input, OutputInterface $output): int
         }
 
         foreach ($this->mediaPool->getContexts() as $context => $contextAttrs) {
-            /** @var ContextInterface $defaultContext */
```
jordisala1991, line 68: The code below, the `getRootCategory` throws the same exception for several cases. When there is no rootCategory it throws an exception, we could try catch it, or implement hasRootCategory

jordisala1991: And the CategoryInterface -> getContext() always returns a Context, but that is not true, it can return null.

VincentLanglet: The `getRootCategory` is deprecated and will be removed in 4.0 (even if there is no NEXT_MAJOR comment).
Seems like it the TODO was introduced here: https://github.com/sonata-project/SonataClassificationBundle/pull/287.

I'll try to resolve this https://github.com/sonata-project/SonataClassificationBundle/pull/684

I think you should now use `getRootCategoriesForContext`.

src/Command/FixMediaContextCommand.php
```diff
@@ -78,18 +78,7 @@ public function execute(InputInterface $input, OutputInterface $output): int
...
-            if (!$defaultCategory) {
-                $output->writeln(sprintf(" > default category for '%s' is missing, creating one", $context));
-                $defaultCategory = $this->categoryManager->create();
-                $defaultCategory->setContext($defaultContext);
-                $defaultCategory->setName(ucfirst($context));
-                $defaultCategory->setEnabled(true);
-                $defaultCategory->setPosition(0);
-
-                $this->categoryManager->save($defaultCategory);
-            }
```
VincentLanglet, line 94: This should be kept.
```
$categories = $this->categoryManager->getRootCategoriesForContext($defaultContext);

if ([] === $categories) {
 ...
}
```

jordisala1991: I don't think so: https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Entity/CategoryManager.php#L252-L295

VincentLanglet: Whoah, it's unnatural having a method `getFoo` creating under the hood a category if there is none.
Some comment should be added in the classification, but maybe here too ?

First thing I thought here was "This call seems useless, we can remove it", but it's clearly not.

jordisala1991: I think the classification should clarify that it always returns the category , even if it has to create it.

Here the best IMHO would be to add integration test that ensures the category is created. But I would prefer to leave this PR without it.

psalm.xml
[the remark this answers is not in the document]
franmomu: [I don't know why they were added in `admin-bundle` either 😬 ](https://github.com/sonata-project/SonataAdminBundle/pull/6977#discussion_r611139782).

IMO it's better to at least point which files or directories, so we'll get notice if a class/method/property of other library gets internal in another part of the project.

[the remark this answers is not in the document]
VincentLanglet: If I remember we have internal method in our test

src/Model/GalleryInterface.php
franmomu: Well, I don't like fluent interfaces for mutable objects (except for builders), I try to avoid them, not sure if they make sense in the interfaces for models.

Ref: https://ocramius.github.io/blog/fluent-interfaces-are-evil/

jordisala1991: I will revert those changes because they are not needed to bump phpstan. IMO fluent entity is not a big deal but no strong feelings to not have them.

psalm.xml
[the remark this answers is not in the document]
jordisala1991: I don't think the psalm.xml is a good place to add all those comments @franmomu . We either live with those or we cange them (also in admin-bundle) to be directly a supress on the file where the internal method is used.

IMO it is better to supress directly on the file so, as you already said, we get nice error messages is something else becomes internal.

src/Command/FixMediaContextCommand.php
jordisala1991: Fixed

src/Model/GalleryInterface.php
jordisala1991: Revert

psalm.xml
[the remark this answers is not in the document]
jordisala1991: Fixed.

tests/Controller/GalleryAdminControllerTest.php
```diff
@@ -48,6 +48,11 @@ class GalleryAdminControllerTest extends TestCase
 
     private $twig;
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/symfony/symfony/blob/5.4/src/Symfony/Bundle/FrameworkBundle/Controller/AbstractController.php#L64-L68
```
VincentLanglet, line 54: We should create an issue/PR in Symfony for this.

tests/Twig/Extension/MediaExtensionTest.php
```diff
@@ -133,6 +133,11 @@ public function getTemplate(): object
         return $this->template;
     }
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/twigphp/Twig/blob/3.x/src/TemplateWrapper.php#L24-L30
```
VincentLanglet, line 139: We should create an issue/PR for this in twig.

tests/Validator/Constraints/ImageUploadDimensionValidatorTest.php
```diff
@@ -86,6 +86,11 @@ public function testWithNoConstraints(): void
         $this->assertNoViolation();
     }
 
+    /**
+     * @psalm-suppress InternalMethod
+     *
+     * @see https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Validator/Test/ConstraintValidatorTestCase.php#L298-L301
```
VincentLanglet, line 92: I don't understand the comment, `class ConstraintViolationAssertion` is not used here

jordisala1991: the `atPath` is a call to that class uses it.

tests/Twig/Extension/MediaExtensionTest.php
jordisala1991: Why?, we are mocking a lot of internal twig things. I am not sure they will remove the internal here.

tests/Controller/GalleryAdminControllerTest.php
jordisala1991: Also don't think they will mark them not internal just for testing purposes.

psalm.xml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="6" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" autoloader="vendor/bin/.phpunit/phpunit/vendor/autoload.php" errorBaseline="psalm-baseline.xml">
+<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="4" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" autoloader="vendor/bin/.phpunit/phpunit/vendor/autoload.php" errorBaseline="psalm-baseline.xml">
```
franmomu, line 2:
```suggestion
<psalm xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="https://getpsalm.org/schema/config" errorLevel="4" findUnusedPsalmSuppress="true" resolveFromConfigFile="true" xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd" autoloader="vendor/bin/.phpunit/phpunit/vendor/autoload.php" errorBaseline="psalm-baseline.xml">
```
with `findUnusedPsalmSuppress` we'll get a warning if the `psalm-suppress` is not valid anymore.

tests/Controller/GalleryAdminControllerTest.php
VincentLanglet: I would say it worth the try. We might end with another way to test this code.

tests/Twig/Extension/MediaExtensionTest.php
VincentLanglet: It might be interesting to start a discussion in order to know/determine the right way to have a TemplateWrapper in the tests. I already remove some `@internal` annotation in Symfony, they seems opened to discussion.

tests/Validator/Constraints/ImageUploadDimensionValidatorTest.php
VincentLanglet: So then, again an issue could be created.

They added a lot of assertion https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Validator/Test/ConstraintValidatorTestCase.php#L265 ; if what we're wanting to do is not possible now without using an internal class ; I would say a feature request to add extra protected method could be a nice idea.

tests/Controller/GalleryAdminControllerTest.php
franmomu: in this case I guess the recommended way is using functional tests

tests/Twig/Extension/MediaExtensionTest.php
jordisala1991: I don't think this is suposed to be tested as unit test. We should change this to be functional, but it is not the purpose of this PR.

src/Command/FixMediaContextCommand.php
VincentLanglet: The PR of Classification is now merged.

PR #1993, opened 2021-06-25
Remove usages of datagrid-bundle

jordisala1991:
## Subject

I am targeting this branch, because {reason}.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added some `Class::newMethod()` to do great stuff.

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


docs/reference/advanced_configuration.rst
```diff
@@ -183,5 +183,5 @@ Full configuration options:
     jms_serializer:
         metadata:
             directories:
-                - { name: 'sonata_datagrid', path: '%kernel.project_dir%/vendor/sonata-project/datagrid-bundle/src/Resources/config/serializer', namespace_prefix: 'Sonata\DatagridBundle' }
+                - { name: 'sonata_admin', path: '%kernel.project_dir%/vendor/sonata-project/admin-bundle/src/Resources/config/serializer', namespace_prefix: 'Sonata\AdminBundle' }
```
jordisala1991, line 186: This should be fixed, the thing is where should this serializer config should be? on Admin or on persistence?

Ref: https://github.com/sonata-project/SonataDatagridBundle/blob/3.x/src/Resources/config/serializer/Pager.BasePager.xml

(IMO it would be better if it is placed on the admin-bundle)

wdyt @VincentLanglet

jordisala1991:
Test fails because we need to adapt some mocks to the new Pager implementation.

docs/reference/advanced_configuration.rst
VincentLanglet: What is this file used for ?

composer.json
```diff
@@ -36,6 +36,7 @@
         "psr/http-factory": "^1.0",
         "psr/http-message": "^1.0",
         "psr/log": "^1.0",
+        "sonata-project/admin-bundle": "^4.0@dev",
```
VincentLanglet, line 39: Datagrid were a require-dev. Why now Admin is a require ?

jordisala1991: Because datagrid being a require-dev was a mistake, I got big errors when trying to use sonata-media 4 without having the datagrid-bundle

docs/reference/advanced_configuration.rst
jordisala1991: Not really sure TBH because I don't use the APIs from Sonata. I think it is used for the FOSRest, in order to return the Pager and have it serialized to json/xml.

composer.json
jordisala1991: Ref: https://github.com/Runroom/runroom-packages/pull/92/checks?check_run_id=2892702348

docs/reference/advanced_configuration.rst
VincentLanglet: Isn't better for bundle like mediaBundle to have his own serializer config in case you want to serialize the Pager differently ?

jordisala1991: With that approach we will end with N Pager serializer configurations, that are all the same, because the fields on the Pager class are always the same, not sure if I understood you correctly.

VincentLanglet: Yes kinda.

But on the otherside why would we need to provide a serializer configuration in the SonataAdmin, not use by SonataAdmin and only for the Pager. A bundle could start to serialize the datagrid. A user could start to serialize the ProxyQuery. Isn't it ?

We can add it to SonataAdmin for simplicity for sure. But if we do, it would require some tests functional tests too.

src/Document/MediaManager.php
```diff
@@ -13,7 +13,7 @@
 
 namespace Sonata\MediaBundle\Document;
 
-use Sonata\DatagridBundle\Pager\PagerInterface;
+use Sonata\AdminBundle\Datagrid\PagerInterface;
```
core23, line 16: This looks wrong. You should use the bundle without the admin stuff

core23 (changes requested):

src/Document/MediaManager.php
VincentLanglet: There is no right or wrong.

The DatagridBundle is just a bunch of duplicated code from the SonataAdminBundle. All this code is now outdated as explained here: https://github.com/sonata-project/SonataDatagridBundle/pull/340#issuecomment-864523118. There is multiple misconception in the DatagridBundle: https://github.com/sonata-project/SonataDatagridBundle/issues/235.

Taking a look at all the package using the DatagridBundle, half of them are also using the AdminBundle which means users and with two different PagerInterface, two different ProxyQueryInterface, and so on. The other half of these bundles are still highly related to the AdminBundle and use it as de dev dependency. I cannot be sure it works without the AdminBundle and if it does, I'm cannot sure a lot of developper are using it without the AdminBundle.

Making the DatagridBundle up-to-date would require a lot of changes, which means a lot of contributor time and a new major version. Looking at the activity of the contributor and the maintainer, we don't have this time. Releasing the AdminBundle 4 was hard enough.

It's better to stop the mess with the DatagridBundle, so we're planning to deprecate it to be more realistic with the time we have. Less package will provide a better quality. Requiring the AdminBundle instead of the DatagridBundle is not a big change. It will even ends with smaller vendors for package witch were requiring both.

core23: I can understand, that there's something wrong and that there are some misconceptions between some bundle, but using admin related classes inside entity repositories looks more wrong than before.

I'm a huge fan of [deptrac](https://github.com/qossmic/deptrac) to define layers that should not communitate to each other. Maybe be should define rules which community is allowed by design and which isn't. The entity -> admin dependency is a no go.

VincentLanglet:
I think we can close this for now. Datagrid is not required anymore in next major

PR #1995, opened 2021-06-26
Move datagrid-bundle to require section

jordisala1991:
## Subject

I am targeting this branch, because this is where it is needed.


jordisala1991:
We need to finish this one in order to remove datagrid-bundle: https://github.com/sonata-project/SonataMediaBundle/pull/1993

But for now, it is more correct to have it on require section, all the other sonata bundles do it this way and if you try to use this bundle without datagrid-bundle it fails because it don't find the required interfaces.

VincentLanglet (approved):

PR #2000, opened 2021-06-29
Fix build

jordisala1991:
## Subject

I am targeting this branch, because the build is broken.

### DIFF
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -20,6 +20,7 @@ use Sonata\AdminBundle\Form\Type\ModelListType;
 use Sonata\AdminBundle\Object\Metadata;
 use Sonata\AdminBundle\Object\MetadataInterface;
 use Sonata\ClassificationBundle\Model\CategoryManagerInterface;
+use Sonata\ClassificationBundle\Model\ContextManagerInterface;
 use Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Sonata\MediaBundle\Provider\Pool;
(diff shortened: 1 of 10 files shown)


src/Admin/BaseMediaAdmin.php
```diff
@@ -90,7 +103,8 @@ public function configurePersistentParameters(): array
         $categoryId = $this->getRequest()->get('category');
 
         if (null !== $this->categoryManager && !$categoryId) {
-            $category = $this->categoryManager->getRootCategory($context);
+            $categories = $this->categoryManager->getRootCategoriesForContext($this->contextManager->find($context));
+            $category = $categories[0];
```
jordisala1991, line 107: I don't like a lot having to do this kind of access to the array, but previously this returned one root category, now returns all (whatever that means). Same for the other places.

VincentLanglet: Category are set this way
```
$this->categories[$context->getId()][$category->getId()] = $category;
```
So I'm not sure `[0]` is the right way to do it.
`current($categories)` maybe instead ?

src/Controller/MediaAdminController.php
```diff
@@ -70,8 +72,12 @@ public function listAction(Request $request): Response
...
 
         $rootCategory = null;
-        if ($this->has('sonata.media.manager.category')) {
-            $rootCategory = $this->get('sonata.media.manager.category')->getRootCategory($context);
+        if ($this->has('sonata.media.manager.category') && $this->has('sonata.media.manager.context')) {
+            $rootCategories = $this->get('sonata.media.manager.category')->getRootCategoriesForContext(
+                $this->get('sonata.media.manager.context')->find($context)
+            );
+
+            $rootCategory = $rootCategories[0];
```
VincentLanglet, line 80: Same, `current($rootCategories)` is better I think

src/Admin/BaseMediaAdmin.php
jordisala1991: Updated.

src/Controller/MediaAdminController.php
jordisala1991: Fixed.

VincentLanglet (approved):

PR #2001, opened 2021-06-29
Add missing service declaration

tifabien:
## Subject

This fix the missing MediaAdminController and GalleryAdminController service declarations.

I am targeting this branch, because only applicable on master branch.

### DIFF
--- a/src/Resources/config/controllers.php
+++ b/src/Resources/config/controllers.php
@@ -40,13 +40,15 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ->tag('container.service_subscriber')
             ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
 
-        ->set(MediaAdminController::class)
-            ->public()
+        ->set('sonata.media.controller.media.admin', MediaAdminController::class)
             ->tag('container.service_subscriber')
             ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
-
-        ->set(GalleryAdminController::class)
+            ->alias(MediaAdminController::class, 'sonata.media.controller.media.admin')
             ->public()
+
+        ->set('sonata.media.controller.gallery.admin', GalleryAdminController::class)
             ->tag('container.service_subscriber')
-            ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)]);
+            ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
+            ->alias(GalleryAdminController::class, 'sonata.media.controller.gallery.admin')
+            ->public();
 };
--- a/src/Resources/config/controllers.php
+++ b/src/Resources/config/controllers.php
@@ -41,14 +41,12 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
 
         ->set('sonata.media.controller.media.admin', MediaAdminController::class)
+            ->public()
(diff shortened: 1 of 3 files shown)


VincentLanglet (commented):
@jordisala1991 Does this mean that every custom CRUDController has to be tag as a service_subscriber with a container set ?

@dmaicher Do you use custom CRUDController with Sonata 4? Did you have to do this ?
Or is done automatically for the dev thanks to autowire/autoconfigure ?

If not, can we do this with a compilerPass in order to avoid extra config for every crud controller ? cc @sonata-project/contributors

jordisala1991:
With autoconfigure and autowire this goes away.

src/Resources/config/controllers.php
```diff
@@ -36,5 +38,15 @@
                 new ReferenceConfigurator('sonata.media.pool'),
             ])
             ->tag('container.service_subscriber')
+            ->call('setContainer', [new ReferenceConfigurator(ContainerInterface::class)])
+
+        ->set(MediaAdminController::class)
```
jordisala1991, line 43: assign a name to this please like above

jordisala1991:
The build will get fixed in #2000

src/Resources/config/controllers.php
tifabien: Name added. 
Note that I had to add also an alias.

VincentLanglet: The `sonata.media.controller.gallery` does not have an alias.
Why the alias would be require here @jordisala1991 ?

jordisala1991: I don't know if it should be needed. Maybe changing the admin definition to not rely on the controller class, but the service id.

tifabien: I confirm that if I don't put this alias, the `sonata.media.controller.media.admin` service is not used. Seems related to the Route builder which deduct the controller and this controller is then instanciated by the SF Controller resolver without its correct configuration.

VincentLanglet (dismissed):

src/Resources/config/controllers.php
jordisala1991: You need to change both this files:

https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_mongodb_admin.php
https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_orm_admin.php

Concretely those lines:

https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_mongodb_admin.php#L42
https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_mongodb_admin.php#L68
https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_orm_admin.php#L42
https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/doctrine_orm_admin.php#L69

jordisala1991 (changes requested):
Waiting for changes

tifabien:
> Waiting for changes

Sorry @jordisala1991 but what do you want for change?

jordisala1991:
> > Waiting for changes
> 
> Sorry @jordisala1991 but what do you want for change?

The ones mentioned here: https://github.com/sonata-project/SonataMediaBundle/pull/2001#discussion_r661331333

jordisala1991 (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #2002, opened 2021-06-30
Increase phpstan

jordisala1991:
## Subject

I am targeting this branch, because this is where it applies.

### DIFF
--- a/src/Controller/MediaAdminController.php
+++ b/src/Controller/MediaAdminController.php
@@ -80,7 +80,7 @@ final class MediaAdminController extends CRUDController
         $datagrid->setValue('context', null, $context);
 
         $rootCategory = null;
-        if ($this->has('sonata.media.manager.category')) {
+        if ($this->has('sonata.media.manager.category') && $this->has('sonata.media.manager.context')) {
             $rootCategories = $this->get('sonata.media.manager.category')->getRootCategoriesForContext(
                 $this->get('sonata.media.manager.context')->find($context)
             );
--- a/src/Controller/MediaAdminController.php
+++ b/src/Controller/MediaAdminController.php
@@ -80,7 +80,7 @@ final class MediaAdminController extends CRUDController
         $datagrid->setValue('context', null, $context);
 
         $rootCategory = null;
-        if ($this->has('sonata.media.manager.category') && $this->has('sonata.media.manager.context')) {
+        if ($this->has('sonata.media.manager.category')) {
             $rootCategories = $this->get('sonata.media.manager.category')->getRootCategoriesForContext(
                 $this->get('sonata.media.manager.context')->find($context)
             );
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -152,15 +152,13 @@ abstract class Gallery implements GalleryInterface
      */
     public function reorderGalleryItems()
     {
-        if ($this->getGalleryItems() && $this->getGalleryItems() instanceof \IteratorAggregate) {
-            $iterator = $this->getGalleryItems()->getIterator();
-            \assert($iterator instanceof \ArrayIterator);
+        $iterator = $this->getGalleryItems()->getIterator();
+        \assert($iterator instanceof \ArrayIterator);
 
-            $iterator->uasort(static function (GalleryItemInterface $a, GalleryItemInterface $b) {
-                return $a->getPosition() <=> $b->getPosition();
-            });
+        $iterator->uasort(static function (GalleryItemInterface $a, GalleryItemInterface $b) {
+            return $a->getPosition() <=> $b->getPosition();
+        });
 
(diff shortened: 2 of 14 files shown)


src/Controller/MediaAdminController.php
```diff
@@ -72,7 +72,7 @@ public function listAction(Request $request): Response
         $datagrid->setValue('context', null, $context);
 
         $rootCategory = null;
-        if ($this->has('sonata.media.manager.category') && $this->has('sonata.media.manager.context')) {
+        if ($this->has('sonata.media.manager.category')) {
```
jordisala1991, line 75: adding this checks makes phpstan not happy...

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Controller/MediaAdminController.php
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
  80     Call to method Symfony\Bundle\FrameworkBundle\Controller\AbstractController::has() with 'sonata.media…' will always evaluate to false.
 ------ ----------------------------------------------------------------------------------------------------------------------------------------
```

Super strange, if I try to split the if like:

```
if ($this->has()) {
}
if ($this->has() {
}
```

it works, but having the condition in the same if or nested it fails.

VincentLanglet: Both has should be kept since we're using both get.

I'll take a look, but maybe it will be better to ignore the error and report an issue to phpstan.

src/Model/Gallery.php
```diff
@@ -152,15 +152,13 @@ public function removeGalleryItem(GalleryItemInterface $galleryItem): void
      */
     public function reorderGalleryItems()
     {
-        if ($this->getGalleryItems() && $this->getGalleryItems() instanceof \IteratorAggregate) {
-            $iterator = $this->getGalleryItems()->getIterator();
-            \assert($iterator instanceof \ArrayIterator);
+        $iterator = $this->getGalleryItems()->getIterator();
+        \assert($iterator instanceof \ArrayIterator);
```
VincentLanglet, line 156: Why is this assert needed ?

jordisala1991: Otherwise the uasort method is not detected:

```
 ------ -----------------------------------------------------------------------------------------------------------------
  Line   src/Model/Gallery.php
 ------ -----------------------------------------------------------------------------------------------------------------
  157    Call to an undefined method Traversable<(int|string), Sonata\MediaBundle\Model\GalleryItemInterface>::uasort().
 ------ -----------------------------------------------------------------------------------------------------------------
```

src/Controller/MediaAdminController.php
jordisala1991: I think is because of the phpstan-symfony plugin, if I disable it, the error goes away.

jordisala1991: Just added an issue: https://github.com/phpstan/phpstan-symfony/issues/178

VincentLanglet: Seems like there is something about in the doc Seems related to https://github.com/phpstan/phpstan-symfony#constant-hassers

jordisala1991: I tried to set it to false, didn't worked, will try later again.

src/Model/Gallery.php
VincentLanglet: We cannot be sure that the iterator is an ArrayIterator. It's not mandatory by the Collection interface.
If someone use a custom implementation it won't work.

It would be better to use another implementation or at least to throw an exception if it's not an array iterator instead of just asserting it.

src/Controller/MediaAdminController.php
VincentLanglet: Isn't `sonata.classification.manager.context` instead ?

https://github.com/sonata-project/SonataClassificationBundle/blob/master/src/Resources/config/orm.xml#L23

jordisala1991: Nope, it is sonata.media.manager.context: https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/category.php

those alias are created because on media-bundle you can decide (even if you have classification-bundle installed) that you don't want to use it.

Maybe it is a bit weird that works this way but for now it is how it is implemented (maybe we can deprecate that later).

jordisala1991: I ignored the error for now, the code is correct.

src/Model/Gallery.php
jordisala1991: Added an error throw. Anyway I don't think adding own collection implementation is something that can be done.

src/Controller/MediaAdminController.php
VincentLanglet: It's fixed in the last version of phpstan

VincentLanglet (approved):

PR #2005, opened 2021-07-01
Set sonata.media.pool service as public

tifabien:
## Subject

Needed because of the InlineValidator in form-extension repo (See https://github.com/sonata-project/form-extensions/blob/master/src/Validator/InlineValidator.php#L53).

I am targeting this branch, because only applicable on master.

### DIFF
--- a/src/Resources/config/providers.php
+++ b/src/Resources/config/providers.php
@@ -27,6 +27,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
     $containerConfigurator->services()
 
         ->set('sonata.media.pool', Pool::class)
+            ->public()
             ->args([''])
 
         ->alias(Pool::class, 'sonata.media.pool')


jordisala1991:
Can you point how we end up requesting `sonata.media.pool` on the inlineValidator from `form-extension`?

tifabien:
`sonata.media.pool` is requested in https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Resources/config/validation.xml#L5

Because we pass a string to the inlineValidator it tries to fetch the service from the container.

VincentLanglet (approved):
Maybe we should create an issue to change to way this constraint works by using dependency injection instead ?

jordisala1991 (approved):

jordisala1991:
Thanks @tifabien

PR #2007, opened 2021-07-03
Fix errors for phpstan level 6

jordisala1991:
## Subject

I am targeting this branch, because this is breaks bc.


jordisala1991:
I think I've found another covariance problem @VincentLanglet :

```
 ------ ----------------------------------------------------------------------- 
  Line   src/Controller/GalleryAdminController.php                              
 ------ ----------------------------------------------------------------------- 
  56     Parameter #1 $admin of method                                          
         Sonata\AdminBundle\Bridge\Exporter\AdminExporter::getAvailableFormats  
         () expects Sonata\AdminBundle\Admin\AdminInterface<object>,            
         Sonata\AdminBundle\Admin\AdminInterface<Sonata\MediaBundle\Model\Gall  
         eryInterface> given.                                                   
 ------ ----------------------------------------------------------------------- 

 ------ ----------------------------------------------------------------------- 
  Line   src/Controller/MediaAdminController.php                                
 ------ ----------------------------------------------------------------------- 
  118    Parameter #1 $admin of method                                          
         Sonata\AdminBundle\Bridge\Exporter\AdminExporter::getAvailableFormats  
         () expects Sonata\AdminBundle\Admin\AdminInterface<object>,            
         Sonata\AdminBundle\Admin\AdminInterface<Sonata\MediaBundle\Model\Medi  
         aInterface> given.                                                     
 ------ ----------------------------------------------------------------------- 
```

VincentLanglet:
Yes, it's because of the way https://github.com/sonata-project/SonataAdminBundle/blob/master/src/Bridge/Exporter/AdminExporter.php#L37-L53 is defined.

`AdminInterface<object>` doesn't accept a more strict definition like `AdminInterface<Media>`.
It wouldn't happen if the SonataAdminInterface had a `template-covariant` instead of a template, but it's impossible since there is method like `setSubject` with the template in the `@param`.

I think it wouldn't happen too if the method were using `AdminInterface` instead of `AdminInterface<object>` but phpstan level 6 is asking for filling template in generics phpdoc.

I'm not sure what is the best solution.

I think we can add a template to the AdminExporter class but, I'm not sure it's the right way here.
Not sure if changing the exporter phpdoc to 
```
@phpstan-template T of object
@phpstan-param AdminInterface<T>
```
could solve the solution, but same I don't know if it has some meaning to not really use a template.

Dunno if introducing a new notation for methods like `@phpstan-covariant` could be a solution with the idea that
- A `@phpstan-covariant` method can use method which could have been defined by a `@template-covariant` method (like `getExportFormats` here)
- A `@phpstan-covariant` method cannot use method which break covariance (like `setSubject`)
- A method using a `@phpstan-covariant` should not complain about covariance.
But that put the issue on a whole new level.

Maybe @ondrejmirtes or @muglug / @weirdan / @orklah could give us some advice about this issue, because we could encounter it more and more often ; so we should start know to fix it the right way.

jordisala1991:
Another error that should be fixed on our side:

```
 ------ --------------------------------------------------------------------- 
  Line   src/Block/FeatureMediaBlockService.php                               
 ------ --------------------------------------------------------------------- 
  94     Method                                                               
         Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
         has no return typehint specified.                                    
  94     Method                                                               
         Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
         has parameter $media with no typehint specified.                     
 ------ --------------------------------------------------------------------- 
```

This method not even exist on Blockbundle 4.0, it was deprecated. Not sure how should we add this css, but we might need to understand how it was done before and change this implementation.

VincentLanglet:
> Another error that should be fixed on our side:
> 
> ```
>  ------ --------------------------------------------------------------------- 
>   Line   src/Block/FeatureMediaBlockService.php                               
>  ------ --------------------------------------------------------------------- 
>   94     Method                                                               
>          Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
>          has no return typehint specified.                                    
>   94     Method                                                               
>          Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
>          has parameter $media with no typehint specified.                     
>  ------ --------------------------------------------------------------------- 
> ```
> 
> This method not even exist on Blockbundle 4.0, it was deprecated. Not sure how should we add this css, but we might need to understand how it was done before and change this implementation.

According to https://github.com/sonata-project/SonataBlockBundle/pull/510, the assets should be moved to the template.

VincentLanglet:
> Yes, it's because of the way https://github.com/sonata-project/SonataAdminBundle/blob/master/src/Bridge/Exporter/AdminExporter.php#L37-L53 is defined.
> 
> `AdminInterface<object>` doesn't accept a more strict definition like `AdminInterface<Media>`.
> It wouldn't happen if the SonataAdminInterface had a `template-covariant` instead of a template, but it's impossible since there is method like `setSubject` with the template in the `@param`.
> 
> I think it wouldn't happen too if the method were using `AdminInterface` instead of `AdminInterface<object>` but phpstan level 6 is asking for filling template in generics phpdoc.
> 
> I'm not sure what is the best solution.
> 
> I think we can add a template to the AdminExporter class but, I'm not sure it's the right way here.
> Not sure if changing the exporter phpdoc to
> 
> ```
> @phpstan-template T of object
> @phpstan-param AdminInterface<T>
> ```
> 
> could solve the solution, but same I don't know if it has some meaning to not really use a template.
> 
> Dunno if introducing a new notation for methods like `@phpstan-covariant` could be a solution with the idea that
> 
> * A `@phpstan-covariant` method can use method which could have been defined by a `@template-covariant` method (like `getExportFormats` here)
> * A `@phpstan-covariant` method cannot use method which break covariance (like `setSubject`)
> * A method using a `@phpstan-covariant` should not complain about covariance.
>   But that put the issue on a whole new level.
> 
> Maybe @ondrejmirtes or @muglug / @weirdan / @orklah could give us some advice about this issue, because we could encounter it more and more often ; so we should start know to fix it the right way.

I've created
https://github.com/vimeo/psalm/issues/6042
https://github.com/phpstan/phpstan/issues/5254

Because I think a new annotation would be a nice way to solve this.

VincentLanglet:
@jordisala1991 I've made https://github.com/sonata-project/SonataAdminBundle/pull/7302, can you try if it solve the issue ?

I added this solution to the issue i've created to phpstan/psalm so I also prefer to wait for response from them in order to know if it can be considered as a right way to solve our issue. (And maybe can be documented)

jordisala1991:
The PR fixes the issues here.

jordisala1991:
> > Another error that should be fixed on our side:
> > ```
> >  ------ --------------------------------------------------------------------- 
> >   Line   src/Block/FeatureMediaBlockService.php                               
> >  ------ --------------------------------------------------------------------- 
> >   94     Method                                                               
> >          Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
> >          has no return typehint specified.                                    
> >   94     Method                                                               
> >          Sonata\MediaBundle\Block\FeatureMediaBlockService::getStylesheets()  
> >          has parameter $media with no typehint specified.                     
> >  ------ --------------------------------------------------------------------- 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This method not even exist on Blockbundle 4.0, it was deprecated. Not sure how should we add this css, but we might need to understand how it was done before and change this implementation.
> 
> According to [sonata-project/SonataBlockBundle#510](https://github.com/sonata-project/SonataBlockBundle/pull/510), the assets should be moved to the template.

To not block this PR (because it is not really related to the fact that this block does not work), I have created an issue for this change: https://github.com/sonata-project/SonataMediaBundle/issues/2009

VincentLanglet (approved):

PR #2008, opened 2021-07-05
Increase psalm to level 5

jordisala1991:
## Subject

I am targeting this branch, because psalm is level 6 here.

### DIFF
--- a/src/CDN/CloudFrontVersion3.php
+++ b/src/CDN/CloudFrontVersion3.php
@@ -121,8 +121,12 @@ final class CloudFrontVersion3 implements CDNInterface
 
             $status = $invalidation['Status'] ?? null;
 
-            if (null === $status || !\in_array($status, self::AVAILABLE_STATUSES, true)) {
-                throw new \RuntimeException('Unable to determine the flush status from the given response.');
+            if (null === $status) {
+                throw new \RuntimeException('Unable to find the flush status from the given response.');
+            }
+
+            if (!\in_array($status, self::AVAILABLE_STATUSES, true)) {
+                throw new \RuntimeException(sprintf('Unable to determine the flush status from the given response: "%s".', $status));
             }
 
             $id = $invalidation['Id'] ?? null;
--- a/src/CDN/CloudFrontVersion3.php
+++ b/src/CDN/CloudFrontVersion3.php
@@ -116,13 +116,22 @@ final class CloudFrontVersion3 implements CDNInterface
                     'CallerReference' => $this->getCallerReference(),
                 ],
             ]);
-            $status = $result->get('Invalidation')['Status'];
+            $invalidation = $result->get('Invalidation');
+            \assert(\is_array($invalidation));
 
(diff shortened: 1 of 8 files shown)


tests/App/Entity/Context.php
```diff
@@ -33,6 +33,6 @@ class Context extends BaseContext
 
     public function getId()
     {
-        return $this->id;
+        return $this->id ?? '';
```
VincentLanglet, line 36: Isn't better to return null if there is no id and check for null value if needed ?

jordisala1991: Yep, but that change needs to be done on SonataClassificationBundle.

VincentLanglet: Why ? What is the error you try to fix ?

jordisala1991: This is the Context entity that is created in order to test the integration with classification-bundle

(Context entity comes from classification-bundle)

VincentLanglet: I've made this https://github.com/sonata-project/SonataClassificationBundle/pull/689

VincentLanglet: It's merged @jordisala1991

jordisala1991: Fixed.

VincentLanglet (approved):

src/Model/NoDriverManager.php
```diff
@@ -38,7 +38,7 @@ public function findAll(): array
      * @param int|null $limit
      * @param int|null $offset
      */
-    public function findBy(array $criteria, ?array $orderBy = null, $limit = null, $offset = null): array
```
franmomu, line 41: I don't think is necessary to remove the return type

src/CDN/CloudFrontVersion3.php
```diff
@@ -116,13 +116,22 @@ public function flushPaths(array $paths): string
...
-            $status = $result->get('Invalidation')['Status'];
+            $invalidation = $result->get('Invalidation');
+            \assert(\is_array($invalidation));
 
-            if (false === array_search($status, self::AVAILABLE_STATUSES, true)) {
-                throw new \RuntimeException(sprintf('Unable to determine the flush status from the given response: "%s".', $status));
+            $status = $invalidation['Status'] ?? null;
+
+            if (null === $status || !\in_array($status, self::AVAILABLE_STATUSES, true)) {
+                throw new \RuntimeException('Unable to determine the flush status from the given response.');
```
franmomu, line 125: Why getting rid of the status? looks like it's useful information

jordisala1991: Re add that exception (I split when having null as status from the not in the array of available statuses

src/Model/NoDriverManager.php
jordisala1991: Re added

franmomu (approved):

VincentLanglet (approved):

PR #2010, opened 2021-07-12
Increase PHPStan to level 7

jordisala1991:
## Subject

I am targeting this branch, because this only applies to master.


jordisala1991:
@VincentLanglet this is where I am trying to add the console command analysis to phpstan, but I still get a ton of errors related to input parameters. Not sure if that is normal, but on admin bundle I don't see that much assertions needed for the `->getInput('foo')` variables.

src/Model/MediaInterface.php
```diff
@@ -28,7 +28,7 @@ interface MediaInterface
     public const MISSING_BINARY_REFERENCE = 'missing_binary_content';
 
     /**
-     * @return int|string|object|null
+     * @return int|string|null
```
core23, line 31: This is correct? What if I want to use a UUID?

phpstan-console-application.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
```
core23, line 3: This file is for tests? If yes, you should move it to the `tests` directory

core23 (changes requested):

phpstan-console-application.php
jordisala1991: This is the same as admin-bundle

src/Model/MediaInterface.php
jordisala1991: In admin-bundle they are not fully supported, and If you look closely at code here, you can't either, there are multiple places where $media->getId() is used directly as an array key.

jordisala1991: This is a discussion that needs to happen on whole sonata ecosystem, and I would love to support uuids everywhere (in fact I added this return previously. But the reality is that it does not support, and it is not only a matter of adding support here only. multiple changes are required on several bundles.

phpstan-console-application.php
VincentLanglet: Yes, since it's phpstan-related, it's better to keep it with phpstan files.

src/Model/MediaInterface.php
VincentLanglet: Agree, if we want to support UUID an issue should first be opened on SonataAdmin because atm the phpdoc is not providing support and some code is using it as string|int.

src/Resizer/CropResizer.php
```diff
@@ -89,9 +89,7 @@ public function getBox(MediaInterface $media, array $settings): Box
     }
 
     /**
-     * @param array<string, int> $settings
-     *
-     * @phpstan-param array{width: int, height: int} $settings
```
VincentLanglet, line 94: Why losing this ?

tests/Provider/YouTubeProviderTest.php
```diff
@@ -34,6 +35,9 @@
  */
 class YouTubeProviderTest extends AbstractProviderTest
 {
+    /**
+     * @param (MockObject&ClientInterface)|null $client
+     */
     public function getProvider(?object $client = null, ?RequestFactoryInterface $messageFactory = null): MediaProviderInterface
```
VincentLanglet, line 41: As soon as you add a phpdoc to a method, psalm is not inheriting the phpdoc anymore.

So you have to add `@return YouTubeProvider`

VincentLanglet (commented):
The command is using `setDefinition` https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Command/AddMassMediaCommand.php#L53

in SonataAdmin, the method `addOption` is used instead. Maybe it's the reason ?

jordisala1991:
> The command is using `setDefinition` https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Command/AddMassMediaCommand.php#L53
> 
> 
> 
> in SonataAdmin, the method `addOption` is used instead. Maybe it's the reason ?

I will try that tomorrow

jordisala1991:
> The command is using `setDefinition` https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Command/AddMassMediaCommand.php#L53
> 
> in SonataAdmin, the method `addOption` is used instead. Maybe it's the reason ?

I have changed that, but it is still the same... not sure if I don't understand the feature and it is normal to having to assert those as string or it should be detected?

phpstan-console-application.php
core23: Okay, please exclude this file from the archive using `.gitattributes`

core23 (dismissed):

phpstan-console-application.php
VincentLanglet: https://github.com/sonata-project/dev-kit/pull/1588

VincentLanglet:
> I have changed that, but it is still the same... not sure if I don't understand the feature and it is normal to having to assert those as string or it should be detected?

The feature should avoid to assert those as string. I have no clue, it seems to be the same configuration than sonataAdmin.
Opening an issue to phpstan-symfony could be an idea, in order to get some help from @ondrejmirtes

src/Command/AddMassMediaCommand.php
```diff
@@ -61,11 +61,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $fp = $this->getFilePointer($input, $output);
 
         $delimiter = $input->getOption('delimiter');
-        \assert(null === $delimiter || is_string($delimiter));
+        \assert(\is_string($delimiter));
```
VincentLanglet, line 64: I think it's better to add this to the baseline rather than doing those always true assert.
When we'll configure correctly phpstan, the baseline will be reported as useless.

And what do you think about opening an issue to the phpstan-symfony repository ? :)

VincentLanglet (changes requested):

tests/Provider/YouTubeProviderTest.php
jordisala1991: Fixed in other way

phpstan-console-application.php
jordisala1991: Merged and fixed.

src/Model/MediaInterface.php
jordisala1991: Issue opened: https://github.com/sonata-project/SonataAdminBundle/issues/7327

src/Command/AddMassMediaCommand.php
jordisala1991: Yep, I just wanted to get a green build first. 

I don't really understand why it doesn't detect it. But to be honest I don't understand how phpstan can know if the parameter will be an array or an string, do you know the internals of this? There are some parameters that we provide a default value, so it could know because of that, but we have a lot of parameters without default value.

I will try to open an issue on phpstan-symfony

src/Resizer/CropResizer.php
jordisala1991: I added those phpdoc initially but not sure if they are correct:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   src/Resizer/CropResizer.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  83     Parameter #1 $settings of method Sonata\MediaBundle\Resizer\CropResizer::createTargetBox() expects array('width' => int, 'height' => int), array<string, mixed> given.
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

The thing is this array $settings is passed on a lot of places, and we are giving it array<string, mixed> on every place.

This is one of the places where the array has more than the width and height: https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Thumbnail/FormatThumbnail.php#L94-L106

Maybe we should fully understand all options that could be there and create a type.

src/Command/AddMassMediaCommand.php
VincentLanglet: Sure, it's done by this dynamic return extension:
https://github.com/phpstan/phpstan-symfony/blob/master/src/Type/Symfony/InputInterfaceGetOptionDynamicReturnTypeExtension.php
especially this code:
https://github.com/phpstan/phpstan-symfony/blob/master/src/Type/Symfony/InputInterfaceGetOptionDynamicReturnTypeExtension.php#L65-L79
It look at the option definition to find out his type.

src/Resizer/CropResizer.php
jordisala1991: I have pushed a new commit to fix this, please take a look to see if I did it right.

VincentLanglet: This seems the right way to fix it to me :)

src/Command/AddMassMediaCommand.php
jordisala1991: I found the problem.

The commands are registered as LazyCommand because we define the title and description on a static variable.

This does not happen on SonataAdmin (missing description).

jordisala1991: https://github.com/phpstan/phpstan-symfony/issues/185 Here is the issue and I tried to fix it too with a PR there.

Ideally all Sonata commands should be as lazy as possible (they should still do the work 😄 )

jordisala1991: Fixed, phpstan will pass once the phpstan/phpstan-symfony#185 is solved.

VincentLanglet (approved):

jordisala1991:
Merged to be able to up to level 8 , If @sonata-project/contributors find something that has to be changed please comment here and I will address it on the next PR.

PR #2013, opened 2021-07-20
Increase phpstan to max level

jordisala1991:
## Subject

I am targeting this branch, because this is the only branch where it is safe to increase phpstan to max level.
This PR tries to set phpstan to max level, preparing the bundle to the next release.

### DIFF
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -102,7 +102,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
         $categoryId = $this->getRequest()->get('category');
 
-        if (null !== $this->categoryManager && !$categoryId) {
+        if (null !== $this->categoryManager && null !== $this->contextManager && !$categoryId) {
             $rootCategories = $this->categoryManager->getRootCategoriesForContext($this->contextManager->find($context));
             $rootCategory = current($rootCategories);
 
@@ -155,7 +155,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
             $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN)
         );
 
-        return new Metadata($object->getName(), $object->getDescription(), $url);
+        return new Metadata((string) $object, $object->getDescription(), $url);
     }
 
     protected function configureListFields(ListMapper $list): void
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -155,7 +155,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
             $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN)
         );
 
-        return new Metadata((string) $object, $object->getDescription(), $url);
+        return new Metadata($this->toString($object), $object->getDescription(), $url);
     }
 
     protected function configureListFields(ListMapper $list): void
--- a/src/Command/AddMassMediaCommand.php
+++ b/src/Command/AddMassMediaCommand.php
@@ -48,6 +48,8 @@ final class AddMassMediaCommand extends Command
 
     protected function configure(): void
     {
+        \assert(null !== static::$defaultDescription);
+
         $this
             ->setDescription(static::$defaultDescription)
             ->addOption('file', null, InputOption::VALUE_OPTIONAL, 'The file to parse')
@@ -61,8 +63,11 @@ final class AddMassMediaCommand extends Command
         $fp = $this->getFilePointer($input, $output);
 
         $delimiter = $input->getOption('delimiter');
+        \assert(null !== $delimiter);
         $enclosure = $input->getOption('enclosure');
+        \assert(null !== $enclosure);
         $escape = $input->getOption('escape');
+        \assert(null !== $escape);
 
         $readHeaders = false;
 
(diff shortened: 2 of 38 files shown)


src/Admin/BaseMediaAdmin.php
```diff
@@ -155,7 +155,7 @@ public function getObjectMetadata(object $object): MetadataInterface
             $provider->getFormatName($object, MediaProviderInterface::FORMAT_ADMIN)
         );
 
-        return new Metadata($object->getName(), $object->getDescription(), $url);
+        return new Metadata((string) $object, $object->getDescription(), $url);
```
VincentLanglet, line 158: What's the purpose of this ?

src/Command/AddMassMediaCommand.php
```diff
@@ -61,8 +63,11 @@ protected function execute(InputInterface $input, OutputInterface $output): int
         $fp = $this->getFilePointer($input, $output);
 
         $delimiter = $input->getOption('delimiter');
+        \assert(null !== $delimiter);
```
VincentLanglet, line 66: Feature-request to phpstan-symfony ?

src/Command/AddMassMediaCommand.php
```diff
@@ -48,6 +48,8 @@ public function __construct(ManagerInterface $mediaManager)
 
     protected function configure(): void
     {
+        \assert(null !== static::$defaultDescription);
```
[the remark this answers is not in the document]
jordisala1991: I think it is declared on symfony side as nullable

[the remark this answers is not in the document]
VincentLanglet: I don't think it can be null since you set it, never set it again, the property is protected and the class is final ?
But this might be too complicated for phpstan for a feature request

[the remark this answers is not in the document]
jordisala1991: https://github.com/symfony/console/blob/5.3/Command/Command.php#L43-L46

src/Admin/BaseMediaAdmin.php
jordisala1991: Not 100% sure, but on admin this is already defined. Looks like the media adds a description and the image of the media. 

https://github.com/sonata-project/SonataAdminBundle/blob/master/src/Admin/AbstractAdmin.php#L1703-L1706

src/Command/AddMassMediaCommand.php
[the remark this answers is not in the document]
jordisala1991: Not sure it will ever be just a string, because there are multiple ways to set the description (Attributes on php8, tags on the dependency injection, the static property and just calling setDescription)

src/Command/AddMassMediaCommand.php
jordisala1991: Not really, this is a problem with the configure option. If you want to not be able to have nullable as a value you must use VALUE_REQUIRED. 

With VALUE_OPTIONAL (even if you define a default value):

```
❯ console sonata:media:add-multiple --delimiter
$delimiter === null
```

Whit VALUE_REQUIRED:

```
❯ console sonata:media:add-multiple --delimiter
In AddMassMediaCommand.php line 114:

  Please provide a CSV file argument or CSV input stream
```

So Im changing those and the other required values to required instead of optional. Also notice that making the option required doesn't force you to set it IF it has a default value.

jordisala1991: Fixed.

src/Admin/BaseMediaAdmin.php
jordisala1991: I just changed it to use $this->toString (like the original admin implementation).

src/Provider/Pool.php
```diff
@@ -127,16 +127,18 @@ public function hasContext(string $name): bool
...
+     *
      * @phpstan-return array{
      *     providers: string[],
      *     formats: array<string, FormatOptions>,
      *     download: array<string, mixed>
-     * }|null
+     * }
      */
-    public function getContext(string $name): ?array
+    public function getContext(string $name): array
```
jordisala1991, line 138: This is the most notable change, I tried to always return an array with the context here. More like we did on the admin bundle. you have a hasContext method, and if you want to be sure you don't run into exception, you need to call it before getContext.

src/Provider/Pool.php
```diff
@@ -157,31 +159,19 @@ public function getContexts(): array
     }
 
     /**
-     * @return string[]|null
+     * @return string[]
      */
-    public function getProviderNamesByContext(string $name): ?array
+    public function getProviderNamesByContext(string $name): array
```
jordisala1991, line 164: Here I assume to the user calls the hasContext previously (same for other methods). wdyt @VincentLanglet ?

src/Provider/Pool.php
VincentLanglet: It could be nice to add the deprecation in 3.x

src/Provider/Pool.php
VincentLanglet: good idea

VincentLanglet (approved):

src/Provider/Pool.php
jordisala1991: To not block this PR I have added this issue: https://github.com/sonata-project/SonataMediaBundle/issues/2014

PR #2016, opened 2021-07-28
Increase psalm to level 4

jordisala1991:
## Subject

I am targeting this branch, because this can only be applied to master.

### DIFF
--- a/src/Metadata/AmazonMetadataBuilder.php
+++ b/src/Metadata/AmazonMetadataBuilder.php
@@ -40,7 +40,7 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
     public const STORAGE_GLACIER = 'GLACIER';
 
     /**
-     * @var array
+     * @var array<string, mixed>
      *
      * @phpstan-var AmazonSettings
      */
@@ -52,7 +52,7 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
     private $mimeTypes;
 
     /**
-     * @var array
+     * @var array<string, string>
      *
      * @phpstan-var array{
      *     'private': self::PRIVATE_ACCESS,
@@ -73,7 +73,7 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
     ];
 
     /**
-     * @var array
+     * @var array<string, string>
      *
      * @phpstan-var array{
      *     'standard': self::STORAGE_STANDARD,
@@ -86,6 +86,8 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
     ];
 
     /**
+     * @param array<string, mixed> $settings
+     *
      * @phpstan-param AmazonSettings $settings
      */
     public function __construct(array $settings, ?MimeTypesInterface $mimeTypes = null)
@@ -104,6 +106,8 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
 
     /**
      * Get data passed from the config.
+     *
+     * @return array<string, array|string>
      *
      * @phpstan-return array{
      *     ACL: self::PRIVATE_ACCESS|self::PUBLIC_READ|self::PUBLIC_READ_WRITE|self::AUTHENTICATED_READ|self::BUCKET_OWNER_READ|self::BUCKET_OWNER_FULL_CONTROL,
@@ -129,7 +133,7 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
      *
      * @param string $filename path to the file inside the S3 bucket
      *
-     * @return array
+     * @return array<string, string>
      *
      * @phpstan-return array{contentType: string}
      */
--- a/src/Metadata/AmazonMetadataBuilder.php
+++ b/src/Metadata/AmazonMetadataBuilder.php
@@ -17,6 +17,15 @@ use Sonata\MediaBundle\Model\MediaInterface;
 use Symfony\Component\Mime\MimeTypes;
 use Symfony\Component\Mime\MimeTypesInterface;
 
+/**
+ * @phpstan-type AmazonSettings = array{
+ *     acl: 'private'|'public'|'open'|'auth_read'|'owner_read'|'owner_full_control',
+ *     storage: 'standard'|'reduced',
+ *     encryption: 'aes256',
+ *     meta: array<string, mixed>,
+ *     cache_control: string
+ * }
+ */
 final class AmazonMetadataBuilder implements MetadataBuilderInterface
 {
     public const PRIVATE_ACCESS = 'private';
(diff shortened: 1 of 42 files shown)


src/Metadata/AmazonMetadataBuilder.php
```diff
@@ -31,12 +40,28 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
     public const STORAGE_GLACIER = 'GLACIER';
 
     /**
-     * @var array<string, mixed>
+     * @var array
```
VincentLanglet, line 43: you could keep array<string,  mixed>

src/Metadata/AmazonMetadataBuilder.php
```diff
@@ -31,12 +40,28 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
...
     /**
-     * @var string[]
+     * @var MimeTypesInterface
+     */
+    private $mimeTypes;
+
+    /**
+     * @var array
+     *
+     * @phpstan-var array{
```
VincentLanglet, line 57: What's the purpose of the private property if you can't change the values ?

src/Metadata/AmazonMetadataBuilder.php
```diff
@@ -48,12 +73,20 @@ final class AmazonMetadataBuilder implements MetadataBuilderInterface
...
-     * @var MimeTypesInterface
+     * @var array
+     *
+     * @phpstan-var array{
+     *     'standard': self::STORAGE_STANDARD,
+     *     'reduced': self::STORAGE_REDUCED,
+     * }
      */
-    private $mimeTypes;
+    private $storage = [
```
VincentLanglet, line 83: What's the purpose of the private property if you can't change the values ?

src/Provider/MediaProviderInterface.php
```diff
@@ -50,7 +50,7 @@ public function addFormat(string $name, array $settings): void;
     /**
      * return the format settings.
      *
-     * @return array<string, mixed>|false the format settings
+     * @return array|false the format settings
```
VincentLanglet, line 53: you could have kept the precision

src/Metadata/AmazonMetadataBuilder.php
jordisala1991: Fixed all

jordisala1991:
The only thing I dont like about having at the same time both the phpstan-var and var (other apply the same) is because you don't get validation on the var, so you could end up doing something like:

```
    /**
     * @return array<int, MediaInterface|string|null>
     *
     * @phpstan-return array{
     *     title: string|null,
     *     caption: string|null,
     *     type: string,
     *     media: MediaInterface
     * }[]
     */
```

It is wrong, because return is telling array<int, but it is clearly an array<string. But I readded all of those because of consistency between projects.

src/Metadata/AmazonMetadataBuilder.php
jordisala1991: You suggest to move it to the method as a normal variable?

VincentLanglet: If the value can be changed, the phpdoc is wrong.

If the value cannot be changed, then using this in a private var is kinda useless. We can either using it directly in the method or using a constant for this array (instead of a private var).

VincentLanglet:
> The only thing I dont like about having at the same time both the phpstan-var and var (other apply the same) is because you don't get validation on the var, so you could end up doing something like:
> 
> ```
>     /**
>      * @return array<int, MediaInterface|string|null>
>      *
>      * @phpstan-return array{
>      *     title: string|null,
>      *     caption: string|null,
>      *     type: string,
>      *     media: MediaInterface
>      * }[]
>      */
> ```
> 
> It is wrong, because return is telling array<int, but it is clearly an array<string. But I readded all of those because of consistency between projects.

I see. Prior to yesterday, this was done because PHPStorm wasn't understanding phpstan- annotation, and some syntax like `array{}`. Now, since the 2021.2 version, PHPStorm understand phpstan annotation, generics and most of the one-line array shape signature.
So we will be able to move the phpstan- annotation to classic annotation. (but will be better in others PR).

src/Metadata/AmazonMetadataBuilder.php
jordisala1991: I will move it to constant

jordisala1991:
> > The only thing I dont like about having at the same time both the phpstan-var and var (other apply the same) is because you don't get validation on the var, so you could end up doing something like:
> > ```
> >     /**
> >      * @return array<int, MediaInterface|string|null>
> >      *
> >      * @phpstan-return array{
> >      *     title: string|null,
> >      *     caption: string|null,
> >      *     type: string,
> >      *     media: MediaInterface
> >      * }[]
> >      */
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It is wrong, because return is telling array<int, but it is clearly an array<string. But I readded all of those because of consistency between projects.
> 
> I see. Prior to yesterday, this was done because PHPStorm wasn't understanding phpstan- annotation, and some syntax like `array{}`. Now, since the 2021.2 version, PHPStorm understand phpstan annotation, generics and most of the one-line array shape signature.
> So we will be able to move the phpstan- annotation to classic annotation. (but will be better in others PR).

Yep, we can do that change in other PR. Also there are another tools like vscode that does not understand that syntax. (maybe we could just support what phpstorm supports).

VincentLanglet:
> Yep, we can do that change in other PR. Also there are another tools like vscode that does not understand that syntax. (maybe we could just support what phpstorm supports).

The "other tool" argument is not a bad argument.
If people prefer to develop with another tool, we shouldn't exclude them.

Phpstorm just started to support the new syntax, I don't know for vscode.

I think @greg0ire is using vim too ; how phpstan/psalm and new phpdoc syntax are supported atm ?

src/Metadata/AmazonMetadataBuilder.php
jordisala1991: Done

src/Metadata/AmazonMetadataBuilder.php
jordisala1991: Done

VincentLanglet (approved):

greg0ire:
I'm not sure what you mean by "support" but I don't think vim users will mind at all if you drop prefixes: there are no "inspections", instead we have https://github.com/dense-analysis/ale which delegates directly to Psalm/Phpstan to get static analysis right in the editor, with the exact same error messages as these tools.
As for auto-complete, it's based on type inference, and for that we have https://nathansmith.io/posts/neovim-lsp/ , which we can plug to a language server. I started using phpactor as a language server for that purpose recently, but  it's too recent for me to know exactly what works and what doesn't.

PR #2018, opened 2021-08-07
Fix more NEXT_MAJOR comments

jordisala1991:
## Subject

I am targeting this branch, because NEXT_MAJOR comments can be removed on master.

### DIFF
--- a/src/Form/DataTransformer/ServiceProviderDataTransformer.php
+++ b/src/Form/DataTransformer/ServiceProviderDataTransformer.php
@@ -32,7 +32,6 @@ final class ServiceProviderDataTransformer implements DataTransformerInterface, 
     public function __construct(MediaProviderInterface $provider)
     {
         $this->provider = $provider;
-        $this->logger = new NullLogger();
     }
 
     public function transform($value)
@@ -49,7 +48,7 @@ final class ServiceProviderDataTransformer implements DataTransformerInterface, 
         try {
             $this->provider->transform($value);
         } catch (\Throwable $e) {
-            \assert(null !== $this->logger);
+            $this->logger = $this->logger ?? new NullLogger();
 
             // #1107 We must never throw an exception here.
             // An exception here would prevent us to provide meaningful errors through the Form
--- a/src/Form/DataTransformer/ServiceProviderDataTransformer.php
+++ b/src/Form/DataTransformer/ServiceProviderDataTransformer.php
@@ -48,12 +48,12 @@ final class ServiceProviderDataTransformer implements DataTransformerInterface, 
         try {
             $this->provider->transform($value);
         } catch (\Throwable $e) {
-            $this->logger = $this->logger ?? new NullLogger();
+            $logger = $this->logger ?? new NullLogger();
 
             // #1107 We must never throw an exception here.
             // An exception here would prevent us to provide meaningful errors through the Form
             // Error message taken from Monolog\ErrorHandler
-            $this->logger->error(
+            $logger->error(
                 sprintf('Caught Exception %s: "%s" at %s line %s', \get_class($e), $e->getMessage(), $e->getFile(), $e->getLine()),
                 ['exception' => $e]
             );
--- a/src/Resources/config/api_form_doctrine_mongodb.php
+++ b/src/Resources/config/api_form_doctrine_mongodb.php
@@ -39,7 +39,6 @@ return static function (ContainerConfigurator $containerConfigurator): void {
                 new ReferenceConfigurator('sonata.media.pool'),
                 '%sonata.media.admin.media.entity%',
             ])
-            // NEXT_MAJOR: make symfony/monolog-bundle a require dependency and remove nullOnInvalid
             ->call('setLogger', [(new ReferenceConfigurator('logger'))->nullOnInvalid()])
 
         ->set('sonata.media.api.form.type.gallery', ApiGalleryType::class)
--- a/src/Resources/translations/SonataAdminBundle.de.xliff
+++ b/src/Resources/translations/SonataAdminBundle.de.xliff
@@ -9,14 +9,6 @@
                 <source>sonata_media</source>
                 <target>Medienbibliothek</target>
             </trans-unit>
-            <trans-unit id="form.label_icon">
-                <source>form_group.gallery</source>
-                <target>form_group.gallery</target>
-            </trans-unit>
-            <trans-unit id="form.label_icon">
-                <source>form_group.options</source>
-                <target>form_group.options</target>
-            </trans-unit>
         </body>
     </file>
 </xliff>
--- a/src/Resources/translations/SonataAdminBundle.de.xliff
+++ b/src/Resources/translations/SonataAdminBundle.de.xliff
@@ -9,6 +9,14 @@
                 <source>sonata_media</source>
                 <target>Medienbibliothek</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>form_group.gallery</target>
+            </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.options</source>
+                <target>form_group.options</target>
+            </trans-unit>
         </body>
     </file>
 </xliff>
--- a/src/Resources/translations/SonataAdminBundle.en.xliff
+++ b/src/Resources/translations/SonataAdminBundle.en.xliff
@@ -9,6 +9,14 @@
                 <source>sonata_media</source>
                 <target>Media Library</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>form_group.gallery</target>
+            </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.options</source>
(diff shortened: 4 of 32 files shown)


src/Resources/config/api_form_doctrine_mongodb.php
```diff
@@ -39,7 +39,6 @@
                 new ReferenceConfigurator('sonata.media.pool'),
                 '%sonata.media.admin.media.entity%',
             ])
-            // NEXT_MAJOR: make symfony/monolog-bundle a require dependency and remove nullOnInvalid
```
jordisala1991, line 42: I don't think we should add `symfony/monolog-bundle` as a required dependency. On SonataAdminBundle we didn't do it.

src/Form/DataTransformer/ServiceProviderDataTransformer.php
```diff
@@ -49,7 +48,7 @@ public function reverseTransform($value)
         try {
             $this->provider->transform($value);
         } catch (\Throwable $e) {
-            \assert(null !== $this->logger);
+            $this->logger = $this->logger ?? new NullLogger();
```
core23, line 51: This looks strange. Is this symfony new best practice?

There's a risk that some other method might use the logger too.

core23 (changes requested):

src/Form/DataTransformer/ServiceProviderDataTransformer.php
VincentLanglet: Yes, why not keeping the old way ?

src/Resources/translations/SonataAdminBundle.en.xliff
```diff
@@ -9,6 +9,14 @@
                 <source>sonata_media</source>
                 <target>Media Library</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>form_group.gallery</target>
```
VincentLanglet, line 14: Gallery

src/Resources/translations/SonataAdminBundle.en.xliff
```diff
@@ -9,6 +9,14 @@
...
+            </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.options</source>
+                <target>form_group.options</target>
```
VincentLanglet, line 18: Options

src/Resources/translations/SonataMediaBundle.fr.xliff
```diff
@@ -571,6 +571,14 @@
                 <source>error.image_too_small</source>
                 <target>Image trop petite, la taille minimum requise est de %min_width% x %min_height%.</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>form_group.gallery</target>
```
VincentLanglet, line 576: Gallerie

src/Resources/translations/SonataMediaBundle.fr.xliff
```diff
@@ -571,6 +571,14 @@
...
+            </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.options</source>
+                <target>form_group.options</target>
```
VincentLanglet, line 580: Options

src/Resources/translations/SonataAdminBundle.de.xliff
```diff
@@ -9,6 +9,14 @@
                 <source>sonata_media</source>
                 <target>Medienbibliothek</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>form_group.gallery</target>
```
VincentLanglet, line 14: User will see the translation changing from `Gallery` to `form_group.gallery`

jordisala1991: Yep, how do you prefer to handle this, We have 2 options:

1. We add translations to all languages with the placeholder
2. We dont add the translation on all languages.

Doing 1 is good because is what is done on AdminBundle (and I believe this was already discussed)

Doing 2 is good because on other languages if you have 'en' as default. But, you will end with  a lot of files of translations without same keys, a mess.

I was following 1. and my intention was to add all the keys to all languages on 3.x.

But not strong opinion on any, just would be nice to know how to proced with translations.

VincentLanglet: There is a general discussion https://github.com/sonata-project/dev-kit/issues/751

> Doing 1 is good because is what is done on AdminBundle (and I believe this was already discussed)

I solved the issues by replacing every placeholder by the translation.

> Doing 2 is good because on other languages if you have 'en' as default. But, you will end with a lot of files of translations without same keys, a mess.

I agree, when I tried to add the translations on SonataAdmin, it was a mess to understand which was missing and which was not. Looking at symfony process https://github.com/symfony/symfony/issues/38710
- There is a website with the status of translations https://symfony-translations.nyholm.tech/
- Issues are created for this https://github.com/symfony/symfony/issues/42315
- They are using number as `id` (so it's easy to check the order)

In our case I would say
- Seing on the dev-kit site the status of the translations
- Having a CI-check about the order of the translations (every files should use the same order than the `en` file).
would be the requirements to avoid this "mess".

Since I couldn't decide between 1 and 2, I added translations in every local when I need some ^^'

OskarStark: "Galerie"

src/Resources/translations/SonataAdminBundle.de.xliff
```diff
@@ -9,6 +9,14 @@
...
+            </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.options</source>
+                <target>form_group.options</target>
```
OskarStark, line 18:
```suggestion
                <target>Optionen</target>
```

src/Resources/translations/SonataMediaBundle.es.xliff
```diff
@@ -531,6 +531,14 @@
                 <source>form.label_class</source>
                 <target>form.label_class</target>
             </trans-unit>
+            <trans-unit id="form.label_icon">
+                <source>form_group.gallery</source>
+                <target>Galería</target>
+            </trans-unit>
+            <trans-unit id="form.label_icon">
```
OskarStark, line 538: The id `form.label_icon` is used twice 🤔

src/Form/DataTransformer/ServiceProviderDataTransformer.php
jordisala1991: Fixed using local variables.

src/Resources/translations/SonataMediaBundle.fr.xliff
jordisala1991: Fixed

src/Resources/translations/SonataMediaBundle.fr.xliff
jordisala1991: Fixed.

src/Resources/translations/SonataAdminBundle.de.xliff
jordisala1991: Fixed

src/Resources/translations/SonataMediaBundle.es.xliff
jordisala1991: Fixed.

src/Resources/translations/SonataAdminBundle.de.xliff
jordisala1991: Fixed the "de" translations, but the conversation is still going.

src/Resources/translations/SonataAdminBundle.en.xliff
jordisala1991: Not touching this file, I reverted the changes, this is for SonataAdminBundle translation domain (we should probably move this to the SonataMediaBundle domain, but that will belong to another Pr.

src/Resources/translations/SonataAdminBundle.en.xliff
jordisala1991: Not touching this file, I reverted the changes, this is for SonataAdminBundle translation domain (we should probably move this to the SonataMediaBundle domain, but that will belong to another Pr.

src/Resources/translations/SonataAdminBundle.de.xliff
jordisala1991: If we are going to use Weblate, for now I will remove those placeholders, because using weblate it is better to not add fake translations, to keep a better track of the current status.

src/Form/DataTransformer/ServiceProviderDataTransformer.php
core23: > I think it's because the `LoggerAwareTrait` allow the logger to be null.

I think we should remove the trait and add a dedicated property that can't be null to fix this problem.

VincentLanglet (approved):

jordisala1991:
Rebasing this is hard than opening it again: https://github.com/sonata-project/SonataMediaBundle/pull/2033

PR #1768, opened 2020-07-17
Use ImageProviderInterface instead of ImageProvider for contructor's …

jerome-fix:
…signature

## Subject

I need to use my own ImageProvider and FileProvider for differents reasons.
Since these base providers are final classes (or declared as final) I can't extend them anymore.

Here my config : 
```
parameters:
  sonata.media.provider.image.class: App\Media\Provider\ImageProvider
  sonata.media.manager.media.class:  App\Manager\MediaManager
  sonata.media.provider.file.class: App\Media\Provider\FileProvider
```

  
Since the introduction of ImageUploadDimension validator I can't upload any new media, because the signature of the constructor expects an Sonata\MediaBundle\Provider\ImageProvider.

So I changed the constructor's signature to take an ImageProviderInterface instead. 

My provider can implement this interface as Sonata\MediaBundle\ImageProvide

 

 

## Changelog

```markdown
### Added
- Added ImageProviderInterface


### Deprecated
 - Deprecations about the "sonata.media.provider.image" service alias,use "Sonata\MediaBundle\Provider\ImageProviderInterface" instead.

```

PR #2019, opened 2021-08-08
Make classes final

### LINKED ISSUE #1935: [NEXT_MAJOR] Close API
## Feature Request
- [x] Add `final` modifier
- [x] REduce visibility to private
- [x] ADd type hints

jordisala1991:
## Subject

I am targeting this branch, because this is a step to master release.

Closes #1935 .
Closes #1768 .


src/Provider/ImageProviderInterface.php
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * @author Jordi Sala <jordism91@gmail.com>
+ *
+ * @phpstan-import-type FormatOptions from MediaProviderInterface
+ */
+interface ImageProviderInterface
+{
+    /**
+     * @return array<string, int|string|bool|array|null>[]
```
VincentLanglet, line 24:
```suggestion
     * @return array<array<string, int|string|bool|array|null>>
```

I personally dont like mixing both array<> and [].
Here, it's easy to miss it IMHO.

jordisala1991: Good point, fixed.

VincentLanglet (approved):

Issue #1824, opened 2020-09-24
Refactor `CDNInterface`


phansys:
## Feature Request
This RFC proposes to refactor the `CDNInterface` interface (and its implementations) in order to receive instances of `MediaInterface` instead of string paths.
This will allow to remove the methods `BaseProvider::getCdnPath()` and `BaseProvider::flushCdn()`. `MediaProviderInterface` interface should declare the method `getCdn(): CDNInterface`.

As a side note, note that currently, the namespace `Sonata\MediaBundle\CDN\` is completely agnostic from the rest of the project, it doesn't use any class from another namespace present in this package.

jordisala1991:
Do you have time to work on this for 4.0 milestone?

phansys:
I hope this feature or some parts of it can be addressed in the stable branch first.
I'll try to take a look.

jordisala1991:
It will depend on wheter you have time for it in the next 2 weeks. My plan is to release rc1 by start of september.

PR #2035, opened 2021-08-12
Bump PHPStan level to 2

phansys:
## Subject
Bump PHPStan level to 2.

## Changelog

```markdown
### Changed
- Several missing or wrong type hints.
```

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -9,3 +9,13 @@ parameters:
             message: "#^Call to static method getInstance\\(\\) on an unknown class Sonata\\\\EasyExtendsBundle\\\\Mapper\\\\DoctrineCollector\\.$#"
             count: 1
             path: src/DependencyInjection/SonataMediaExtension.php
+
+        -
+            message: "#^Call to an undefined method Sonata\\\\ClassificationBundle\\\\Model\\\\CategoryManagerInterface\\:\\:getRootCategories\\(\\)\\.$#"
+            count: 1
+            path: src/Model/CategoryManager.php
+
+        -
+            message: "#^Call to an undefined method Sonata\\\\ClassificationBundle\\\\Model\\\\CategoryManagerInterface\\:\\:getRootCategory\\(\\)\\.$#"
+            count: 1
+            path: src/Model/CategoryManager.php
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -2,7 +2,7 @@ includes:
     - phpstan-baseline.neon
 
 parameters:
-    level: 1
+    level: 2
     symfony:
         console_application_loader: ./phpstan-console-application.php
     scanFiles:
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -20,10 +20,15 @@ use Sonata\AdminBundle\Form\Type\ModelListType;
 use Sonata\BlockBundle\Meta\Metadata;
 use Sonata\MediaBundle\Form\DataTransformer\ProviderDataTransformer;
 use Sonata\MediaBundle\Model\CategoryManagerInterface;
+use Sonata\MediaBundle\Model\MediaInterface;
 use Sonata\MediaBundle\Provider\MediaProviderInterface;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Form\Extension\Core\Type\HiddenType;
 
+/**
+ * @phpstan-template T of MediaInterface
+ * @phpstan-extends AbstractAdmin<T>
+ */
 abstract class BaseMediaAdmin extends AbstractAdmin
 {
     /**
@@ -39,10 +44,11 @@ abstract class BaseMediaAdmin extends AbstractAdmin
     protected $classnameLabel = 'Media';
 
     /**
-     * @param string                   $code
-     * @param string                   $class
-     * @param string                   $baseControllerName
-     * @param CategoryManagerInterface $categoryManager
+     * @param string $code
+     * @param string $class
+     * @param string $baseControllerName
+     *
+     * @phpstan-param class-string<T> $class
      */
     public function __construct($code, $class, $baseControllerName, Pool $pool, ?CategoryManagerInterface $categoryManager = null)
     {
@@ -68,7 +74,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
         }
 
         $filter = $this->getRequest()->get('filter');
-        if (null !== $filter && \array_key_exists('context', $this->getRequest()->get('filter'))) {
+        if (null !== $filter && \array_key_exists('context', $filter)) {
             $context = $filter['context']['value'];
         } else {
             $context = $this->getRequest()->get('context', $this->pool->getDefaultContext());
@@ -121,7 +127,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
             if (null !== $this->categoryManager && null !== $categoryId = $this->getPersistentParameter('category')) {
                 $category = $this->categoryManager->find($categoryId);
 
-                if ($category && $category->getContext()->getId() === $context) {
+                if (null !== $category && $category->getContext()->getId() === $context) {
                     $media->setCategory($category);
                 }
             }
@@ -140,6 +146,8 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
     /**
      * @final since sonata-project/media-bundle 3.36.0
+     *
+     * @param MediaInterface $object
      */
     public function getObjectMetadata($object)
     {
@@ -164,13 +172,11 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
     protected function configureFormFields(FormMapper $form)
     {
-        $media = $this->getSubject();
+        $media = $this->hasSubject() ? $this->getSubject() : $this->getNewInstance();
+        // NEXT_MAJOR: Remove the previous line and uncomment the following one.
+        // $media = $this->getSubject();
 
-        if (!$media) {
-            $media = $this->getNewInstance();
-        }
-
-        if (!$media || !$media->getProviderName()) {
+        if (null === $media->getProviderName()) {
             return;
         }
 
@@ -180,7 +186,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
         $provider = $this->pool->getProvider($media->getProviderName());
 
-        if ($media->getId()) {
+        if (null !== $media->getId()) {
             $provider->buildEditForm($form);
         } else {
             $provider->buildCreateForm($form);
--- a/src/Command/AddMassMediaCommand.php
+++ b/src/Command/AddMassMediaCommand.php
@@ -14,6 +14,7 @@ declare(strict_types=1);
 namespace Sonata\MediaBundle\Command;
 
 use Doctrine\Persistence\ManagerRegistry;
+use Sonata\Doctrine\Model\ClearableManagerInterface;
 use Sonata\Doctrine\Model\ManagerInterface;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Console\Input\InputInterface;
@@ -120,7 +121,11 @@ class AddMassMediaCommand extends BaseCommand
     protected function optimize()
     {
         if (null !== $this->managerRegistry) {
-            $this->managerRegistry->getManager()->getUnitOfWork()->clear();
+            $manager = $this->managerRegistry->getManager();
+
+            if ($manager instanceof ClearableManagerInterface) {
+                $manager->clear();
+            }
         }
     }
 }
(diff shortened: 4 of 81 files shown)


phansys:
>If you up phpstan levels or add phpdoc please use master for reference and help with the merge after the Pr gets accepted. It could cause multiple conflicts

I need these changes in the 3.x branch.
Have we updated our policies about this?

jordisala1991:
> >If you up phpstan levels or add phpdoc please use master for reference and help with the merge after the Pr gets accepted. It could cause multiple conflicts
> 
> 
> 
> I need these changes in the 3.x branch.
> 
> Have we updated our policies about this?

You can do them on 3.x, nothing has changed.
What Im telling is that those changes are already on master probably, take a look and backport what you need, to avoid unnecesary conflicts.

phansys:
Thank you @jordisala1991, When this PR is merged, I'll try to perform the merge against `master`.

src/Admin/BaseMediaAdmin.php
```diff
@@ -53,6 +60,11 @@ public function __construct($code, $class, $baseControllerName, Pool $pool, ?Cat
         $this->categoryManager = $categoryManager;
     }
 
+    /**
+     * @param MediaInterface $object
+     *
+     * @phpstan-param T $object
+     */
```
VincentLanglet, line 67: Why is this needed, it should be in the parent interface, isn't it ?

src/Admin/BaseMediaAdmin.php
```diff
@@ -103,6 +115,11 @@ public function getPersistentParameters()
         ]);
     }
 
+    /**
+     * @return MediaInterface
+     *
+     * @phpstan-return T
+     */
```
VincentLanglet, line 122: same

src/Admin/BaseMediaAdmin.php
```diff
@@ -138,6 +155,11 @@ public function getPool()
         return $this->pool;
     }
 
+    /**
+     * @param MediaInterface $object
+     *
+     * @phpstan-param T $object
+     */
```
VincentLanglet, line 162: same

src/Admin/BaseMediaAdmin.php
```diff
@@ -161,13 +183,9 @@ protected function configureListFields(ListMapper $list)
 
     protected function configureFormFields(FormMapper $form)
     {
-        $media = $this->getSubject();
-
-        if (!$media) {
-            $media = $this->getNewInstance();
-        }
+        $media = $this->hasSubject() ? $this->getSubject() : $this->getNewInstance();
```
VincentLanglet, line 186: I'm not sure it makes a lot of sens since SonataAdmin have a check to throw an error if there is no subject when building the form/show.

src/Command/AddMassMediaCommand.php
```diff
@@ -120,7 +122,10 @@ protected function insertMedia(array $data, OutputInterface $output)
     protected function optimize()
     {
         if (null !== $this->managerRegistry) {
-            $this->managerRegistry->getManager()->getUnitOfWork()->clear();
+            $manager = $this->managerRegistry->getManager();
+
+            \assert($manager instanceof DocumentManagerInterface || $manager instanceof EntityManagerInterface);
+            $manager->getUnitOfWork()->clear();
```
VincentLanglet, line 128: I think we introduce a ClearableManagerInterface, couldn't we do an instanceof check instead ?

https://github.com/sonata-project/sonata-doctrine-extensions/pull/323/files

src/Command/BaseCommand.php
```diff
@@ -29,7 +30,7 @@
 abstract class BaseCommand extends Command
 {
     /**
-     * @var ManagerInterface
+     * @var ManagerInterface&ClearableManagerInterface
```
VincentLanglet, line 33: I'm not sure the `ClearableManagerInterface` is required.

When using `clear, we should just make an `instanceof` check before calling clear.
This is what we did on master.

src/Command/UpdateCdnStatusCommand.php
```diff
@@ -84,6 +85,9 @@ public function execute(InputInterface $input, OutputInterface $output)
         $this->output = $output;
 
         $provider = $this->getProvider();
+
+        \assert($provider instanceof BaseProvider);
```
VincentLanglet, line 89: What the purpose of this assert ?

I prefer to understand because it often show an issue
- with our return type
- with the use of the return values
- with our interfaces

phansys: I need to address #1824 in order to fix this comment.

src/Admin/BaseMediaAdmin.php
phansys: I agree, but if we throw an exception in this case, it could be considered a BC-break, since it isn't the current behavior.

VincentLanglet: Since there is the exception for 4.0 https://github.com/sonata-project/SonataAdminBundle/blob/4.x/src/Admin/AbstractAdmin.php#L896 and a deprecation for 3.x, I would not add the `hasSubject` check here.

And we could add a next major to explain that in 4.0, `$media = $this->getSubject()` will be enough, without the null check.

src/Command/UpdateCdnStatusCommand.php
VincentLanglet: Can we have a comment with the ref of the issue then ? (just in case)

src/Admin/BaseMediaAdmin.php
phansys: I think I don't get your point, as `$media` is being used several times in this method trusting it is always an instance of `MediaInterface`.

src/Command/UpdateCdnStatusCommand.php
phansys: I finally added the `@method` annotation at `CDNIterface`.

src/Admin/BaseMediaAdmin.php
VincentLanglet: In next major
```
$media = $this->hasSubject() ? $this->getSubject() : $this->getNewInstance();
```
can be replaced to
```
$media = $this->getSubject();
```

phansys: I agree that is the proper behavior for the next major, but I don't understand what are you suggesting here for the 3.x branch, where `$this->getSubject()` returns `null`. Could you please provide a suggestion?

VincentLanglet (commented):
Honestly I feel like we're passing a lot of time on this.
- The PR is huge, might be easier to create smaller pr fixing 5-10 files at the times.
- The conflict on the merge will be hell
- PHPStan errors should be solved the same way we did on master, we already had a lot of reflexion and discussion about the best way to do it.

BTW, by adding `@method`, the label is not pedantic anymore.

You said you "need" this for your project. I kinda curious why, and if we couldn't just fix the errors needed for your project and keep working on releasing master as fast as possible.
- Phpstan is already lvl max
- Psalm will be lvl 3: https://github.com/sonata-project/SonataMediaBundle/pull/2038
- Support for php8 was added
- Support for Symfony 5 was added
- I dunno if everything is final and type hinted, i need to check

Do you have a list of what is missing for v4 @jordisala1991 ?

src/Admin/BaseMediaAdmin.php
VincentLanglet: I just thought about adding a `Next_major` comment in order to not forget to do the change during the merge into master.

```
// NEXT_MAJOR: use the next line instead.
$media = $this->hasSubject() ? $this->getSubject() : $this->getNewInstance();
// $media = $this->getSubject();
```

phansys:
>The PR is huge, might be easier to create smaller pr fixing 5-10 files at the times.

I'll try to provide parts of these changes in multiple PRs. In fact, all the PRs I created the last 2 weeks in this project were extracted from the findings in this PR.

>The conflict on the merge will be hell

I know, I'll perform the merge when this PR is merged.

>PHPStan errors should be solved the same way we did on master, we already had a lot of reflexion and discussion about the best way to do it.
BTW, by adding @method, the label is not pedantic anymore.

I'll try to check for the changes done in the `master` branch again.

>You said you "need" this for your project. I kinda curious why, and if we couldn't just fix the errors needed for your project and keep working on releasing master as fast as possible.

The majority of errors reported by PHPStan in my projects are generated by the lack of some methods in the interfaces (which are only present in their implementations), and by covariance issues regarding the implementation of these interfaces (in some cases, from my project's classes which also extend or implement interfaces declared here).
I must check if I'm able to identify these specific cases and provide a narrower PR for them.

core23:
Can you rebase this @phansys ?

phpstan.neon.dist
```diff
@@ -2,14 +2,14 @@ includes:
...
+    level: 2
     bootstrapFiles:
         - vendor/bin/.phpunit/phpunit/vendor/autoload.php
-    symfony:
-        console_application_loader: ./phpstan-console-application.php
     scanFiles:
         # NEXT_MAJOR: Remove the scanFiles part
         - .phpstan/stubs/MetadataInterface.file
+    symfony:
+        console_application_loader: ./phpstan-console-application.php
```
ChristianGripp: Maybe move this file to `.phpstan/`

VincentLanglet: We tried to not use a .phpstan folder for others bundles

ChristianGripp: Then we should exclude the file via `.gitattributes`

VincentLanglet: https://github.com/sonata-project/dev-kit/blob/master/templates/project/.gitattributes.twig#L16

src/PHPCR/BaseGallery.php
```diff
@@ -44,10 +45,12 @@ public function getUuid()
 
     public function addGalleryHasMedia(GalleryHasMediaInterface $galleryHasMedia)
     {
+        \assert($galleryHasMedia instanceof BaseGalleryHasMedia);
```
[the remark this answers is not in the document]
phansys: I don't understand your comment here. Can you please provide a suggestion?

jordisala1991 (dismissed):
It is nearly impossible to review this PR, all of this changes and more are done already on 4.x and I can't go one by one. If you are sure about doing this, you have my approval, but you merge this one on 4.x. Keep in mind that this should produce 0 changes on 4.x, if it produces any it means that something here is wrong (most likely, because on 4.x we have psalm level 2 and phpstan level 8 checking the code)

VincentLanglet (dismissed):
I agree with @jordisala1991

phansys:
>Keep in mind that this should produce 0 changes on 4.x, if it produces any it means that something here is wrong (most likely, because on 4.x we have psalm level 2 and phpstan level 8 checking the code)

I'll try to find some time to perform this check in order to confirm the consistency.

VincentLanglet:
> I'll try to find some time to perform this check in order to confirm the consistency.

I would recommend to merge this and then merge 3.x into 4.x, the conflict should show the differences.

phansys:
>I would recommend to merge this and then merge 3.x into 4.x, the conflict should show the differences.

Sure, that check is what I have in mind.

phpstan-baseline.neon
```diff
@@ -9,3 +9,28 @@ parameters:
...
+
+        -
+            message: "#^Call to an undefined method Sonata\\\\ClassificationBundle\\\\Model\\\\CategoryManagerInterface\\:\\:getRootCategory\\(\\)\\.$#"
+            count: 1
+            path: src/Model/CategoryManager.php
+
+        -
+            message: "#^Call to an undefined method Twig\\\\Environment\\:\\:expects\\(\\)\\.$#"
+            count: 1
+            path: tests/Block/GalleryBlockServiceTest.php
```
ChristianGripp: We should fix this error here: 
https://github.com/sonata-project/SonataBlockBundle/blob/3.x/src/Test/BlockServiceTestCase.php#L50-L53

VincentLanglet (dismissed):

VincentLanglet (dismissed):

VincentLanglet:
@phansys Are you available to merge this, and do the 3.x into 4.x merge (to check).
Then we'll release the next 3.x version.

phansys:
>@phansys Are you available to merge this, and do the 3.x into 4.x merge (to check).
Then we'll release the next 3.x version.

Sorry for the delay @VincentLanglet. I'll try to find time this week.

src/Model/GalleryInterface.php
```diff
@@ -93,23 +83,29 @@ public function getCreatedAt();
     public function setDefaultFormat($defaultFormat);
```
ChristianGripp, line 93: You can add `@return void` to all setters

core23 (dismissed):

VincentLanglet (approved):

phansys:
Friendly ping @jordisala1991, @core23.
3.37.0 (#2218) will be released soon and I like to include these changes there.

jordisala1991 (approved):

core23:
There are a lot of merge conflicts. Can you try to merge this into the 4.x branch?

PR #2037, opened 2021-08-14
Add `ResizableThumbnailInterface`

### LINKED ISSUE #1861: Add interface for format resizers
## Feature Request

We are checking for the existance of a method on a class here: 

https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/DependencyInjection/Compiler/ThumbnailCompilerPass.php#L35

The only class that implement that method is:

https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Thumbnail/FormatThumbnail.php#L46

We should really add an interface with the method and check that instead (adding the corresponding BC layers).
This issue becomes more problematic when trying to make this bundle compatible with php 8, because `is_callable` is not able to check with a string class name, but needs an object.

First saw in: https://github.com/sonata-project/SonataMediaBundle/pull/1859

jordisala1991:
## Subject

Closes #1861 

## Changelog

```markdown
### Added
- Added `ResizableThumbnailInterface`
```

### DIFF
--- a/src/Thumbnail/ResizableThumbnailInterface.php
+++ b/src/Thumbnail/ResizableThumbnailInterface.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Thumbnail;
+
+use Sonata\MediaBundle\Resizer\ResizerInterface;
+
+interface ResizableThumbnailInterface
+{
+    public function addResizer(string $id, ResizerInterface $resizer): void;
+
+    /**
+     * @throws \LogicException if resizer is not found
+     */
+    public function getResizer(string $id): ResizerInterface;
+}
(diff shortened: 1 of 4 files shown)


phansys (dismissed):

src/Thumbnail/ResizableThumbnailInterface.php
```diff
@@ -0,0 +1,26 @@
...
+namespace Sonata\MediaBundle\Thumbnail;
+
+use Sonata\MediaBundle\Resizer\ResizerInterface;
+
+interface ResizableThumbnailInterface
+{
+    public function addResizer(string $id, ResizerInterface $resizer): void;
+
+    /**
+     * @throws \LogicException if resizer is not found
```
VincentLanglet, line 23: Should we provide an `hasResizer` method then ?

src/Thumbnail/ResizableThumbnailInterface.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Thumbnail;
+
+use Sonata\MediaBundle\Resizer\ResizerInterface;
+
+interface ResizableThumbnailInterface
```
VincentLanglet, line 18: We should add the interface to `3.x` in order to allow people to implement it.

src/Thumbnail/ResizableThumbnailInterface.php
jordisala1991: Added

VincentLanglet (approved):

jordisala1991:
Fixed on the merge from 3.x #2071

PR #2038, opened 2021-08-14
Increase psalm to level 3

jordisala1991:
## Subject

I am targeting this branch, because this is where we can increase it..

### DIFF
--- a/src/Generator/IdGenerator.php
+++ b/src/Generator/IdGenerator.php
@@ -38,12 +38,24 @@ final class IdGenerator implements GeneratorInterface
         $id = $media->getId();
 
         if (!is_numeric($id)) {
-            return '';
+            throw new \InvalidArgumentException(sprintf(
+                'Unable to generate path for media without numeric id using %s.',
+                self::class
+            ));
+        }
+
+        $context = $media->getContext();
+
+        if (null === $context) {
+            throw new \InvalidArgumentException(sprintf(
+                'Unable to generate path for media without context using %s.',
+                self::class
+            ));
         }
 
         $repFirstLevel = (int) ($id / $this->firstLevel);
         $repSecondLevel = (int) (($id - ($repFirstLevel * $this->firstLevel)) / $this->secondLevel);
 
-        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', $repFirstLevel + 1, $repSecondLevel + 1);
+        return sprintf('%s/%04s/%02s', $context, $repFirstLevel + 1, $repSecondLevel + 1);
     }
 }
--- a/src/Generator/IdGenerator.php
+++ b/src/Generator/IdGenerator.php
@@ -35,11 +35,15 @@ final class IdGenerator implements GeneratorInterface
 
     public function generatePath(MediaInterface $media): string
     {
-        $mediaId = (int) $media->getId();
+        $id = $media->getId();
 
-        $rep_first_level = (int) ($mediaId / $this->firstLevel);
-        $rep_second_level = (int) (($mediaId - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
+        if (!is_numeric($id)) {
+            return '';
+        }
 
-        return sprintf('%s/%04s/%02s', $media->getContext(), $rep_first_level + 1, $rep_second_level + 1);
+        $repFirstLevel = (int) ($id / $this->firstLevel);
+        $repSecondLevel = (int) (($id - ($repFirstLevel * $this->firstLevel)) / $this->secondLevel);
+
+        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', $repFirstLevel + 1, $repSecondLevel + 1);
     }
 }
--- a/src/Generator/UuidGenerator.php
+++ b/src/Generator/UuidGenerator.php
@@ -21,6 +21,6 @@ class UuidGenerator implements GeneratorInterface
     {
         $id = (string) $media->getId();
 
-        return sprintf('%s/%04s/%02s', $media->getContext(), substr($id, 0, 4), substr($id, 4, 2));
+        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', substr($id, 0, 4), substr($id, 4, 2));
     }
 }
--- a/src/Generator/UuidGenerator.php
+++ b/src/Generator/UuidGenerator.php
@@ -20,7 +20,15 @@ class UuidGenerator implements GeneratorInterface
     public function generatePath(MediaInterface $media): string
     {
         $id = (string) $media->getId();
+        $context = $media->getContext();
 
-        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', substr($id, 0, 4), substr($id, 4, 2));
+        if (null === $context) {
+            throw new \InvalidArgumentException(sprintf(
+                'Unable to generate path for media without context using %s.',
+                self::class
+            ));
+        }
+
+        return sprintf('%s/%04s/%02s', $context, substr($id, 0, 4), substr($id, 4, 2));
     }
 }
--- a/src/Provider/BaseVideoProvider.php
+++ b/src/Provider/BaseVideoProvider.php
@@ -94,20 +94,32 @@ abstract class BaseVideoProvider extends BaseProvider
 
     public function generatePublicUrl(MediaInterface $media, string $format): string
     {
+        $id = $media->getId();
+
+        if (null === $id) {
+            return '';
+        }
+
         return $this->getCdn()->getPath(sprintf(
             '%s/thumb_%s_%s.jpg',
             $this->generatePath($media),
-            $media->getId(),
+            $id,
             $format
         ), $media->getCdnIsFlushable());
     }
 
     public function generatePrivateUrl(MediaInterface $media, string $format): string
     {
+        $id = $media->getId();
+
+        if (null === $id) {
+            return '';
+        }
+
         return sprintf(
             '%s/thumb_%s_%s.jpg',
             $this->generatePath($media),
-            $media->getId(),
+            $id,
             $format
         );
     }
@@ -176,7 +188,7 @@ abstract class BaseVideoProvider extends BaseProvider
             $response = $this->sendRequest('GET', $url);
         } catch (\RuntimeException $e) {
             throw new \RuntimeException(
-                'Unable to retrieve the video information for :'.$url,
+                sprintf('Unable to retrieve the video information for: %s', $url),
                 \is_int($e->getCode()) ? $e->getCode() : 0,
                 $e
             );
@@ -185,7 +197,7 @@ abstract class BaseVideoProvider extends BaseProvider
         $metadata = json_decode($response, true);
 
         if (null === $metadata) {
-            throw new \RuntimeException('Unable to decode the video information for :'.$url);
+            throw new \RuntimeException(sprintf('Unable to decode the video information for: %s', $url));
         }
 
         return $metadata;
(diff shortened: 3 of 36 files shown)


src/Generator/IdGenerator.php
```diff
@@ -40,6 +40,6 @@ public function generatePath(MediaInterface $media): string
         $rep_first_level = (int) ($mediaId / $this->firstLevel);
         $rep_second_level = (int) (($mediaId - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
 
-        return sprintf('%s/%04s/%02s', $media->getContext(), $rep_first_level + 1, $rep_second_level + 1);
+        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', $rep_first_level + 1, $rep_second_level + 1);
```
jordisala1991, line 43: For other cases like throwing exceptions or logging things, I don't think it matters that much if we do the ?? '' trick. But for this cases we might do an if before and throw an exception if it is null?

cc @VincentLanglet @phansys

VincentLanglet: I don't know what's this generatePath method was used for.

For a media without context, it was casted to `''` previously. If this can't work like this it's better indeed to add an exception "Cannot generate path for media without context".

src/Provider/FileProvider.php
```diff
@@ -257,9 +257,11 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
             throw new \RuntimeException(sprintf('Cannot use X-Sendfile or X-Accel-Redirect with non %s.', Local::class));
         }
 
+        $directory = $adapter->getDirectory();
+
         $filename = sprintf(
             '%s/%s',
-            $adapter->getDirectory(),
+            false !== $directory ? $directory : '',
```
VincentLanglet, line 264: Should we do this or throw an exception ?

src/Provider/DailyMotionProvider.php
```diff
@@ -120,7 +120,7 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
 
     public function getReferenceUrl(MediaInterface $media): string
     {
-        return sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference());
+        return sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference() ?? '');
```
VincentLanglet, line 123: Should we do this or an exception ?

src/Provider/FileProvider.php
```diff
@@ -83,7 +83,7 @@ public function getReferenceImage(MediaInterface $media): string
         return sprintf(
             '%s/%s',
             $this->generatePath($media),
-            $media->getProviderReference()
+            $media->getProviderReference() ?? ''
```
VincentLanglet, line 86: same

src/Provider/FileProvider.php
```diff
@@ -378,7 +380,12 @@ protected function doTransform(MediaInterface $media): void
      */
     protected function setFileContents(MediaInterface $media, ?string $contents = null): void
     {
-        $file = $this->getFilesystem()->get(sprintf('%s/%s', $this->generatePath($media), $media->getProviderReference()), true);
+        $file = $this->getFilesystem()->get(sprintf(
+            '%s/%s',
+            $this->generatePath($media),
+            $media->getProviderReference() ?? ''
```
VincentLanglet, line 386: same

src/Provider/FileProvider.php
jordisala1991: I think this is the same case, we should throw an exception. I will change those cases.

jordisala1991: Done.

src/Generator/IdGenerator.php
jordisala1991: Lets start this review again. I will resolve all your conversations and create a new ones, so we don't miss anything.

src/Generator/IdGenerator.php
```diff
@@ -35,11 +35,15 @@ public function __construct(int $firstLevel = 100000, int $secondLevel = 1000)
 
     public function generatePath(MediaInterface $media): string
     {
-        $mediaId = (int) $media->getId();
+        $id = $media->getId();
 
-        $rep_first_level = (int) ($mediaId / $this->firstLevel);
-        $rep_second_level = (int) (($mediaId - ($rep_first_level * $this->firstLevel)) / $this->secondLevel);
+        if (!is_numeric($id)) {
+            return '';
```
jordisala1991, line 41: Does it make sense to generate a Exception here? Maybe yes, because it wouldn't make sense to generate a path without the id of the media right?

src/Generator/IdGenerator.php
```diff
@@ -35,11 +35,15 @@ public function __construct(int $firstLevel = 100000, int $secondLevel = 1000)
...
+        }
 
-        return sprintf('%s/%04s/%02s', $media->getContext(), $rep_first_level + 1, $rep_second_level + 1);
+        $repFirstLevel = (int) ($id / $this->firstLevel);
+        $repSecondLevel = (int) (($id - ($repFirstLevel * $this->firstLevel)) / $this->secondLevel);
+
+        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', $repFirstLevel + 1, $repSecondLevel + 1);
```
jordisala1991, line 47: Not sure if we should throw an exception here. What if I dont have `$context` but the other information is there. Should we generate a path without the context or not?

src/Generator/UuidGenerator.php
```diff
@@ -21,6 +21,6 @@ public function generatePath(MediaInterface $media): string
     {
         $id = (string) $media->getId();
 
-        return sprintf('%s/%04s/%02s', $media->getContext(), substr($id, 0, 4), substr($id, 4, 2));
+        return sprintf('%s/%04s/%02s', $media->getContext() ?? '', substr($id, 0, 4), substr($id, 4, 2));
```
jordisala1991, line 24: Same thing about context here.

src/Provider/BaseVideoProvider.php
```diff
@@ -94,20 +94,32 @@ public function getReferenceFile(MediaInterface $media): File
 
     public function generatePublicUrl(MediaInterface $media, string $format): string
     {
+        $id = $media->getId();
+
+        if (null === $id) {
+            return '';
```
jordisala1991, line 100: Same thing about ids here.

src/Provider/BaseVideoProvider.php
```diff
@@ -94,20 +94,32 @@ public function getReferenceFile(MediaInterface $media): File
...
             $format
         ), $media->getCdnIsFlushable());
     }
 
     public function generatePrivateUrl(MediaInterface $media, string $format): string
     {
+        $id = $media->getId();
+
+        if (null === $id) {
+            return '';
```
jordisala1991, line 116: Same thing about ids here.

src/Provider/FileProvider.php
```diff
@@ -80,11 +80,13 @@ public function getProviderMetadata(): MetadataInterface
...
     {
-        return sprintf(
-            '%s/%s',
-            $this->generatePath($media),
-            $media->getProviderReference()
-        );
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            return '';
```
jordisala1991, line 86: Same thing about reference to videos here.

src/Provider/FileProvider.php
```diff
@@ -257,13 +259,17 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
...
 
-        $filename = sprintf(
-            '%s/%s',
-            $adapter->getDirectory(),
-            $this->generatePrivateUrl($media, $format)
-        );
+        $directory = $adapter->getDirectory();
+
+        if (false === $directory) {
+            throw new \RuntimeException('Cannot retrieve directory from the adapter.');
```
jordisala1991, line 265: I did throw an exception here, it is correct?

src/Provider/FileProvider.php
```diff
@@ -378,7 +385,20 @@ protected function doTransform(MediaInterface $media): void
      */
     protected function setFileContents(MediaInterface $media, ?string $contents = null): void
     {
-        $file = $this->getFilesystem()->get(sprintf('%s/%s', $this->generatePath($media), $media->getProviderReference()), true);
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            throw new \RuntimeException(sprintf(
```
jordisala1991, line 391: I also throw an exception here, it should be consistent with the other providerReference cases.

src/Provider/VimeoProvider.php
```diff
@@ -119,7 +119,13 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
 
     public function getReferenceUrl(MediaInterface $media): string
     {
-        return sprintf('https://vimeo.com/%s', $media->getProviderReference());
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            return '';
```
jordisala1991, line 125: Same thing about references to videos.

src/Provider/YouTubeProvider.php
```diff
@@ -232,7 +232,13 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
 
     public function getReferenceUrl(MediaInterface $media): string
     {
-        return sprintf('https://www.youtube.com/watch?v=%s', $media->getProviderReference());
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            return '';
```
jordisala1991, line 238: Same thing about references to videos.

tests/App/Provider/TestProvider.php
```diff
@@ -60,11 +60,13 @@ public function getAbsolutePath(MediaInterface $media): void
...
         // A copy of the code from \Sonata\MediaBundle\Provider\FileProvider::getReferenceImage()
-        $this->prevReferenceImage = sprintf(
-            '%s/%s',
-            $this->generatePath($media),
-            $media->getProviderReference()
-        );
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            return '';
```
jordisala1991, line 66: Same thing about references to videos.

jordisala1991:
@VincentLanglet Ignore all the messages starting with `Same thing`, I added to not forge any case, but there are 3 main cases:

* Missing id to generate some path
* Missing context to generate some path
* Missing providerReference to generate some path

The Id and providerReference seems logic to throw an exception if they are not there. Not sure about the context tho, it does not seems necessary to generate the path.

But the question is, are those method generating paths safe to change to throw exception or it could be a problem? Should I add the throws to the interfaces then?

src/Generator/IdGenerator.php
VincentLanglet: It make sens to me to generate something like an InvalidArgument.

src/Generator/IdGenerator.php
VincentLanglet: I'm not sure if the path works without context here, so it may sens to throw an exception indeed

src/Provider/DailyMotionProvider.php
```diff
@@ -120,7 +120,13 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
 
     public function getReferenceUrl(MediaInterface $media): string
     {
-        return sprintf('http://www.dailymotion.com/video/%s', $media->getProviderReference());
+        $providerReference = $media->getProviderReference();
+
+        if (null === $providerReference) {
+            return '';
```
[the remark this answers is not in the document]
VincentLanglet: I would say it does not, how can I expect a reference url, without reference.
If it does not, an InvalidArgument exception seems better indeed.

src/Provider/FileProvider.php
VincentLanglet: Seems right to me, that's the strategy I generally use

src/Provider/FileProvider.php
VincentLanglet: I would consider this as an InvalidArgument Exception

VincentLanglet:
> But the question is, are those method generating paths safe to change to throw exception or it could be a problem? 

Dunno, "we'll see".
I would say it's better to check there is an id/provider/... before calling for method instead of having to check the value return.
The value will always be valid.

> Should I add the throws to the interfaces then?

Yes.

jordisala1991:
@VincentLanglet I applied the changes from return ''; to throw an Exception.

VincentLanglet (approved):

PR #2041, opened 2021-08-14
Avoid using pager if not needed

jordisala1991:
## Subject

### DIFF
--- a/src/Block/GalleryListBlockService.php
+++ b/src/Block/GalleryListBlockService.php
@@ -125,7 +125,7 @@ final class GalleryListBlockService extends AbstractBlockService implements Edit
         $order = $blockContext->getSetting('order');
         \assert(\is_string($order));
         $sortDirection = $blockContext->getSetting('sort');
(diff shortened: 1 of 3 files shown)


VincentLanglet (dismissed):
Just a phpstan error and test need to be updated

src/Block/GalleryListBlockService.php
```diff
@@ -125,7 +125,7 @@ public function execute(BlockContextInterface $blockContext, ?Response $response
         $order = $blockContext->getSetting('order');
         \assert(\is_string($order));
         $sortDirection = $blockContext->getSetting('sort');
-        \assert(\is_string($sortDirection));
+        \assert(\is_string($sortDirection) && \in_array($sortDirection, ['asc', 'desc'], true));
```
VincentLanglet, line 128: is_string check might no be needed then

VincentLanglet (dismissed):

src/Block/GalleryListBlockService.php
jordisala1991: Fixed.

VincentLanglet (approved):

PR #2044, opened 2021-08-15
Fix calls to undefined methods from Gaufrette

phansys:
## Subject
Fix calls to undefined methods from Gaufrette.

These methods were replaced at knplabs/gaufrette 0.2.0. See KnpLabs/Gaufrette#98.
These issues were detected by PHPStan at #2035.

## Changelog

```markdown
### Added
- Method `Sonata\MediaBundle\Filesystem\Replicate::createStream()`;
- Implemented `Gaufrette\Adapter\StreamFactory` in `Sonata\MediaBundle\Filesystem\Replicate`.

### Fixed
- Missing implementation for `Gaufrette\Adapter\FileFactory` interface at `Sonata\MediaBundle\Filesystem\Replicate`;
- Call to undefined method `Adapter::createFileStream()`;
- Call to undefined method `Adapter::listDirectory()`.

### Deprecated
- `Sonata\MediaBundle\Filesystem\Replicate::createFileStream()` in favor of `createStream()`.
```

### DIFF
--- a/src/Filesystem/Replicate.php
+++ b/src/Filesystem/Replicate.php
@@ -14,14 +14,17 @@ declare(strict_types=1);
 namespace Sonata\MediaBundle\Filesystem;
 
 use Gaufrette\Adapter as AdapterInterface;
+use Gaufrette\Adapter\FileFactory;
 use Gaufrette\Adapter\MetadataSupporter;
+use Gaufrette\Adapter\StreamFactory;
 use Gaufrette\Filesystem;
+use Gaufrette\Stream;
 use Psr\Log\LoggerInterface;
 
 /**
  * @final since sonata-project/media-bundle 3.21.0
  */
-class Replicate implements AdapterInterface, MetadataSupporter
+class Replicate implements AdapterInterface, FileFactory, MetadataSupporter, StreamFactory
 {
     /**
      * @var AdapterInterface
@@ -39,7 +42,7 @@ class Replicate implements AdapterInterface, MetadataSupporter
     /**
      * @var LoggerInterface
      *
-     * NEXT_MAJOR change visibiity to private
+     * NEXT_MAJOR change visibility to private
      */
     protected $logger;
 
@@ -53,12 +56,9 @@ class Replicate implements AdapterInterface, MetadataSupporter
      */
     private $secondary;
 
-    /**
-     * @param LoggerInterface $logger
-     */
     public function __construct(AdapterInterface $primary, AdapterInterface $secondary, ?LoggerInterface $logger = null)
     {
-        // NEXT_MAJOR: remove master nad slave
+        // NEXT_MAJOR: remove master and slave.
         $this->master = $primary;
         $this->slave = $secondary;
         $this->primary = $primary;
@@ -192,7 +192,9 @@ class Replicate implements AdapterInterface, MetadataSupporter
     {
         if ($this->primary instanceof MetadataSupporter) {
             return $this->primary->getMetadata($key);
-        } elseif ($this->secondary instanceof MetadataSupporter) {
+        }
+
+        if ($this->secondary instanceof MetadataSupporter) {
             return $this->secondary->getMetadata($key);
         }
 
@@ -203,6 +205,8 @@ class Replicate implements AdapterInterface, MetadataSupporter
      * Gets the class names as an array for both adapters.
      *
      * @return string[]
+     *
+     * @phpstan-return class-string<AdapterInterface>[]
      */
     public function getAdapterClassNames()
     {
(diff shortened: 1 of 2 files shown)


jordisala1991 (dismissed):
I remember to change something like that on master too, to make phpstan happy. Can you check there too?

src/Filesystem/Replicate.php
```diff
@@ -214,16 +214,48 @@ public function getAdapterClassNames()
 
     public function createFile($key, Filesystem $filesystem)
     {
+        \assert($this->primary instanceof FileFactory);
```
VincentLanglet, line 217: Why this assert is needed ? Shouldnt we change some type hint or throw an exception instead ?

src/Filesystem/Replicate.php
```diff
@@ -214,16 +214,48 @@ public function getAdapterClassNames()
...
+
         return $this->primary->createFile($key, $filesystem);
     }
 
+    public function createStream(string $key, Filesystem $filesystem): Stream
+    {
+        \assert($this->primary instanceof StreamFactory);
```
VincentLanglet, line 224: Same

src/Filesystem/Replicate.php
jordisala1991: I think I solved on master doing an IF instead.

phansys: I backported the `if` from `master` only for the primary adapter, which is the current behavior.

src/Filesystem/Replicate.php
```diff
@@ -214,16 +216,52 @@ public function getAdapterClassNames()
...
+        throw new \LogicException(sprintf('Primary adapter MUST implement %s.', FileFactory::class));
+    }
+
+    public function createStream(string $key, Filesystem $filesystem): Stream
+    {
+        if ($this->primary instanceof StreamFactory) {
+            return $this->primary->createStream($key, $filesystem);
+        }
+
+        throw new \LogicException(sprintf('Primary adapter MUST implement %s.', FileFactory::class));
```
jordisala1991, line 232: shouldn't you check for secondary too? if not, this change should go to master to when merging.

VincentLanglet (approved):

src/Filesystem/Replicate.php
phansys: As I'm not using this feature, I don't know if this change respects BC.
Are you aware why only the primary is currently used?

jordisala1991: I dont know either, thats why we changed it on master.

phansys: Even if the argument doesn't sound very solid, in front of the lack of a better answer, I adhered the suggestion.

src/Filesystem/Replicate.php
```diff
@@ -214,16 +218,65 @@ public function getAdapterClassNames()
...
+        @trigger_error(sprintf(
+            'Method "%s()" is deprecated since sonata-project/media-bundle 3.x and will be removed'
+            .' in version 4.0. Use "createStream()" instead.',
+            __METHOD__
+        ), \E_USER_DEPRECATED);
+
+        return $this->createStream($key);
     }
 
     public function listDirectory($directory = '')
```
phansys, line 270: Should we deprecate this method? It is removed in the `master` branch.
It seems to be present only at [`Gaufrette\Adapter\Ftp`](
https://github.com/KnpLabs/Gaufrette/blob/786247eba04d4693e88a80ca9fdabb634675dcac/src/Gaufrette/Adapter/Ftp.php#L230).

VincentLanglet (approved):

phansys:
I'm surprised to see all these findings addressed in the `master` branch, but not in `3.x`, even when they are also affecting the stable branch.
@sonata-project/contributors, have we changed our policies about situations like these or was it a mistake?

VincentLanglet:
> I'm surprised to see all these findings addressed in the `master` branch, but not in `3.x`, even when they are also affecting the stable branch.
> @sonata-project/contributors, have we changed our policies about situations like these or was it a mistake?

When deprecation or new interface were introduced, I asked @jordisala1991 to add them back to 3.x in order to help the developer to upgrade from 3.x to 4.x. But we could have missed some for sure.

Here, lot of findings were phpstan/psalm related.
When bumping static analysis (like I did on AdminBundle), a lot of potential issues were discovered but
- Some of them require BC-break
- Some of them are not real bug

Backporting those fixes to 3.x is not always that easy, and can create a lot of conflict during merge 3.x into master.
When working on SonataAdminBundle 4, I had dozen of merge with 100+ files with conflict. It takes a lot of time and slow down the new major. So I won't say nothing here since
- The goal is to release the version 4 stable as soon as possible
- No issue were opened to report some of those potential bug

Honestly, if I would have followed at 100% the policy, SonataAdmin 4 would have not even an alpha atm.

phansys:
I agree with all of your statements, but I'd like to be clear about one point: releasing the version 4 is one of the biggest goals we have for long time, but not the only one. IMO, we should try to provide all the fixes in the lowest maintained branch when possible. Believe me, I know how painful it can be, but this is a key concept around our policies.

jordisala1991:
We try our best, but it is not always enough. 

Priorities can also be different between different sonata members. My priority is to not see Sonata die because we are stuck on Sf 4/php 7. Tbh I dont even care about new features just want things stable and updated.

If we do some mistakes, other Sonata members can help and fix them too (like you are doing, and I appreciate it), thats why we are a team.

phansys:
https://github.com/sonata-project/SonataMediaBundle/pull/2044#issuecomment-899106251 :muscle:

src/Filesystem/Replicate.php
jordisala1991: Maybe we can do the same Logic as the others, if the Ftp is implemented call
it, otherwise exception.

I took the decision of removing it because I didnt found it

phansys: >Maybe we can do the same Logic as the others, if the Ftp is implemented call
it, otherwise exception.

That was the first solution I tried, but as unlike the other cases, `Ftp` is not an interface, I chose the `method_exists()` approach.
Additionally, this method is currently returning the result of `Ftp::listDirectory()` only for the primary adapter. How should we handle the case for the secondary adapter? At first sight this method doesn't throw an exception or return other value than array.
WDYT?

jordisala1991 (approved):

jordisala1991:
Thank you @phansys

PR #2082, opened 2021-08-22
Fix some return types

phansys:
## Subject
Fix some return types.

I am targeting this branch, because these changes try to respect BC.

See #1991.

## Changelog

```markdown
### Fixed
- Return type at `BaseMediaEventSubscriber::getProvider()`.

### Changed
- Return `null` from `BaseMediaEventSubscriber::getMedia()` if the related media does not implement `MediaInterface`.
```

### DIFF
--- a/src/Provider/FileProvider.php
+++ b/src/Provider/FileProvider.php
@@ -210,9 +210,17 @@ class FileProvider extends BaseProvider implements FileProviderInterface
 
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
+        if (self::FORMAT_REFERENCE === $format) {
             return $this->getReferenceImage($media);
         }
+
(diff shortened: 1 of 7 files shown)


src/Provider/FileProvider.php
```diff
@@ -210,10 +210,18 @@ public function getHelperProperties(MediaInterface $media, $format, $options = [
 
     public function generatePrivateUrl(MediaInterface $media, $format)
     {
-        if (MediaProviderInterface::FORMAT_REFERENCE === $format) {
+        if (self::FORMAT_REFERENCE === $format) {
             return $this->getReferenceImage($media);
         }
 
+        // throw new \InvalidArgumentException(sprintf(
```
VincentLanglet, line 217: Should we trigger a deprecation then ?

phansys: I thought about this, but it will be unfixable.

VincentLanglet (approved):

PR #2083, opened 2021-08-22
Fix support for "kriswallsmith/buzz" <0.16

phansys:
## Subject
Fix support for "kriswallsmith/buzz" <0.16.

Function `Browser::sendRequest()` was introduced in "kriswallsmith/buzz" ^0.16, but we are using it relying it's always available:
https://github.com/sonata-project/SonataMediaBundle/blob/08bf82969673cd78e166a6c81408716c5587f7a1/src/Client/BuzzClient.php#L40
See https://github.com/kriswallsmith/Buzz/blob/master/CHANGELOG.md#added-6.

## Changelog

```markdown
### Fixed
- Call to undefined function `Buzz\Browser::sendRequest()` with "kriswallsmith/buzz" <0.16.
```

### DIFF
--- a/src/Client/BuzzClient.php
+++ b/src/Client/BuzzClient.php
@@ -37,6 +37,10 @@ final class BuzzClient implements ClientInterface
 
     public function sendRequest(RequestInterface $request): ResponseInterface
     {
-        return $this->browser->sendRequest($request);
+        if (method_exists($this->browser, 'sendRequest')) {
+            return $this->browser->sendRequest($request);
+        }
+
+        return $this->browser->send($request);
     }
 }


VincentLanglet (commented):
The <0.16 release is more than 2 year old.
I'm kinda concern if we should support it or just update the composer.json.

Our policy is not to support all the version of all the dependency. But mainly the recent (and supported) ones.

If we decide to keep the support, we should at least add a NEXT_MAJOR comment;

phansys:
>I'm kinda concern if we should support it or just update the composer.json.

I tried to remove the support for version 0.15, but is not possible since the version 0.16 is not compatible with PHP 8.x.
Additionally, I think we should not invest too much effort on this dependency, since the `BuzzClient` class is removed in version 4.0:
https://github.com/sonata-project/SonataMediaBundle/blob/08bf82969673cd78e166a6c81408716c5587f7a1/src/Client/BuzzClient.php#L21-L26

jordisala1991 (approved):

VincentLanglet (approved):

haivala:
Does this mean that I need media bundle release 4 to be able to upgrade to php 8 or is there another way?

In my project:
```
# composer.phar why-not php 8.0
kriswallsmith/buzz  v0.16.1  requires  php (^5.4 || ^7.0)
# composer.phar why kriswallsmith/buzz
sonata-project/media-bundle  3.32.0  requires  kriswallsmith/buzz (^0.15 || ^0.16)
```

VincentLanglet:
Yes.

Either you wait for the media bundle release 4, for which you can help.
Either you or someone else make the buzz support in the 3.x branch optionnal.

phansys:
>Does this mean that I need media bundle release 4 to be able to upgrade to php 8 or is there another way?

No, you can still use PHP 8 in 3.x version with kriswallsmith/buzz ^0.15, where the PHP constraint is more relaxed ([`"php": ">=5.3.0"`](https://github.com/kriswallsmith/Buzz/blob/d4041666c3ffb379af02a92dabe81c904b35fab8/composer.json#L15-L17)).
The support for that version is what this PR is fixing.

VincentLanglet:
Indeed, I looked at `0.15.2`, but `0.15` is "compatible": https://packagist.org/packages/kriswallsmith/buzz#v0.15

PR #2090, opened 2021-09-04
Improve generation of thumbnails for files

jordisala1991:
## Subject

## Changelog

```markdown
### Added
- Added thumbnail placeholder generator for files.
- Split thumbnail interface to control wether they are generated or not.
```


src/Thumbnail/FileThumbnail.php
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public function generate(MediaProviderInterface $provider, MediaInterface $media): void
+    {
+        // nothing to generate, as the thumbnails for files are already generated
+    }
+
+    public function delete(MediaProviderInterface $provider, MediaInterface $media, $formats = null): void
+    {
+        // feature not available
```
VincentLanglet, line 56: Should we throw an exception ?

jordisala1991: I copied another thumbnail implementation. Should I change it too?

VincentLanglet: Honestly I dunno if we should throw a method not implemented exception or doing nothing becase some manager could use the delete method every time and rely on the fact the feature is really removed on not.

I would say the best solution would be to remove the `delete` method from the interface and add a DeletableThumbnailInterface for this. This way you can remove the delete method here.

Do you want to do this here, in another PR or never ? ^^

jordisala1991: I tried to do it in this PR, wdyt?

src/Provider/BaseProvider.php
```diff
@@ -153,12 +154,16 @@ public function requireThumbnails(): bool
 
     public function generateThumbnails(MediaInterface $media): void
     {
-        $this->thumbnail->generate($this, $media);
+        if ($this->thumbnail instanceof GeneratableThumbnailInterface) {
```
VincentLanglet, line 157: I don't know enough this code to know if we have to do it this way or throw an exception instead. Do you know ?

src/Provider/BaseProvider.php
```diff
@@ -153,12 +154,16 @@ public function requireThumbnails(): bool
...
+            $this->thumbnail->generate($this, $media);
+        }
     }
 
     public function removeThumbnails(MediaInterface $media, $formats = null): void
     {
-        $this->thumbnail->delete($this, $media, $formats);
+        if ($this->thumbnail instanceof GeneratableThumbnailInterface) {
```
VincentLanglet, line 164: Same

src/Thumbnail/ConsumerThumbnail.php
```diff
@@ -60,12 +60,13 @@ public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterf
 
     public function generate(MediaProviderInterface $provider, MediaInterface $media): void
     {
-        $backend = $this->backend;
-        $id = $this->id;
+        if (!$this->thumbnail instanceof GeneratableThumbnailInterface) {
```
VincentLanglet, line 63: Same

src/Thumbnail/ConsumerThumbnail.php
```diff
@@ -80,6 +81,8 @@ public function generate(MediaProviderInterface $provider, MediaInterface $media
 
     public function delete(MediaProviderInterface $provider, MediaInterface $media, $formats = null): void
     {
-        $this->thumbnail->delete($provider, $media, $formats);
+        if ($this->thumbnail instanceof GeneratableThumbnailInterface) {
```
VincentLanglet, line 84: Same

src/Thumbnail/FileThumbnail.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function __construct(Packages $packages)
+    {
+        $this->packages = $packages;
+    }
+
+    public function generatePublicUrl(MediaProviderInterface $provider, MediaInterface $media, string $format): string
+    {
+        if (MediaProviderInterface::FORMAT_ADMIN !== $format) {
+            throw new \RuntimeException(sprintf('Unable to generate thumbnail for the given format %s.', $format));
```
VincentLanglet, line 38: Shouldn't it be an `InvalidArgumentException` or something like `Unsupported format` ?

VincentLanglet: And `InvalidArgumentException` might be the behavior of other Thumbmail instead of returning `''`, for instance in `FormatThumbnail`, when the id is null.

src/Thumbnail/FileThumbnail.php
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        return $this->packages->getUrl('bundles/sonatamedia/file.png');
+    }
+
+    public function generatePrivateUrl(MediaProviderInterface $provider, MediaInterface $media, string $format): string
+    {
+        throw new \RuntimeException('Unable to generate private thumbnail url for media files.');
```
VincentLanglet, line 46: BadMethodCallException maybe ?

I consider that if we throw a RuntimeException it must be added to the interface when it's not mandatory for LogicException. (And it's more a logic exception to call a method which is not supposed to be called)

tests/Provider/FileProviderTest.php
```diff
@@ -43,8 +43,12 @@ class FileProviderTest extends AbstractProviderTest
     public function getProvider(): MediaProviderInterface
     {
         $resizer = $this->createMock(ResizerInterface::class);
+        $thumbnail = $this->createMock(ThumbnailInterface::class);
+        $metadata = $this->createMock(MetadataBuilderInterface::class);
```
phansys, line 47: This seems a stub, at least in the context of this method.

tests/Thumbnail/FileThumbnailTest.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Sonata\MediaBundle\Tests\Security;
+
+use PHPUnit\Framework\MockObject\MockObject;
+use PHPUnit\Framework\TestCase;
+use Sonata\MediaBundle\Provider\FileProviderInterface;
+use Sonata\MediaBundle\Tests\App\Entity\Media;
+use Sonata\MediaBundle\Thumbnail\FileThumbnail;
+use Symfony\Component\Asset\Packages;
+
+class FileThumbnailTest extends TestCase
```
phansys, line 23:
```suggestion
final class FileThumbnailTest extends TestCase
```

phansys (changes requested):

src/Thumbnail/FileThumbnail.php
```diff
@@ -0,0 +1,48 @@
...
+namespace Sonata\MediaBundle\Thumbnail;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+use Sonata\MediaBundle\Provider\MediaProviderInterface;
+use Symfony\Component\Asset\Packages;
+
+/**
+ * @author Jordi Sala Morales <jordism91@gmail.com>
+ */
+final class FileThumbnail implements ThumbnailInterface
```
phansys, line 23: The namespace already includes the term "Thumbnail".
IMO, "File" is enough for this class.

VincentLanglet: I don't agree with this, or the interface could have been called `Interface` since it's in the Thumbnail namespace.
It avoids conflict with another File class and helps to give context when using the class.
The `ArrayCollection` is not called `Array` and there is lot of similar example.

src/Provider/BaseProvider.php
jordisala1991: I think it is better this way. This code could be called by subclasses or this own class, for each media saved.

src/Thumbnail/FileThumbnail.php
jordisala1991: All the code on this bundle and other sonata bundle are the same way. On this same folder for example, all other Thumbnails are called that way. Or you can check for Providers, all include the folder name again.

This is for consistency, but also I don't agree calling it simply "File", it could be misleading, it is not a file.

jordisala1991:
Applied reviews.

VincentLanglet (approved):

src/Thumbnail/FileThumbnail.php
```diff
@@ -0,0 +1,48 @@
...
+
+    public function __construct(Packages $packages)
+    {
+        $this->packages = $packages;
+    }
+
+    public function generatePublicUrl(MediaProviderInterface $provider, MediaInterface $media, string $format): string
+    {
+        if (MediaProviderInterface::FORMAT_ADMIN !== $format) {
+            throw new \InvalidArgumentException(sprintf('Unsupported format %s.', $format));
```
phansys, line 38:
```suggestion
            throw new \InvalidArgumentException(sprintf('Unsupported format "%s".', $format));
```

src/Thumbnail/GeneratableThumbnailInterface.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Sonata\MediaBundle\Thumbnail;
+
+use Sonata\MediaBundle\Model\MediaInterface;
+use Sonata\MediaBundle\Provider\MediaProviderInterface;
+
+/**
+ * @author Jordi Sala Morales <jordism91@gmail.com>
+ */
+interface GeneratableThumbnailInterface
```
phansys, line 22: I don't know how correct it can be, but I think "Generable" fits better than "Generatable" in this case.
BTW, with a too specific name like the one this interface currently has, the `delete()` method seems unrelated to its responsibilities. IMO, the case could be different if the name were `Thumbnail(Manager|Handler)Interface` or similar.

jordisala1991: Renamed to Generable. I kinda don't like super generics names because we will end up with more methods there instead of more interfaces.

I think about it this way: Thumbnails that are generated should be also deleted at some point (when you remove the media). There are the File Thumbnails, that are for files and can't be generated, so they cannot be removed either. And the last case if for Liip thumbnails, generated on demand and stored on some cache folder (no need to delete them because they will be removed on cache removal).

wdyt?

VincentLanglet (approved):

VincentLanglet (approved):

src/Thumbnail/GeneratableThumbnailInterface.php
phansys: It's fine for me. I just want to point out that this level of specificity might require a `DeletableThumbnailInterface`.
But I don't want to block this PR.

phansys (approved):

jordisala1991:
Lets go with this implementation for now. the bundle needs more changes so we are still in time to change the interface again

Issue #2106, opened 2021-09-12
Update documentation


jordisala1991:
## Documentation Request

For the 4.0 release, we should update the docs to reflect recent changes on Sonata.

jordisala1991:
One thing that we could do is to remove the php and symfony versions from the docs. wdyt @sonata-project/contributors ? I think this should be the same for all sonata bundles. The problem with having the versions on the docs is that it becomes outdated and does not reflect what is supported, but on the other hand it is nice reading the docs and figuring out what versions are supported right away.

VincentLanglet:
> One thing that we could do is to remove the php and symfony versions from the docs. wdyt @sonata-project/contributors ? I think this should be the same for all sonata bundles. The problem with having the versions on the docs is that it becomes outdated and does not reflect what is supported, but on the other hand it is nice reading the docs and figuring out what versions are supported right away.

I think it's indeed a good idea to remove it and let people rely on the composer.json

jordisala1991:
@eerison said:

link: https://docs.sonata-project.org/projects/SonataMediaBundle/en/3.x/reference/creating_a_provider_class/#register-the-class-with-the-service-container
don't need this

 - '@sonata.media.http.client'
 - '@sonata.media.metadata.proxy'

wbloszyk:
> @eerison said:
> 
> link: https://docs.sonata-project.org/projects/SonataMediaBundle/en/3.x/reference/creating_a_provider_class/#register-the-class-with-the-service-container
> don't need this
> 
> * '@sonata.media.http.client'
> * '@sonata.media.metadata.proxy'

If I remember correctly http.client is required for some provider like YouTubeProvider, but not require for local FileProvider. Mybe nice idea will be make it optional? Also in docs.

eerison:
Hey guys I guess  I was wrong, the provider in the exemple (VimeoProvider) there are those dependencies as optional, but if it's optional, in the doc should say what happen if you don't put this.

```php
 ?MetadataBuilderInterface $metadata = null,
 ?RequestFactoryInterface $requestFactory = null 
 ```
 
Other thing:

what do you think to add the others providers in the documentation? maybe just the list should be fine. because I don't see which providers exist into Sonata, I need to check in the code to see.

VincentLanglet:
> Hey guys I guess I was wrong, the provider in the exemple (VimeoProvider) there are those dependencies as optional, but if it's optional, in the doc should say what happen if you don't put this.
> 
[...]

PR #2112, opened 2021-09-13
Remove deprecated routes and fix documentation to reflect recent changes.

jordisala1991:
## Subject

I am targeting this branch, because this should be done on my last PR #2110 .

Also part of #2106

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -19,7 +19,6 @@ Optional dependencies:
 And the persistence bundle (choose one):
 
 * `SonataDoctrineOrmAdminBundle <https://sonata-project.org/bundles/doctrine-orm-admin>`_
-* `SonataDoctrinePHPCRAdminBundle <https://sonata-project.org/bundles/doctrine-phpcr-admin>`_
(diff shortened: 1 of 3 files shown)


docs/reference/installation.rst
```diff
@@ -8,7 +8,7 @@ Installation
 Prerequisites
 -------------
 
-PHP ^7.2 and Symfony ^4.4 are needed to make this bundle work, there are
+PHP ^7.3 and Symfony ^4.4 are needed to make this bundle work, there are
```
core23, line 11: IMHO we should remove the version information from out docs. Composer will solve this problem

jordisala1991: I will revert this change, in any case it could be done first on 3.x, since the version constraints are the same.

(I also agree that we could remove the versions from docs because they always get outdated).

jordisala1991: Done. I will move this conversation to the docs update issue on this bundle.

VincentLanglet (approved):

core23 (approved):

PR #2113, opened 2021-09-14
Deprecate Pixlr integration

### LINKED ISSUE #1092: Remove inline scripts
For many reasons, we should remove all scripts from the templates. We could replace them with data listerns like in the page bundle: https://github.com/sonata-project/SonataPageBundle/pull/562.
## General

Inline scripts couldn't be cached and would pollute the html output.
## Security

There is also a [security bundle](https://github.com/nelmio/NelmioSecurityBundle#content-security-policy) which introduces some interesting information about security. You could completly disable inline script via a http header to reduce possible XSS attacks. 

For more information: http://www.w3.org/TR/CSP/
## Todo

The following files contains inline scripts and should be removed in the next major release, because this is a BC break.
- [ ] https://github.com/sonata-project/SonataMediaBundle/blob/master/Resources/views/MediaAdmin/edit.html.twig
- [ ] https://github.com/sonata-project/SonataMediaBundle/blob/master/Resources/views/Extra/pixlr_exit.html.twig

### LINKED ISSUE #2107: Deprecate Pixlr integration?
## Deprecation Request

Should we deprecate Pixlr integration? I am not even sure if it works or is broken (it is a really old integration and the Pixlr webpage changed a lot).

Does it make sense to have this kind of feature here or should we focus on image upload / download?

jordisala1991:
## Subject

Closes #2107 .
Closes #1092 .

## Changelog

```markdown
### Deprecated
- Deprecated Pixlr integration.
```

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -441,11 +441,18 @@ class Configuration implements ConfigurationInterface
             ->end();
     }
 
+    /**
+     * NEXT_MAJOR: Remove this method.
+     */
     private function addExtraSection(ArrayNodeDefinition $node): void
     {
         $node
             ->children()
                 ->arrayNode('pixlr')
+                    ->setDeprecated(...$this->getBackwardCompatibleArgumentsForSetDeprecated(
+                        'The node "%node%" is deprecated and will be removed in version 4.0.',
+                        '3.x'
+                    ))
                     ->info('More info at https://pixlr.com/')
                     ->addDefaultsIfNotSet()
                     ->children()
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,11 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Deprecate Pixlr integration
+
+Integration with Pixlr is deprecated now. There is no replacement since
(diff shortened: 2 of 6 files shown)


core23 (dismissed):

src/DependencyInjection/Configuration.php
```diff
@@ -446,6 +446,10 @@ private function addExtraSection(ArrayNodeDefinition $node): void
         $node
             ->children()
                 ->arrayNode('pixlr')
+                    ->setDeprecated(...$this->getBackwardCompatibleArgumentsForSetDeprecated(
+                        'The node "%node%" is deprecated and will be removed in version 4.0.',
+                        '3.28'
```
franmomu, line 451:
```suggestion
                        '3.x'
```

VincentLanglet: And should we add a NEXT_MAJOR comment ?

jordisala1991: Fixed.

core23 (changes requested):
You can also add a deprecation warning for the template

core23 (approved):

VincentLanglet (approved):

franmomu (approved):

franmomu:
Thanks @jordisala1991 !

PR #2118, opened 2021-09-15
[RFC] Deprecate ReST API

jordisala1991:
## Subject

## Changelog

```markdown
### Deprecated
- Deprecated ReST API with FOSRest, Nelmio Api Docs and JMS Serializer.
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Deprecate API
+
+Integration with FOSRest, JMS Serializer and Nelmio Api Docs is deprecated, the ReST API provided with this bundle will be removed on 4.0.
+
+If you are relying on this, consider moving to other solution like [API Platform](https://api-platform.com/) instead.
+
 ### Deprecate Pixlr integration
 
 Integration with Pixlr is deprecated now. There is no replacement since
--- a/src/Extra/ApiMediaFile.php
+++ b/src/Extra/ApiMediaFile.php
@@ -16,7 +16,9 @@ namespace Sonata\MediaBundle\Extra;
 use Symfony\Component\HttpFoundation\File\File;
 
 /**
- * @final since sonata-project/media-bundle 3.21.0
+ * NEXT_MAJOR: Remove this file.
+ *
+ * @deprecated since sonata-project/media-bundle 3.x, to be removed in 4.0.
  */
 class ApiMediaFile extends File
 {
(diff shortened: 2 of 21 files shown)


jordisala1991:
With this deprecation there are many things that are solved.

1. We stop offering an API that becomes outdated very quickly since each version of Nelmio changes completely the annotations
2. We can remove a lot of dependencies: fosRest, nelmio, JMS serializer
3. We don't have to worry about Datagrid-bundle anymore because on 4.0 it won't be needed.

I think we should take this PR as the way to go for the rest of bundles and apply the same. wdyt @sonata-project/contributors ?

UPGRADE-3.x.md
```diff
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Deprecate API
+
+Integration with FOSRest, JMS Serializer and Nelmio Api Docs is deprecated, the API provided with this bundle will be removed on 4.0.
```
phansys, line 9:
```suggestion
Integration with FOSRest, JMS Serializer and Nelmio Api Docs is deprecated, the ReST API provided with this bundle will be removed on 4.0.
```

UPGRADE-3.x.md
```diff
@@ -4,6 +4,12 @@ UPGRADE 3.x
...
+
+If you are relying on this, consider moving to other solution like Api Platform instead.
```
phansys, line 11:
```suggestion
If you are relying on this, consider moving to other solution like [API Platform](https://api-platform.com/) instead.
```

phansys (changes requested):
IMO, the title and the changelog must also reflect that we are talking specifically about the ReST API.

phansys (dismissed):

VincentLanglet (commented):
Does someone in the @sonata-project/contributors is using the Rest Api ?

src/Extra/ApiMediaFile.php
```diff
@@ -16,7 +16,11 @@
 use Symfony\Component\HttpFoundation\File\File;
 
 /**
+ * NEXT_MAJOR: Remove this file.
+ *
  * @final since sonata-project/media-bundle 3.21.0
```
core23, line 21: IMHO you can remove the `@final` information

core23 (changes requested):

src/Extra/ApiMediaFile.php
jordisala1991: Does it matter? I will remove it once I do the manual merge

core23: Yes, because this information is for the user and can be very confusing. It could either be final or deprecated.

jordisala1991: I don't see how final since 3.21 and deprecated since 3.34 could confuse our users, but changed!.

wbloszyk (approved):

core23 (approved):

VincentLanglet (approved):

jordisala1991:
3 approvals + @phansys did an approval before. Lets do this.

PR #2125, opened 2021-09-19
Improve templates

### LINKED ISSUE #2009: Fix styles for FeatureMediaBlockService
## Subject

On SonataBlockBundle 3.x the method getStylesheets and getJavascript was deprecated, and 4.0 removed it: sonata-project/SonataBlockBundle#510

We need to find a way to change our implementation (https://github.com/sonata-project/SonataMediaBundle/blob/master/src/Block/FeatureMediaBlockService.php#L94-L99) to the guidelines provided.

jordisala1991:
## Subject

I am targeting this branch, because this could be considered BC break.

Closes #2009

## Changelog

```markdown
### Changed
- Block templates are suposed to be overriden by the application to add its specific design.
```

### DIFF
--- a/src/Resources/views/Block/block_feature_media.html.twig
+++ b/src/Resources/views/Block/block_feature_media.html.twig
@@ -12,15 +12,9 @@ file that was distributed with this source code.
 {% extends sonata_block.templates.block_base %}
 
 {% block block %}
-    <div class="sonata-media-block-feature-media-container {{ settings.class }}">
-        <div class="sonata-media-block-feature-media-container-{{ settings.orientation }}">
-            {% if media and settings.format %}
-                {% media media, settings.format %}
-            {% else %}
-                {# no media selected or format defined... #}
-            {% endif %}
-        </div>
+    {% sonata_template_box 'This is the feature media block. Feel free to override it.' %}
 
+    <div class="{{ settings.class }}">
         {% if settings.title %}
             <h3>
                 {% if settings.icon %}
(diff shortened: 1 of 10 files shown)


src/Resources/views/Block/block_feature_media.html.twig
```diff
@@ -12,15 +12,9 @@ file that was distributed with this source code.
 {% extends sonata_block.templates.block_base %}
 
 {% block block %}
-    <div class="sonata-media-block-feature-media-container {{ settings.class }}">
```
jordisala1991, line 15: Those clases make no sense, this block is suposed to be used in the user app so he must define the template and all classes and styles.

src/Resources/views/Block/block_gallery.html.twig
```diff
@@ -12,63 +12,28 @@ file that was distributed with this source code.
...
-                        {{ settings.title|trans({}, settings.translation_domain) }}
-                    {% else %}
-                        {{ settings.title }}
-                    {% endif %}
-                </h3>
-            {% endif %}
-
-            {% sonata_template_box 'This is the gallery media block. Feel free to override it.' %}
-
-            <div id="carousel-{{ block.id }}" class="carousel slide"
```
jordisala1991, line 32: same for all of this, the library used by this block should be chosen and implemented by the end user (or maybe he wants to show the gallery with only css and html, without JS.

phansys (approved):

VincentLanglet (approved):

Issue #2109, opened 2021-09-12
Replace SonataNotificationBundle integration by symfony/messenger


jordisala1991:
## Feature Request

I think symfony/messenger is the way to go. Instead of having to maintain our own solution, we could provide integration with Symfony. wdyt @sonata-project/contributors ?

IMO there is no need to directly deprecate SonataNotificationBundle but we could instead stop using it on next major releases, so when everything is done, there would be no dependencies and we can deprecate it without losing anything.

wbloszyk:
IMO at first we should decide what to do with `SonataNotificationBundle`.  I do not using it in my projects and I do not know it very well. But I think we should compare functionality of these two bundles at first. The most important thing for me (IMO) is integrate notification with `AdminBundle`.

Will be awesome to require `symfony/messenger` and suggest `sonata-project/notification-bundle`.

PR #2131, opened 2021-09-21
Add Symfony messenger integration

jordisala1991:
## Subject

Part of #2109 

## Changelog

```markdown
### Added
- Added optional integration with Symfony messenger to handle async thumbnail generation
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -111,6 +111,7 @@
         "sonata-project/seo-bundle": "^2.5",
         "symfony/browser-kit": "^4.4 || ^5.2",
         "symfony/http-client": "^4.4 || ^5.2",
+        "symfony/messenger": "^4.4 || ^5.3",
         "symfony/phpunit-bridge": "^5.2",
         "symfony/security-csrf": "^4.4 || ^5.2",
         "vimeo/psalm": "^4.7.2"
@@ -124,6 +125,7 @@
         "sonata-project/notification-bundle": "If you want to generate thumbnails asynchronously.",
         "sonata-project/seo-bundle": "If you're sensible to SEO.",
         "symfony/http-client": "If you want to load external content.",
+        "symfony/messenger": "If you want to generate thumbnails asynchronously.",
         "symfony/monolog-bundle": "If you want to log exceptions produced when dealing with images.",
         "tilleuls/ckeditor-sonata-media-bundle": "If you want to use CKEditor.",
         "twig/extra-bundle": "Auto configures the Twig Intl extension."
--- a/docs/reference/messenger.rst
+++ b/docs/reference/messenger.rst
@@ -1,0 +1,48 @@
+Messenger
+=========
+
+The bundle provides a specific thumbnail class to generate thumbnails through an asynchronous task. So there no processing
+time for the user after uploading a file.
+
+It is recommended to read about `Symfony Messenger`_ if it is your first time using it.
+
+First, you need to install Symfony Messenger:
+
+.. code-block:: bash
+
+    composer require symfony/messenger
+
+After installing the dependency, you need to enable the integration and change the thumbnail configuration
+for each provider:
+
+.. code-block:: yaml
+
+    # config/packages/sonata_media.yaml
+
+    sonata_media:
+        messenger:
+            enabled: true
+        providers:
+            image:
+                thumbnail:  sonata.media.thumbnail.messenger
+            vimeo:
+                thumbnail:  sonata.media.thumbnail.messenger
+            youtube:
+                thumbnail:  sonata.media.thumbnail.messenger
+            dailymotion:
+                thumbnail:  sonata.media.thumbnail.messenger
+
+To handle async messages, make sure you configure messenger with an async transport:
+
+.. code-block:: yaml
+
+    # config/packages/messenger.yaml
+
+    framework:
+        messenger:
+            transports:
+                async: "%env(MESSENGER_TRANSPORT_DSN)%"
+            routing:
+                'Sonata\MediaBundle\Messenger\GenerateThumbnailsMessage': async
+
+.. _`Symfony Messenger`: https://symfony.com/doc/current/messenger.html
--- a/docs/reference/messenger.rst
+++ b/docs/reference/messenger.rst
@@ -45,4 +45,26 @@ To handle async messages, make sure you configure messenger with an async transp
             routing:
                 'Sonata\MediaBundle\Messenger\GenerateThumbnailsMessage': async
 
+You can also change the default bus for generate thumbnails:
+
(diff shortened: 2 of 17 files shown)


composer.json
```diff
@@ -111,6 +111,7 @@
         "sonata-project/seo-bundle": "^2.5",
         "symfony/browser-kit": "^4.4 || ^5.2",
         "symfony/http-client": "^4.4 || ^5.2",
+        "symfony/messenger": "^4.4 || ^5.3",
```
VincentLanglet, line 114: time to bump 5.2 for others too ?

VincentLanglet (dismissed):

composer.json
jordisala1991: I can do it on next Pr, but on 3.x it does not seem super necessary since it is not full sf 5 support, only a few components. On 4.x is already on 5.3 for all.

VincentLanglet: Maybe it's better to add support for symfony messenger 5.2 then ? for consistency

jordisala1991: If it is something important I can create another Pr for 3.x to up the versions of the other components.

docs/reference/messenger.rst
```diff
@@ -0,0 +1,48 @@
...
+.. code-block:: yaml
+
+    # config/packages/sonata_media.yaml
+
+    sonata_media:
+        messenger:
+            enabled: true
+        providers:
+            image:
+                thumbnail:  sonata.media.thumbnail.consumer.messenger
```
ChristianGripp: This is correct? I did not find the service id outside the docs.

jordisala1991: woops, no, the .consumer needs to be removed.

jordisala1991: Fixed.

src/Messenger/GenerateThumbnailsMessage.php
```diff
@@ -0,0 +1,41 @@
...
+ */
+final class GenerateThumbnailsMessage
+{
+    /**
+     * @var int|string|null
+     */
+    private $mediaId;
+
+    /**
+     * @param int|string|null $mediaId
```
ChristianGripp: A null value looks wrong

src/Messenger/GenerateThumbnailsMessage.php
```diff
@@ -0,0 +1,41 @@
...
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Sonata\MediaBundle\Messenger;
+
+/**
+ * @author Jordi Sala Morales <jordism91@gmail.com>
```
ChristianGripp, line 17: You can add `@psalm-pure` annotation here.

core23 (changes requested):

src/DependencyInjection/Configuration.php
```diff
@@ -565,6 +566,20 @@ private function addAdapterSection(ArrayNodeDefinition $node): void
...
+    {
+        $node
+            ->children()
+                ->arrayNode('messenger')
+                    ->canBeEnabled()
+                    ->addDefaultsIfNotSet()
+                    ->children()
+                        ->scalarNode('bus')->isRequired()->defaultValue('messenger.default_bus')->end()
+                    ->end()
+                ->end()
```
wbloszyk: IMO we should allow configure bus for each message (which can be more then one in the future):
```suggestion
                ->arrayNode('messenger_buses')
                    ->canBeEnabled()
                    ->addDefaultsIfNotSet()
                    ->children()
                        ->scalarNode('generate_thumbnails')->isRequired()->defaultValue('messenger.default_bus')->end()
                    ->end()
                ->end()
```

src/Messenger/GenerateThumbnailsMessage.php
jordisala1991: What does that annotation do?

ChristianGripp: It communicates that the object has no more dynamic parts / changing properties once the constructor was called:

https://psalm.dev/articles/immutability-and-beyond#purity

jordisala1991: Applied

src/Messenger/GenerateThumbnailsMessage.php
jordisala1991: Fixed.

src/DependencyInjection/Configuration.php
jordisala1991: Changed to reflect the bus is only for generate thumbnails.

tests/App/Entity/Gallery.php
```diff
@@ -23,16 +23,15 @@
...
      * @ORM\GeneratedValue
      * @ORM\Column(type="integer")
      */
     protected $id;
 
-    /**
-     * Get id.
-     */
-    public function getId(): int
+    public function getId()
```
wbloszyk, line 34: Can we keep return type hints here? And in other files too?
```suggestion
    public function getId(): ?int
```

tests/App/Entity/Gallery.php
```diff
@@ -23,16 +23,15 @@
 class Gallery extends BaseGallery
 {
     /**
+     * @var int|null
+     *
```
wbloszyk, line 27:
```suggestion
```

wbloszyk (changes requested):

tests/App/Entity/Gallery.php
jordisala1991: We can't, on master that typehint is removed because the getId() is a method that comes from the interface and it can accept both int/string.

tests/App/Entity/Gallery.php
jordisala1991: We can't, on master that typehint is removed because the getId() is a method that comes from the interface and it can accept both int/string.

wbloszyk (approved):

core23 (approved):

core23:
Never tried the symfony messanger, but can you do the same for the page bundle to [remove the notification bundle](https://github.com/sonata-project/SonataPageBundle/pull/1251) @jordisala1991?

jordisala1991:
> Never tried the symfony messanger, but can you do the same for the page bundle to [remove the notification bundle](https://github.com/sonata-project/SonataPageBundle/pull/1251) @jordisala1991?

The idea is to remove notification bundle yes!

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #2135, opened 2021-09-25
Add missing files

wbloszyk:
Revert missing file while #1968

### DIFF
--- a/src/Resources/config/no_driver_admin.php
+++ b/src/Resources/config/no_driver_admin.php
@@ -10,9 +10,7 @@ declare(strict_types=1);
  * For the full copyright and license information, please view the LICENSE
  * file that was distributed with this source code.
  */
-
-use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
-
-return static function (ContainerConfigurator $containerConfigurator): void {
-    $containerConfigurator->services();
-};
+/*
+ * This file will be loaded when SonataAdminBundle is enabled and db_driver has its default value,
+ * not corresponding to real persistence mechanism.
+ */
--- a/src/Resources/config/no_driver_admin.php
+++ b/src/Resources/config/no_driver_admin.php
@@ -11,9 +11,6 @@ declare(strict_types=1);
  * file that was distributed with this source code.
  */
 
-use Sonata\MediaBundle\Generator\NoDriverGenerator;
-use Sonata\MediaBundle\Model\NoDriverGalleryManager;
-use Sonata\MediaBundle\Model\NoDriverMediaManager;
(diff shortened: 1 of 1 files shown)


jordisala1991:
Please add a comment like the previous one: https://github.com/sonata-project/SonataMediaBundle/blob/db9aa7b6b49a3b2977c4ff15766698f899234b72/src/Resources/config/no_driver_admin.xml#L3

src/Resources/config/no_driver_admin.php
```diff
@@ -0,0 +1,18 @@
...
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $containerConfigurator->services();
```
jordisala1991: I dont think this line is needed

wbloszyk: In MediaBundle exists some block which require Admin services. 
IMO will be nice to allow register admin service with `manager_type: no_driver`.
https://docs.sonata-project.org/projects/SonataAdminBundle/en/4.x/getting_started/creating_an_admin/#step-2-register-the-admin-class 

Then in `no_driver_admin.php` we should add this admins. 

`no_driver` should display Admin panel with alert about missing driver.

jordisala1991: I dont understand what you mean. 

I think there is a bug about this file missing. But the line I mention it is not needed. 

The other thing you mention is maybe a feature for 3.x. Not for 4.x.

wbloszyk: I mean we should integrate `no_driver` in AdminBundle too. Then we should register admin services in `no_driver_admin.php`.

Agree. Services should be add in 3.x

src/Resources/config/no_driver_admin.php
```diff
@@ -0,0 +1,16 @@
...
+
+/*
+ * This file is part of the Sonata Project package.
+ *
+ * (c) Thomas Rabaix <thomas.rabaix@sonata-project.org>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+/*
```
jordisala1991, line 13: Did you try this change? I never saw a services file in php without the return static function.

wbloszyk: It is working on my demo.

jordisala1991 (approved):

VincentLanglet (approved):

PR #2137, opened 2021-09-27
Make FeatureMediaBlockService not extend from MediaBlockService

jordisala1991:
In the process, getStylesheet is removed since it is not used on
BlockBundle 4.0, and load is refactor to ensure we can load the
correct entity with its given id (not necessarily a integer).

## Subject

## Changelog

```markdown
### Changed
- Changed `FeatureMediaBlockService` to not extend `MediaBlockService`.
- Changed `MediaBlockService` to be final.
```

### DIFF
--- a/src/Block/FeatureMediaBlockService.php
+++ b/src/Block/FeatureMediaBlockService.php
@@ -22,10 +22,10 @@ use Sonata\BlockBundle\Form\Mapper\FormMapper;
 use Sonata\BlockBundle\Meta\Metadata;
 use Sonata\BlockBundle\Meta\MetadataInterface;
 use Sonata\BlockBundle\Model\BlockInterface;
-use Sonata\Doctrine\Model\ManagerInterface;
 use Sonata\Form\Type\ImmutableArrayType;
 use Sonata\Form\Validator\ErrorElement;
 use Sonata\MediaBundle\Model\MediaInterface;
+use Sonata\MediaBundle\Model\MediaManagerInterface;
 use Sonata\MediaBundle\Provider\Pool;
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 use Symfony\Component\Form\Extension\Core\Type\TextareaType;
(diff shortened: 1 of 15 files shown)


src/Block/FeatureMediaBlockService.php
```diff
@@ -13,22 +13,65 @@
...
+
+    /**
+     * @param AdminInterface<MediaInterface>   $mediaAdmin
+     * @param ManagerInterface<MediaInterface> $mediaManager
+     */
+    public function __construct(
+        Environment $twig,
+        Pool $pool,
+        AdminInterface $mediaAdmin,
+        ManagerInterface $mediaManager
```
ChristianGripp: You can use `MediaManagerInterface` here

core23 (changes requested):

src/Block/FeatureMediaBlockService.php
jordisala1991: Replaced all usages of ManagerInterface by the correct interface (Media or Gallery).

src/Block/FeatureMediaBlockService.php
```diff
@@ -13,22 +13,64 @@
...
+    public function __construct(
+        Environment $twig,
+        Pool $pool,
+        AdminInterface $mediaAdmin,
+        MediaManagerInterface $mediaManager
+    ) {
+        parent::__construct($twig);
+
+        $this->pool = $pool;
+        $this->mediaAdmin = $mediaAdmin;
```
wbloszyk, line 70: This property make some issue. 
When we set `db_driver: 'no_driver'` then `sonata.media.admin.media` does not exists and throw exception. 
We should:
Allow create admin service with `manager_type: none` (or `manager_type: no_driver`).
Or make `mediaAdmin` optional - then also we should not allow generate edit form without `mediaAdmin`. 
Also if we do not have admin, then we should throw exception while generate edit form (is used by `EditableBlockService`). So can you change it to optional?

wbloszyk: Or we should find way to replace them.

jordisala1991: This is working exactly the same as the MediaBlockService.
The no_driver option is not suposed to be used by the users, it is just a temporary option.

If you want to improve that part. Please create another issue explaining why is that important and/or open a Pull request showing some code to implement that part.

wbloszyk (approved):

core23 (approved):

PR #2139, opened 2021-09-29
Deprecate SonataNotificationBundle integration

### LINKED ISSUE #2109: Replace SonataNotificationBundle integration by symfony/messenger

jordisala1991:
## Subject

Closes https://github.com/sonata-project/SonataMediaBundle/issues/2109

## Changelog

```markdown
### Deprecated
- Deprecated SonataNotificationBundle integration. Use Symfony Messenger integration instead.
```

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,12 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Deprecate Async processing with SonataNotificationBundle
+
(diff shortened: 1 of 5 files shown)


wbloszyk (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #2145, opened 2021-10-02
Move to twig functions

### LINKED ISSUE #2103: Move to twig functions
## Feature Request

We are currently using custom twig tags like `{% media %}`. We should do like twig itself an move everything that is not a structure control like if , for , etc.. to functions, like `{{ media() }}`.

I think this can be done without breaking BC, but to move to `TwigRuntime` we might need to do it on master, because otherwise we would be duplicating code.

jordisala1991:
## Subject

Closes #2103.

## Changelog

```markdown
### Added
- Added twig functions to load medias, thumbnails or paths
### Deprecated
- Deprecated using twig tags to load media, thumbnails or paths
```

### DIFF
--- a/src/Twig/Extension/FormatterMediaExtension.php
+++ b/src/Twig/Extension/FormatterMediaExtension.php
@@ -14,11 +14,13 @@ declare(strict_types=1);
 namespace Sonata\MediaBundle\Twig\Extension;
 
 use Sonata\MediaBundle\Model\MediaInterface;
+use Sonata\MediaBundle\Twig\MediaRuntime;
 use Sonata\MediaBundle\Twig\TokenParser\MediaTokenParser;
 use Sonata\MediaBundle\Twig\TokenParser\PathTokenParser;
 use Sonata\MediaBundle\Twig\TokenParser\ThumbnailTokenParser;
 use Twig\Extension\AbstractExtension;
 use Twig\Extension\ExtensionInterface;
+use Twig\TwigFunction;
 
 /**
  * @final since sonata-project/media-bundle 3.21.0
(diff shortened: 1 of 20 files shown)


jordisala1991:
I am missing docs and upgrade note, but it would be nice to get a first review about code.

src/Twig/Extension/FormatterMediaExtension.php
```diff
@@ -105,26 +124,42 @@ public function path($media, $format)
...
         return $this->getTwigExtension()->getTests();
     }
 
     public function getFunctions()
     {
-        return $this->getTwigExtension()->getFunctions();
+        return [
+            new TwigFunction('sonata_media', [MediaRuntime::class, 'media']),
+            new TwigFunction('sonata_thumbnail', [MediaRuntime::class, 'thumbnail']),
+            new TwigFunction('sonata_path', [MediaRuntime::class, 'path']),
```
jordisala1991, line 156: this is my proposal about names of new functions, but we could remove the 'sonata_' if you prefer it. I was trying to avoid colisions with possible user defined functions.

wdyt @VincentLanglet ?

VincentLanglet: I think it's ok to have those prefix for the reason you gave (and for context). And I already saw method with the sonata prefix in some other bundles.

VincentLanglet (dismissed):

VincentLanglet:
I assume we can do https://github.com/sonata-project/SonataAdminBundle/issues/7474 the same way

VincentLanglet (approved):

jordisala1991:
After this one is merged we should do a release on 3.x (probably the last minor).

wbloszyk (approved):

wbloszyk:
What do you think about use this solution (twig functions with `sonata_` prefix) in IntlBundle? It will resolve https://github.com/sonata-project/SonataIntlBundle/issues/299

jordisala1991:
> What do you think about use this solution (twig functions with `sonata_` prefix) in IntlBundle? It will resolve [sonata-project/SonataIntlBundle#299](https://github.com/sonata-project/SonataIntlBundle/issues/299)

I think we should prefix all our functions / filters, because of the collisions that could happen. But I am not sure about the intlbundle functions. Are they meant to replace the default one provided by symfony or they are just different? If they are just different, we should indeed prefix them. If they are meant to be a replacement, then further discussion might be needed.

PR #2146, opened 2021-10-05
Added more information for VimeoProvider in service configuration

eerison:
## Update documentation for VimeoProvider in service configuration

I am targeting this branch, because it's related with current version.

## Changelog

```markdown
### Added

- Added more documentation for `VimeoProvider` in service configuration.
```

this ticket is related with Issue: #2106

### DIFF
--- a/docs/reference/creating_a_provider_class.rst
+++ b/docs/reference/creating_a_provider_class.rst
@@ -265,8 +265,8 @@ added to the provider pool.
             - '@sonata.media.cdn.server'
             - '@sonata.media.generator.default'
             - '@sonata.media.thumbnail.format'
-            - '@sonata.media.http.client'
-            - '@sonata.media.metadata.proxy'
+            # - '@sonata.media.http.client' // It's an optional parameter.
+            # - '@sonata.media.metadata.proxy' // This parameter is required when you are using PSR client.
         calls:
             -
                 - setTemplates


docs/reference/creating_a_provider_class.rst
```diff
@@ -265,8 +265,8 @@ added to the provider pool.
             - '@sonata.media.cdn.server'
             - '@sonata.media.generator.default'
             - '@sonata.media.thumbnail.format'
-            - '@sonata.media.http.client'
-            - '@sonata.media.metadata.proxy'
+            # - '@sonata.media.http.client' //It's a optinonal parameter
```
eerison:
```suggestion
            # - '@sonata.media.http.client' //It's an optional parameter.
```

docs/reference/creating_a_provider_class.rst
```diff
@@ -265,8 +265,8 @@ added to the provider pool.
             - '@sonata.media.cdn.server'
             - '@sonata.media.generator.default'
             - '@sonata.media.thumbnail.format'
-            - '@sonata.media.http.client'
-            - '@sonata.media.metadata.proxy'
+            # - '@sonata.media.http.client' //It's an optional parameter.
+            # - '@sonata.media.metadata.proxy' // This parameter is required only, When you are using PSR client.
```
eerison:
```suggestion
            # - '@sonata.media.metadata.proxy' // This parameter is required when you are using PSR client.
```

docs/reference/creating_a_provider_class.rst
```diff
@@ -277,7 +277,6 @@ added to the provider pool.
                 - ['@sonata.media.resizer.simple']
         tags:
             - { name: sonata.media.provider }
-        public: true
```
jordisala1991: This might be true for 4.x, but on 3.x I believe it is needed for the providers to be public.

eerison: Hey @jordisala1991 

But the lest version of Media bundle support symfony 4.4 or higher

3.x

jordisala1991: Yes, but that does not relate to need a public service or not. In this case, we need it public because on SonataMediaBundle 3.x there are multiple places where we use the container directly to fetch services. That is not the case for 4.x. That's why there is this difference between the 2 versions.

eerison: But in this case it's an example when I'll create a custom provider, I guess in this case the dev should create without public: true , don't you?

It will be in the project source and not in sonata source 🤔

jordisala1991: Yep, but if the SonataMediaBundle code does something like:

`$this->container->get($media->getProviderName());`

If the definition is not public, it will fail.

jordisala1991: This was the PR introducing the public explicitly for the providers: https://github.com/sonata-project/SonataMediaBundle/pull/1432

It doesn't explain why it is needed.
Anyway, on my opinion, it is safer to keep the public true here (unless you or someone else can confirm and searched on SonataMediaBundle code for any suspicious code that might get a provider directly from the container) and remove it from 4.x because it is already working without it.

eerison: Ohhh now I got it

yeah you are right without `public:true`, it'll broken here

```
$this->container->get($media->getProviderName());
```

eerison: Hey @jordisala1991 thank you for your explanation :) But I'm using provider without `public:true` and I don't get any kind of error 🤯 

```yaml
    sonata.media.provider.jpg:
        class: Sonata\MediaBundle\Provider\ImageProvider
        arguments:
            - sonata.media.provider.jpg
            - '@sonata.media.filesystem.local'
            - '@sonata.media.cdn.server'
            - '@sonata.media.generator.default'
            - '@sonata.media.thumbnail.format'
            - [ 'jpg' ]
            - [ 'image/pjpeg', 'image/jpeg' ]
            - '@sonata.media.adapter.image.gd'
            - '@sonata.media.metadata.proxy'
        tags:
            - { name: sonata.media.provider }
```

docs/reference/creating_a_provider_class.rst
```diff
@@ -265,8 +265,8 @@ added to the provider pool.
             - '@sonata.media.cdn.server'
             - '@sonata.media.generator.default'
             - '@sonata.media.thumbnail.format'
-            - '@sonata.media.http.client'
-            - '@sonata.media.metadata.proxy'
+            # - '@sonata.media.http.client' //It's an optional parameter.
```
jordisala1991: can you add an space between // and I?

eerison: done ;)

VincentLanglet (approved):

jordisala1991 (approved):

VincentLanglet:
Thanks

PR #2152, opened 2021-10-06
Fix with null media and add safe output for html

jordisala1991:
## Subject

I am targeting this branch, because this is bug fixing.

## Changelog

```markdown
### Fixed
- Fixed safe output of `sonata_media` and `sonata_thumbnail` functions
- Fixed calling `sonata_media` and `sonata_thumbnail` functions with null media
```

### DIFF
--- a/src/Twig/MediaRuntime.php
+++ b/src/Twig/MediaRuntime.php
@@ -47,8 +47,8 @@ final class MediaRuntime implements RuntimeExtensionInterface
     }
 
     /**
-     * @param int|string|MediaInterface $media
-     * @param array<string, mixed>      $options
+     * @param int|string|MediaInterface|null $media
+     * @param array<string, mixed>           $options
      */
     public function media($media, string $format, array $options = []): string
     {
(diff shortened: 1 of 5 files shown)


src/Twig/MediaRuntime.php
```diff
@@ -140,11 +140,11 @@ public function path($media, string $format): string
     }
 
     /**
-     * @param MediaInterface|int|string $media
+     * @param int|string|MediaInterface|null $media
      */
     private function getMedia($media): ?MediaInterface
     {
-        if (!$media instanceof MediaInterface) {
+        if (!$media instanceof MediaInterface && null !== $media) {
```
jordisala1991, line 147: we might deprecate after this PR the posibility to pass null here. wdyt @sonata-project/contributors ?

IMO we should allow it on 3.x (to not complicate more the update) but with a deprecation message, and remove it on 4.x)

VincentLanglet: Agree, i dont find natural to support null

jordisala1991: Yep, it is needed to introduce it on 3.x again, because otherwise we are breaking BC, but we can remove it on 4.x for sure. I will prepare another PR to deprecate it.

franmomu (approved):

PR #2158, opened 2021-10-08
Deprecate passing null to twig media functions

jordisala1991:
## Subject

## Changelog

```markdown
### Deprecated
- Deprecated passing null value for the media parameter to twig functions.
```

### DIFF
--- a/src/Twig/MediaRuntime.php
+++ b/src/Twig/MediaRuntime.php
@@ -144,6 +144,20 @@ final class MediaRuntime implements RuntimeExtensionInterface
      */
     private function getMedia($media): ?MediaInterface
     {
+        // NEXT_MAJOR: Throw exception instead and remove references to null media being accepted.
+        if (!\is_int($media) && !\is_string($media) && !$media instanceof MediaInterface) {
+            @trigger_error(
+                'Using SonataMediaBundle custom Twig functions with a media that is not'
+                .' the identifier of the media or the media itself is deprecated since'
+                .' sonata-project/media-bundle 3.x and will be removed in version 4.0.',
+                \E_USER_DEPRECATED
+            );
+            // throw new \TypeError(sprintf(
+            //     'Media parameter must be either an identifier or the media itself for Twig functions, %s given.',
+            //     $media
+            // ));
+        }
+
         if (!$media instanceof MediaInterface && null !== $media) {
             $media = $this->mediaManager->find($media);
         }
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,11 +4,17 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
-# Deprecated `MigrateToJsonTypeCommand`
+### Deprecate `null` media value on Twig functions
+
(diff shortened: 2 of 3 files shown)


franmomu (dismissed):

src/Twig/MediaRuntime.php
```diff
@@ -144,6 +144,17 @@ public function path($media, string $format): string
...
     {
+        // NEXT_MAJOR: Throw exception instead and remove references to null media being accepted.
+        if (null === $media) {
+            @trigger_error(
+                'Using SonataMediaBundle custom Twig functions with a null media is'
+                .'deprecated since sonata-project/media-bundle 3.x and will be removed'
+                .' in version 4.0.',
+                \E_USER_DEPRECATED
+            );
+            // throw new \RuntimeException('Media parameter must not be null for Twig functions.');
```
VincentLanglet: InvalidArgumentException

VincentLanglet (dismissed):

src/Twig/MediaRuntime.php
jordisala1991: Done.

VincentLanglet (approved):

src/Twig/MediaRuntime.php
```diff
@@ -144,6 +144,17 @@ public function path($media, string $format): string
      */
     private function getMedia($media): ?MediaInterface
     {
+        // NEXT_MAJOR: Throw exception instead and remove references to null media being accepted.
```
ChristianGripp, line 147: Maybe remove `null` from the parameter typehints?

core23 (dismissed):

src/Twig/MediaRuntime.php
VincentLanglet: Might be enough indeed, like we don't check if someone pass a boolean or an object.

VincentLanglet (changes requested):

src/Twig/MediaRuntime.php
jordisala1991: I can remove null typehint only on 4.x, after the merge

VincentLanglet: Sure, but then, do we still need to throw an InvalidArgument ?
Or the method might need to be rewritten in a way we don't only check for null value but also value which is not int|string|MediaInterface

VincentLanglet (dismissed):

src/Twig/MediaRuntime.php
VincentLanglet: Anyway, we can still merge this and discuss it on the 4.x PR

jordisala1991: Oh I see, since we are not relying on phpdoc and we can't typehint it (we are not on php 8) it might be better to check if the value is not accepted here? and throw a TypeError on 4.x?

VincentLanglet: Yes

jordisala1991: Wdyt about the new implementation?

VincentLanglet (dismissed):

VincentLanglet (approved):

core23 (approved):

PR #2163, opened 2021-10-11
Manual merge with NEXT_MAJOR resolved

jordisala1991:


src/Twig/MediaRuntime.php
```diff
@@ -140,11 +140,18 @@ public function path($media, string $format): string
...
-     * @param int|string|MediaInterface|null $media
+     * @param int|string|MediaInterface $media
      */
     private function getMedia($media): ?MediaInterface
     {
-        if (!$media instanceof MediaInterface && null !== $media) {
+        if (!\is_int($media) && !\is_string($media) && !$media instanceof MediaInterface) {
+            throw new \TypeError(sprintf(
+                'Media parameter must be either an identifier or the media itself for Twig functions, %s given.',
+                $media
```
VincentLanglet: you're casting media to string here
you have to do
```
is_object($media) ? 'instance of '.get_class($media) : get_type($media)
```

tests/Twig/MediaRuntimeTest.php
```diff
@@ -64,6 +64,8 @@ protected function setUp(): void
 
     public function testWithNullInput(): void
```
VincentLanglet, line 71: I'm not sure we need to keep this test.

But if you want to keep it you'll need a phpstan-ignore and a psalm suppress like we did for other tests like this.

VincentLanglet (changes requested):

core23 (dismissed):

tests/Twig/MediaRuntimeTest.php
jordisala1991: To test we throw an exception we need it, otherwise it does not cover that if. I will add ignores.

src/Twig/MediaRuntime.php
jordisala1991: Done!

tests/Twig/MediaRuntimeTest.php
jordisala1991: Fixed.

VincentLanglet (approved):

PR #2167, opened 2021-10-16
3.35.0

jordisala1991:


jordisala1991:
We have the same problem across all sonata packages @VincentLanglet

jordisala1991:
```
Loading composer repositories with package information
Restricting packages listed in "symfony/symfony" to "4.4.*"
Updating dependencies
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires symfony/phpunit-bridge *@dev, found symfony/phpunit-bridge[9.5.99] but these were not loaded, likely because it conflicts with another require.

Running update with --no-dev does not mean require-dev is ignored, it just means the packages will not be installed. If dev requirements are blocking the update you have to resolve those problems.
Error: Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires symfony/phpunit-bridge *@dev, found symfony/phpunit-bridge[9.5.99] but these were not loaded, likely because it conflicts with another require.

Script [ $COMPOSER_DEV_MODE -eq 0 ] || vendor/bin/simple-phpunit install handling the post-update-cmd event returned with error code 2
Error: The process '/usr/local/bin/composer' failed with exit code 2
```

jordisala1991:
Installing phpunit directly instead of simple-phpunit can be a work around, but this will require some work on all sonata packages.

composer.json
```diff
@@ -101,6 +101,7 @@
         "phpstan/phpstan-phpunit": "^0.12.16",
         "phpstan/phpstan-strict-rules": "^0.12.10",
         "phpstan/phpstan-symfony": "^0.12.21",
+        "phpunit/phpunit": "^9.5",
```
ChristianGripp: We should not rely on `phpunit` as a direct dependency, because this depencencies can conflict with our project specific.

ChristianGripp: Maybe we can use the [composer-bin](https://github.com/bamarni/composer-bin-plugin) plugin to install all tools. I'm using this for all my other projects like https://github.com/nucleos/NucleosUserBundle and works quite good.

VincentLanglet:
> Installing phpunit directly instead of simple-phpunit can be a work around, but this will require some work on all sonata packages.

This would solve https://github.com/sonata-project/dev-kit/issues/1559, might be the good time then.

composer.json
VincentLanglet: See https://github.com/symfony/test-pack/pull/10 and https://github.com/symfony/recipes/pull/906
Symfony is now recommending to require phpunit directly.

We had an issue about it https://github.com/sonata-project/dev-kit/issues/1559

VincentLanglet:
Other solution would be https://github.com/sonata-project/SonataClassificationBundle/pull/746

composer.json
jordisala1991: Fixed.

VincentLanglet (approved):

PR #2168, opened 2021-10-16
Update `FixMediaContextCommand::__construct()` to avoid requiring a useless argument

phansys:
## Subject
Update `FixMediaContextCommand::__construct()` to avoid requiring a useless argument.

`CategoryManagerInterface::getRootCategory()` will always return an instance of `CategoryInterface`:
https://github.com/sonata-project/SonataMediaBundle/blob/af6dc96b377164cdfee8664bc9056ae882d58c25/src/Model/CategoryManagerInterface.php#L24-L29
So, the check for an empty root category and the attempt to create a new one are useless:
https://github.com/sonata-project/SonataMediaBundle/blob/af6dc96b377164cdfee8664bc9056ae882d58c25/src/Command/FixMediaContextCommand.php#L84-L95

## Changelog

```markdown
### Changed
- `FixMediaContextCommand::__construct()` signature in order to accept an instance of `ContextManagerInterface` as argument 2.

### Deprecated
- Passing other type than `null` or `ContextManagerInterface` as argument 2 for `FixMediaContextCommand::__construct()`;
- Calling `FixMediaContextCommand::__construct()` with more than 2 arguments.
```

## To do
- [x] Add an upgrade note.

### DIFF
--- a/UPGRADE-3.x.md
+++ b/UPGRADE-3.x.md
@@ -4,6 +4,11 @@ UPGRADE 3.x
 UPGRADE FROM 3.x to 3.x
 =======================
 
+### Sonata\MediaBundle\Command\FixMediaContextCommand::__construct()
+
+Argument 3 in `FixMediaContextCommand::__construct()` is deprecated, it is moved to the argument 2.
+Passing other type than `null` or `Sonata\ClassificationBundle\Model\ContextManagerInterface` as argument 2 is deprecated.
+
 ### Deprecate `null` media value on Twig functions
 
 Our custom twig functions (`sonata_media`, `sonata_thumbnail` and `sonata_path`) used to accept a null value and produce an empty result.
--- a/src/Command/FixMediaContextCommand.php
+++ b/src/Command/FixMediaContextCommand.php
@@ -32,21 +32,52 @@ class FixMediaContextCommand extends Command
     private $mediaPool;
 
     /**
-     * @var CategoryManagerInterface|null
-     */
-    private $categoryManager;
-
-    /**
      * @var ContextManagerInterface|null
      */
     private $contextManager;
 
-    public function __construct(Pool $mediaPool, ?CategoryManagerInterface $categoryManager = null, ?ContextManagerInterface $contextManager = null)
+    /**
+     * NEXT_MAJOR: Remove the argument 3 and use `?ContextManagerInterface $contextManager = null` in argument 2.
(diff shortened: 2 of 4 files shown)


jordisala1991:
I think this is not correct. The method always returns because it is creating it if it does not exist. So at the end you need to call it to ensure it is created. Check 4.x branch for related changes / related Prs

phansys:
> I think this is not correct. The method always returns because it is creating it if it does not exist. So at the end you need to call it to ensure it is created. Check 4.x branch for related changes / related Prs

I understand your suggestion, and I'll enumerate the points I think are relevant from my perspective:
1. Currently (at 3.x), the docblock at `CategoryManagerInterface::getRootCategory()` states this method is not able to return other type than `CategoryInterface`, so the check `!$defaultCategory` will always evaluate to `false`;
2. In the 4.x branch
    1. `CategoryManagerInterface` was removed (https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Model/CategoryManagerInterface.php);
    2. The usage of `Sonata\ClassificationBundle\Model\CategoryManagerInterface` at `FixMediaContextCommand` seems to be doing nothing: https://github.com/sonata-project/SonataMediaBundle/blob/2af12f428a3a6a70fa2af58ac50dc399b1a9ffb1/src/Command/FixMediaContextCommand.php#L75, although checking the implementation, it seems to have other responsibilities other than just "get": https://github.com/sonata-project/SonataClassificationBundle/blob/6bd4b8f24eda7633733b8e99a71c732b66457ffc/src/Entity/CategoryManager.php#L110-L119.

GIven this situation, I think we should:
1. In 3.x update the docblock at `CategoryManagerInterface::getRootCategory()` in order to expose it is able to return `null`.
2. Add comments in the 4.x branch (in the method and its callers) indicating that the call to `getRootCategoriesForContext()` is required to trigger the creation of the categories if they don't exist.

Please, let me know what do you think and if I am missing something.
Thank you.

jordisala1991:
For the second point , do you mean doing that docblock adding on SonataClassificationBundle right?

Besides that question I agree with what you say. If we want to go further, maybe the method should be getOrCreated or something else. But atleast having some comment is good to understand what its doing.

We removed those interfaces because they are on SonataClassificationBundle and are used as dev dependencies (without causing recursion because classification no longer has media as a dependency)

PR #2179, opened 2021-10-19
4.0.0-alpha1

jordisala1991:
After this merge I will open the 5.x branch and do the dev-kit PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.0.0-alpha1](https://github.com/sonata-project/SonataMediaBundle/compare/3.35.1...4.0.0-alpha1) - 2021-10-20
+### Added
+- [[#2164](https://github.com/sonata-project/SonataMediaBundle/pull/2164)] Added property type hints ([@jordisala1991](https://github.com/jordisala1991))
+
+### Changed
+- [[#2157](https://github.com/sonata-project/SonataMediaBundle/pull/2157)] `cdnIsFlushable` property on Media entity is non nullable now. ([@jordisala1991](https://github.com/jordisala1991))
+- [[#2137](https://github.com/sonata-project/SonataMediaBundle/pull/2137)] Changed `FeatureMediaBlockService` to not extend `MediaBlockService`. ([@jordisala1991](https://github.com/jordisala1991))
+- [[#2137](https://github.com/sonata-project/SonataMediaBundle/pull/2137)] Changed `MediaBlockService` to be final. ([@jordisala1991](https://github.com/jordisala1991))
+- [[#2125](https://github.com/sonata-project/SonataMediaBundle/pull/2125)] Block templates are suposed to be overriden by the application to add its specific design. ([@jordisala1991](https://github.com/jordisala1991))
+
+### Fixed
+- [[#2156](https://github.com/sonata-project/SonataMediaBundle/pull/2156)] Missing CDN part when generating public URL for FileProvider with reference format ([@nicolas-anaxago](https://github.com/nicolas-anaxago))
+- [[#2150](https://github.com/sonata-project/SonataMediaBundle/pull/2150)] Fixed load controllers only when SonataAdminBundle is enabled. ([@jordisala1991](https://github.com/jordisala1991))
+
+### Removed
+- [[#2164](https://github.com/sonata-project/SonataMediaBundle/pull/2164)] Removed support for PHP 7.3 ([@jordisala1991](https://github.com/jordisala1991))
+
 ## [3.35.1](https://github.com/sonata-project/SonataMediaBundle/compare/3.35.0...3.35.1) - 2021-10-18
 ### Fixed
 - [[#2175](https://github.com/sonata-project/SonataMediaBundle/pull/2175)] Fixed calling `media_thumbnail` instead of `sonata_thumbnail` ([@jordisala1991](https://github.com/jordisala1991))


VincentLanglet (dismissed):

wbloszyk (dismissed):

wbloszyk:
Not related to this PR:
We should add other admin persistance bundle to conflict too:
https://github.com/sonata-project/SonataClassificationBundle/blob/4.x/composer.json#L42-L45

jordisala1991:
> Not related to this PR: We should add other admin persistance bundle to conflict too: https://github.com/sonata-project/SonataClassificationBundle/blob/4.x/composer.json#L42-L45

Yep, that can be done, not a blocker for this PR tho, I don't think we will have problems there for now.

wbloszyk:
> > Not related to this PR: We should add other admin persistance bundle to conflict too: https://github.com/sonata-project/SonataClassificationBundle/blob/4.x/composer.json#L42-L45
> 
> Yep, that can be done, not a blocker for this PR tho, I don't think we will have problems there for now.

Or we can remove persistance bundle from conflict becouse AdminBundle is required in them.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.0.0-alpha.1](https://github.com/sonata-project/SonataMediaBundle/compare/3.35.1...4.0.0-alpha.1) - 2021-10-19
+See UPGRADE-4.0.md
```
ChristianGripp: You can generate a full changelog using the command from my dev-kit PR

jordisala1991: > You can generate a full changelog using the command from my dev-kit PR

It almost works, I had to return true on the canBeReleased because we don't have properly tags for the manual merge PRs, and maybe there are some other things that has to be changed. Anyway I got a quick fix and I was able to generate it.

jordisala1991:
To avoid problems with the alpha versions, I think we need to follow https://getcomposer.org/doc/articles/versions.md#vcs-tags-and-branches

I will take out the dot between alpha and 1. So it is 4.0.0-alpha1

core23 (approved):

VincentLanglet:
> To avoid problems with the alpha versions, I think we need to follow https://getcomposer.org/doc/articles/versions.md#vcs-tags-and-branches
> 
> I will take out the dot between alpha and 1. So it is 4.0.0-alpha1

Can you add a test to check that we support this too: https://github.com/sonata-project/dev-kit/blob/753862c55e2e202b4f5f1264b0e85c798b3f4706/src/Action/DetermineNextReleaseVersion.php#L35

phansys (approved):
I have the same feeling about changing "alpha.1" with "alpha1".
https://semver.org/#spec-item-11
>Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
>
>Identifiers consisting of only digits are compared numerically.
>
>Identifiers with letters or hyphens are compared lexically in ASCII sort order.
>
>Numeric identifiers always have lower precedence than non-numeric identifiers.
>
>A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.
>
>Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.

That said, I think both are valid as build metadata.

jordisala1991:
> I have the same feeling about changing "alpha.1" with "alpha1".
> 
> https://semver.org/#spec-item-11
> 
> >Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
> 
> >
> 
> >Identifiers consisting of only digits are compared numerically.
> 
> >
> 
> >Identifiers with letters or hyphens are compared lexically in ASCII sort order.
> 
> >
> 
> >Numeric identifiers always have lower precedence than non-numeric identifiers.
> 
> >
> 
> >A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.
> 
> >
> 
> >Example: 1.0.0-alpha < 1.0.0-alpha.1 < 1.0.0-alpha.beta < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0.
> 
> 
> 
> That said, I think both are valid as build metadata.

I dont understand, do you think it is okay to use alpha1 or not? So far we tried alpha-1 and alpha.1 and both had problems. Im okay keeping alpha.1, whatever works better

jordisala1991:
> > To avoid problems with the alpha versions, I think we need to follow https://getcomposer.org/doc/articles/versions.md#vcs-tags-and-branches
> 
> > 
> 
> > I will take out the dot between alpha and 1. So it is 4.0.0-alpha1
> 
> 
> 
> Can you add a test to check that we support this too: https://github.com/sonata-project/dev-kit/blob/753862c55e2e202b4f5f1264b0e85c798b3f4706/src/Action/DetermineNextReleaseVersion.php#L35

Yep, I can take a look, dont think we have a problem there since we cant create non stable releases from devkit (nice feature to be added btw). Tomorrow I will merge this one and prepare the needed PRs if I see some problems on devkit

phansys:
>I dont understand, do you think it is okay to use alpha1 or not? So far we tried alpha-1 and alpha.1 and both had problems. Im okay keeping alpha.1, whatever works better

I'd love if we could use the SemVer proposal "alpha.1", but I'm fine if we need to use "alpha1" instead, as it is still valid build metadata.

PR #2183, opened 2021-10-29
Rename to primary and secondary

jordisala1991:
## Subject

## Changelog

```markdown
### Changed
- Changed replicate filesystem configuration, from `master` and `slave` to `primary` and `secondary`.
```

### DIFF
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -1,6 +1,30 @@
 UPGRADE FROM 3.x to 4.0
 =======================
 
+### Replicate Filesystem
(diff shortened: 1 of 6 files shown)


VincentLanglet (commented):
Shouldn't we provide a BC-way to do this change ?

phansys (commented):
I agree, we should provide a BC layer including the corresponding deprecations.
Otherwise, this change should be introduced in 3.x, in order to allow the replace in 4.x.

core23 (approved):

PR #2188, opened 2021-10-30
Manual merge

jordisala1991:
With NEXT_MAJOR solved


UPGRADE-3.x.md
```diff
@@ -18,6 +22,34 @@ This command was introduced a long time ago to migrate from array to json. Make
...
+            slave: sonata.media.adapter.filesystem.local
+```
+
+After:
+
+```yaml
+sonata_media:
+    cdn:
+        fallback:
+            primary: sonata.media.cdn.panther
```
phansys, line 46: IMO, we should not use this CDN in the "After" example, as it is also deprecated.

src/DependencyInjection/Configuration.php
```diff
@@ -269,10 +257,11 @@ private function addFilesystemSection(ArrayNodeDefinition $node): void
                             ->end()
                         ->end()
 
+                        // NEXT_MAJOR: Remove master and slave and make primary and secondary required
```
phansys:
```suggestion
```

UPGRADE-3.x.md
jordisala1991: This has to be solved on 3.x branch as this is just a manual merge.

phansys (changes requested):

phansys (approved):

VincentLanglet (approved):

PR #2191, opened 2021-10-31
Add finals to abstract classes

jordisala1991:
This is to show how it will end. The PR will be done on 3.x first to mark them as final with annotations, then on the up merge we will change to php code.

Things not marked as final:

BaseMediaAdmin
* __construct
* prePersist
* configureListFields
* configureFormFields
* alterNewInstance
* configurePersistentParameters

BaseGallery (Document and Entity)
* __construct
* prePersist
* preUpdate

BaseGalleryItem (Document and Entity)
* prePersist
* preUpdate

BaseMedia (Document and Entity)
* __construct
* prePersist
* preUpdate

Gallery (Model)
* __toString

GalleryItem (Model)
* __toString

Media (Model)
* __toString
* getStatusList

GalleryManager
* create

BaseProvider
* __construct
* getProviderMetadata
* preRemove
* postRemove
* prePersist
* preUpdate
* validate

BaseVideoProvider
* __construct
* getProviderMetadata
* buildEditForm
* buildCreateForm
* buildMediaType
* postUpdate
* postPersist
* postRemove
* getMetadata
* getBoxHelperProperties

BaseMediaEventSubscriber
* __construct

@VincentLanglet wdyt?

### DIFF
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -75,7 +75,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
         $object->setContext($parameters['context']);
     }
 
-    protected function configurePersistentParameters(): array
+    final protected function configurePersistentParameters(): array
     {
         $parameters = [];
 
@@ -124,7 +124,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
         ]);
     }
 
-    protected function alterNewInstance(object $object): void
+    final protected function alterNewInstance(object $object): void
     {
         if ($this->hasRequest()) {
             if ($this->getRequest()->isMethod('POST')) {
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -75,7 +75,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
         $object->setContext($parameters['context']);
     }
 
-    final protected function configurePersistentParameters(): array
+    protected function configurePersistentParameters(): array
     {
         $parameters = [];
 
@@ -124,7 +124,7 @@ abstract class BaseMediaAdmin extends AbstractAdmin
         ]);
     }
 
-    final protected function alterNewInstance(object $object): void
+    protected function alterNewInstance(object $object): void
     {
         if ($this->hasRequest()) {
             if ($this->getRequest()->isMethod('POST')) {
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -106,7 +106,7 @@ abstract class Gallery implements GalleryInterface
         return $this->defaultFormat;
     }
 
-    public function setGalleryItems(Collection $galleryItems): void
+    final public function setGalleryItems(Collection $galleryItems): void
     {
         $this->galleryItems = new ArrayCollection();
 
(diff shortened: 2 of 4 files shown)


VincentLanglet:
Seems good

core23:
> Things not marked as final:
> 
> BaseMediaAdmin
> 
> * __construct
> * prePersist
> * configureListFields
> * configureFormFields

Why? There is no override inside the impl classes

VincentLanglet:
> > Things not marked as final:
> > BaseMediaAdmin
> > 
> > * __construct
> > * prePersist
> > * configureListFields
> > * configureFormFields
> 
> Why? There is no override inside the impl classes

I think the idea was to allow people to use the BaseMediaAdmin for their admin. And those are really important extension points.

jordisala1991:
> > Things not marked as final:
> > BaseMediaAdmin
> > 
> > * __construct
> > * prePersist
> > * configureListFields
> > * configureFormFields
> 
> Why? There is no override inside the impl classes

We are not marking the models as final, so anyone can add more fields, so it makes sense to be able to customize the list and the form too.

core23:
When adding the final modifier to all kind of classes, we should apply this to the admin classes as well. When someone needs to extend this, we could remove the final modifier.


> We are not marking the models as final, so anyone can add more fields, so it makes sense to be able to customize the list and the form too.

We can use `AdminExtensions` to modify the admin list and forms.

jordisala1991:
> When adding the final modifier to all kind of classes, we should apply this to the admin classes as well. When someone needs to extend this, we could remove the final modifier.
> 
> 
> 
> 
> 
> > We are not marking the models as final, so anyone can add more fields, so it makes sense to be able to customize the list and the form too.
> 
> 
> 
> We can use `AdminExtensions` to modify the admin list and forms.

This has to be modified on 3.x . It was the purpose of the last Pr merged. If you feel like there are more methods to be made final, feel free to open another Pr. To me it was the last one, this task has taken too much time for me. I am only able to finish this Pr to add final to methods that do not exist on 3.x

jordisala1991:
⚠️ This Pr only contains final changes to classes that did not exist on 3.x or method that did not exist on 3.x ⚠️

src/Admin/BaseMediaAdmin.php
```diff
@@ -63,7 +63,7 @@ public function prePersist(object $object): void
         $object->setContext($parameters['context']);
     }
 
-    public function configurePersistentParameters(): array
+    final public function configurePersistentParameters(): array
```
VincentLanglet: not sure about this one.

src/Admin/BaseMediaAdmin.php
```diff
@@ -112,7 +112,7 @@ public function configurePersistentParameters(): array
         ]);
     }
 
-    public function alterNewInstance(object $object): void
+    final public function alterNewInstance(object $object): void
```
VincentLanglet: not sure about this one too.

src/Admin/BaseMediaAdmin.php
VincentLanglet: but this can be protected

src/Admin/BaseMediaAdmin.php
VincentLanglet: but this can be protected

src/Admin/BaseMediaAdmin.php
VincentLanglet: And prepersist can be protected too

src/Admin/BaseMediaAdmin.php
ChristianGripp: According to https://github.com/sonata-project/SonataMediaBundle/pull/2194#discussion_r740782622, we should mark everything as final by default

The scope of the method comes from the `AdminInterface`

VincentLanglet: `alterNewInstance` does not come from the AdminInterface.
It's already not final in the AbstractAdmin because it's an extension point.
That's why we should not open the visibility, it's protected in the AbstarctAdmin and should stay it.
Extension point of the AbstractAdmin shouldn't be final in AbstractMediaAdmin because it's still an extension point.

In the same way configureListField wasn't final.

src/Admin/BaseMediaAdmin.php
jordisala1991: I will take a look at those protected, and change them in another Pr

src/Admin/BaseMediaAdmin.php
jordisala1991: Yes, I was not sure about this one, I will revert it.

jordisala1991:
About the protected thing: #2201

src/Admin/BaseMediaAdmin.php
jordisala1991: Revert the final keyword

src/Admin/BaseMediaAdmin.php
jordisala1991: Revert the final keyword.

VincentLanglet (approved):

VincentLanglet:
Since there are getter/setter, properties should be private and not protected @jordisala1991

jordisala1991:
> Since there are getter/setter, properties should be private and not protected @jordisala1991 

I will take a look at that too.

PR #2192, opened 2021-10-31
Use FilesystemInterface from Gaufrette

jordisala1991:
## Subject

## Changelog

```markdown
### Changed
- Use `FilesystemInterface` instead of concrete class from Gaufrette
```


phansys (approved):

jordisala1991:
Unfortunately the interface does not contain all the methods and there are no other interfaces. I would prefer to avoid method exists or other tricks, and since we plan to remove gaufrette on 5.0 we might not care a lot about this.

PR #2193, opened 2021-10-31
Remove dependency with doctrine/dbal

jordisala1991:
## Subject

The dependency is not really required, and that method should/will be deprecated on next major of doctrine extensions (because it is coupled with the orm, I couldn't find the getConnection on the document manager)

To me it is pedantic, because this only affects on the conflict section (for our users), for us it means not having to deal with one package here which is nice.

### DIFF
--- a/src/Model/NoDriverGalleryManager.php
+++ b/src/Model/NoDriverGalleryManager.php
@@ -13,7 +13,6 @@ declare(strict_types=1);
 
 namespace Sonata\MediaBundle\Model;
 
-use Doctrine\DBAL\Connection;
 use Sonata\MediaBundle\Exception\NoDriverException;
 
 /**
@@ -68,7 +67,7 @@ final class NoDriverGalleryManager implements GalleryManagerInterface
         throw new NoDriverException();
     }
 
-    public function getConnection(): Connection
+    public function getConnection()
     {
         throw new NoDriverException();
     }
--- a/src/Model/NoDriverGalleryManager.php
+++ b/src/Model/NoDriverGalleryManager.php
@@ -67,6 +67,12 @@ final class NoDriverGalleryManager implements GalleryManagerInterface
         throw new NoDriverException();
     }
 
+    /**
+     * Do not add return typehint to this method, it forces a dependency with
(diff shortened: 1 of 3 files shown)


phansys:
>I couldn't find the getConnection on the document manager

Indeed, the method was [removed in version 2.x](
https://github.com/doctrine/mongodb-odm/compare/1.3.7...2.0.0#diff-12a2324f373722122b496af69d59ee2b3a19ac4d50b37511973397ed707e50f1L223).

src/Model/NoDriverGalleryManager.php
```diff
@@ -68,7 +67,7 @@ public function getTableName(): string
         throw new NoDriverException();
     }
 
-    public function getConnection(): Connection
+    public function getConnection()
```
phansys, line 76: IMO, we should add a `NEXT_MAJOR` comment explaining the rationale about the missing return docblock or type declaration.

jordisala1991: I didn't add one because we will be able to change this when we want, not only on NEXT_MAJOR. This class is internal (and the other one too).

The rationale about this change is:

1. The retun typehint is not needed because the parent is not adding it.
2. We prefer not to add the typehint because that couples us with doctrine dbal and it is only used here
3. This method should also dissapear from the interface of the doctrine extension package because it is only needed for orm (and it can be solved in another way too, there is no need to have it)

Do you agree? want me to try a comment explaining that (not related to any NEXT_MAJOR tho)?

jordisala1991: Let me know what do you think of the message I added.

phansys (approved):

VincentLanglet (approved):
What about adding the deprecation to the doctrine extension repository ?

jordisala1991:
> What about adding the deprecation to the doctrine extension repository ?

Yep, That has to be done indeed.

VincentLanglet:
> > What about adding the deprecation to the doctrine extension repository ?
> 
> Yep, That has to be done indeed.

Just discovered I've already deprecated it in https://github.com/sonata-project/sonata-doctrine-extensions/pull/365

PR #2194, opened 2021-10-31
Mark method from abstract classes as final

jordisala1991:
## Subject

This change will help with #2191 . There are methods on that PR that are not present on this one, so this PR only focuses on the finals that will be useful for the 4.x version (there is no need to mark something final that in 4.x is removed).

## Changelog

```markdown
### Changed
- Added final annotation to methods that will be final on 4.x.
```

### DIFF
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -62,51 +62,81 @@ abstract class Gallery implements GalleryInterface, GalleryMediaCollectionInterf
         return $this->getName() ?? '-';
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function setName($name)
     {
         $this->name = $name;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function getName()
     {
         return $this->name;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function setEnabled($enabled)
     {
         $this->enabled = $enabled;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function getEnabled()
     {
         return $this->enabled;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function setUpdatedAt(?\DateTime $updatedAt = null)
     {
         $this->updatedAt = $updatedAt;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function getUpdatedAt()
     {
         return $this->updatedAt;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function setCreatedAt(?\DateTime $createdAt = null)
     {
         $this->createdAt = $createdAt;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
     public function getCreatedAt()
     {
         return $this->createdAt;
     }
 
+    /**
+     * @final since sonata-project/media-bundle 3.x
+     */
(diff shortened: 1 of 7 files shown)


jordisala1991:
Not sure what kind of label or changelog is needed for this change. @VincentLanglet wdyt?

jordisala1991:
I will take care of this merge to 4.x, it will be a mess but I can do it.

phansys:
>Not sure what kind of label or changelog is needed for this change. @VincentLanglet wdyt?

IMO, these changes require a MINOR release, and the changelog label should be "Changed", since these changes are marking existing methods as final.

phansys (dismissed):

phansys:
Additionally, I think this PR also deserves an upgrade note, since these docblocks generate deprecation messages if these methods are overridden.

jordisala1991:
> Additionally, I think this PR also deserves an upgrade note, since these docblocks generate deprecation messages if these methods are overridden.

Added.

phansys (approved):

VincentLanglet (approved):

src/Model/Gallery.php
```diff
@@ -62,51 +62,81 @@ public function __toString()
         return $this->getName() ?? '-';
     }
 
+    /**
```
ChristianGripp, line 65: We should not do this for model classes because they are more likely to be customised and overridden.

core23 (dismissed):

src/Model/Gallery.php
jordisala1991: I dont know what kind of customization is needed for a getter and a setter, can you provide an example?

jordisala1991: Also, if some user needs to customize the methods, he can create a class from the Interfaces, no need to use this Model.

VincentLanglet: Or at least open an issue with an explanation of the needs, and we can still remove the final if we find it's justified.

It's easier to add final now because adding final is not BC but removing final is.

ChristianGripp: e.g. you want to update the `updatedAt` or `enabled` state when setting some property.

> Also, if some user needs to customize the methods, he can create a class from the Interfaces, no need to use this Model.

Yes you could create some custom class, but that's not the idea of reusable models. With the same logic we could make the entity classes final.

jordisala1991: You could do that on the prePersist or preUpdate (which are not final on purpose)

jordisala1991: As @VincentLanglet said, what about making them final and if someone complains on a concrete use case, we can change it because it will be BC. For your example, using the prePersist could solve the issue (and if you really need to modify all the methods, you might as well create from the interface directly). wdyt?

VincentLanglet: By the way we should make the decision consistent between the Classification and the Media bundle about entities

jordisala1991: IMO it should be as we did for any other final TBH, it is final until it can be proven that there is a real need that cannot be solved in other ways.

When checking for final, we should ensure we added it for classes and for abstract classes.

That being said, the more time we spend on this, the more we delay releases and working on other bundles...

VincentLanglet: > IMO it should be as we did for any other final TBH, it is final until it can be proven that there is a real need that cannot be solved in other ways.
> 
> When checking for final, we should ensure we added it for classes and for abstract classes.

I agree

ChristianGripp: Still not a fan, but feel to merge this if I'm the only one with this considerations.

PR #2196, opened 2021-11-01
Make phpstan more strict

jordisala1991:
We should use those defaults on all projects

## Changelog

```markdown
### Changed
- Moved `__construct()` method from Entity/Document to Model.
```

### DIFF
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -36,6 +36,11 @@ abstract class Gallery implements GalleryInterface
      */
     protected Collection $galleryItems;
 
+    public function __construct()
+    {
+        $this->galleryItems = new ArrayCollection();
+    }
+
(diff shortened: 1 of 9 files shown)


VincentLanglet (approved):

core23 (approved):

src/Model/Gallery.php
```diff
@@ -36,6 +36,11 @@ abstract class Gallery implements GalleryInterface
      */
     protected Collection $galleryItems;
 
+    public function __construct()
+    {
+        $this->galleryItems = new ArrayCollection();
```
ChristianGripp, line 41: We should add some kind of changelog entry for this.

jordisala1991: Done

PR #2204, opened 2021-11-07
Make admin dependency optional for blocks

core23:
## Subject

This is a patch for the upcoming stable release. As the admin bundle is an optional dependency, it should be possible to render the blocks without it.

## Changelog

```markdown
### Changed
- Make "sonata-project/admin-bundle" dependency optional for blocks
```

### DIFF
--- a/src/Block/FeatureMediaBlockService.php
+++ b/src/Block/FeatureMediaBlockService.php
@@ -43,19 +43,19 @@ final class FeatureMediaBlockService extends AbstractBlockService implements Edi
     private Pool $pool;
 
     /**
-     * @var AdminInterface<MediaInterface>
+     * @var AdminInterface<MediaInterface>|null
      */
-    private AdminInterface $mediaAdmin;
+    private ?AdminInterface $mediaAdmin;
 
     private MediaManagerInterface $mediaManager;
 
     /**
-     * @param AdminInterface<MediaInterface> $mediaAdmin
+     * @param AdminInterface<MediaInterface>|null $mediaAdmin
      */
     public function __construct(
         Environment $twig,
         Pool $pool,
-        AdminInterface $mediaAdmin,
+        ?AdminInterface $mediaAdmin,
         MediaManagerInterface $mediaManager
     ) {
         parent::__construct($twig);
@@ -220,6 +220,10 @@ final class FeatureMediaBlockService extends AbstractBlockService implements Edi
 
     private function getMediaBuilder(FormMapper $form): FormBuilderInterface
     {
+        if (null === $this->mediaAdmin) {
+            throw new \LogicException('The SonataAdminBundle is required to render the edit form.');
+        }
+
         $fieldDescription = $this->mediaAdmin->createFieldDescription('media', [
             'translation_domain' => 'SonataMediaBundle',
             'edit' => 'list',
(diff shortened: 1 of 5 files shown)


src/Block/FeatureMediaBlockService.php
```diff
@@ -43,7 +43,7 @@ final class FeatureMediaBlockService extends AbstractBlockService implements Edi
     private Pool $pool;
 
     /**
-     * @var AdminInterface<MediaInterface>
+     * @var AdminInterface<MediaInterface>|null
      */
     private AdminInterface $mediaAdmin;
```
jordisala1991: you will need to declare it nullable

VincentLanglet (dismissed):

tests/Block/FeatureMediaBlockServiceTest.php
```diff
@@ -30,6 +30,7 @@ protected function setUp(): void
         $this->blockService = new FeatureMediaBlockService(
             $this->twig,
             new Pool('default'),
+            /* @phpstan-ignore-next-line */
```
jordisala1991: Why do you need to ignore it?

jordisala1991 (commented):
Also, dont you need to change dependency injection to pass null on invalid?

tests/Block/FeatureMediaBlockServiceTest.php
ChristianGripp: Because phpstan complains about the missing generic information. AFAIK this is not possible for phpunit mocks.

VincentLanglet: I dunno if it's different when doing 
```
$this->admin = $this->createStub(AdminInterface::class);
```
when admin is a property with a `@var` phpdoc.

The other option would be to do 
```
/** @var ... */
$admin = $this->createStub(AdminInterface::class);
```

jordisala1991: To me it looks better with the var thing instead of ignoring the line, which seems like a todo or a place to come later and fix something

VincentLanglet: I agree

wbloszyk (dismissed):

src/Block/MediaBlockService.php
```diff
@@ -209,6 +209,10 @@ private function getFormatChoices(?MediaInterface $media = null): array
 
     private function getMediaBuilder(FormMapper $form): FormBuilderInterface
     {
+        if (null === $this->mediaAdmin) {
+            throw new \LogicException('The SonataAdminBundle is required to render the edit form');
```
phansys:
```suggestion
            throw new \LogicException('The SonataAdminBundle is required to render the edit form.');
```

src/Block/FeatureMediaBlockService.php
```diff
@@ -220,6 +220,10 @@ private function getFormatChoices(?MediaInterface $media = null): array
 
     private function getMediaBuilder(FormMapper $form): FormBuilderInterface
     {
+        if (null === $this->mediaAdmin) {
+            throw new \LogicException('The SonataAdminBundle is required to render the edit form');
```
phansys:
```suggestion
            throw new \LogicException('The SonataAdminBundle is required to render the edit form.');
```

phansys (dismissed):

phansys:
Reading the "Fixed" section in the changelog, its hard to understand what this PR is fixing. IMO, the description sounds more accurate for the "Changed" or the "Removed" sections.

jordisala1991 (approved):

phansys (approved):

VincentLanglet (approved):

VincentLanglet:
Thanks

PR #2226, opened 2021-11-27
Manual merge

phansys:


src/Admin/GalleryAdmin.php
```diff
@@ -22,7 +22,7 @@
 use Symfony\Component\Form\Extension\Core\Type\ChoiceType;
 
 /**
- * @phpstan-extends AbstractAdmin<\Sonata\MediaBundle\Model\GalleryInterface>
+ * @phpstan-extends AbstractAdmin<\Sonata\MediaBundle\Model\GalleryInterface<\Sonata\MediaBundle\Model\GalleryItemInterface>>
```
jordisala1991, line 25: Why on the media admin you added the template but not here? It looks strange

VincentLanglet: And then the MediaAdmin is loosing the template from the BaseMedia so I would say it lost his interest.
Im not sure we should introduce all those generic if noone need them for now.

phansys: I've removed the generic in `MediaAdmin`, as I think we can make that move in a different PR if it is required later.
Thank you for pointing.

core23 (dismissed):

src/Model/GalleryInterface.php
```diff
@@ -15,6 +15,9 @@
 
 use Doctrine\Common\Collections\Collection;
 
+/**
+ * @phpstan-template T of GalleryItemInterface
```
VincentLanglet, line 19: This template is useless if it's not use anywhere else.

Either it should be removed (And if we remove it it's not needed in the BaseGallery, Gallery, ...) or be used in the phpdoc of 
```
public function setGalleryItems(Collection $galleryItems): void;
    public function getGalleryItems(): Collection;
    public function addGalleryItem(GalleryItemInterface $galleryItem): void;
    public function removeGalleryItem(GalleryItemInterface $galleryItem): void;
```

VincentLanglet (changes requested):

VincentLanglet (approved):

jordisala1991 (approved):

PR #2227, opened 2021-11-27
Allow `iterable` as argument 1 for `MediaInterface::setGalleryItems()` and `GalleryInterface::setGalleryItems()`

phansys:
## Subject
Allow `iterable` as argument 1 for `MediaInterface::setGalleryItems()` and `GalleryInterface::setGalleryItems()`.

I am targeting this branch, because here is where the changed type declarations were added.

## Changelog

```markdown
### Changed
- Argument 1 in order to allow any iterable type at `MediaInterface::setGalleryItems()` and `GalleryInterface::setGalleryItems()`.
```

### DIFF
--- a/src/Model/Gallery.php
+++ b/src/Model/Gallery.php
@@ -106,9 +106,9 @@ abstract class Gallery implements GalleryInterface
         return $this->defaultFormat;
     }
 
-    final public function setGalleryItems(Collection $galleryItems): void
+    final public function setGalleryItems(iterable $galleryItems): void
     {
-        $this->galleryItems = new ArrayCollection();
+        $this->galleryItems->clear();
 
         foreach ($galleryItems as $galleryItem) {
             $this->addGalleryItem($galleryItem);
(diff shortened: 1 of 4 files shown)


src/Model/Gallery.php
```diff
@@ -106,9 +106,9 @@ final public function getDefaultFormat(): string
         return $this->defaultFormat;
     }
 
-    final public function setGalleryItems(Collection $galleryItems): void
+    final public function setGalleryItems(iterable $galleryItems): void
```
jordisala1991, line 109: Not sure about that, an array is iterable right? But if I pass array, that clear method wont work

phansys: The `clear()` method will be executed on `ArrayCollection` (the value set for the `galleryItems` property at `__construct()`), not in the array provided as argument.

VincentLanglet (approved):

src/Model/Gallery.php
VincentLanglet: Should we add a test ?

phansys: I think these changes are covered by PHPStan, but I can add a test if you consider it is required.

VincentLanglet: Indeed i think you're right

jordisala1991 (approved):

PR #2263, opened 2022-01-16
Improve functional tests

jordisala1991:
## Subject

I am targeting this branch, because this branch is where the functional tests are done.


jordisala1991:
Detecting trying the features of this bundle on the 4.x branch. Not sure why it happens or why it wasn't happening before. Probably the same bug exists for 3.x since: https://github.com/sonata-project/SonataMediaBundle/pull/2035

The code has probably never work for executing pre and post remove code for entities with relations, but before it didn't crash, now it does if we do not call the load.

Not sure what happens with the ODM, but it should work different because it does not use Proxies the same way.

src/Provider/BaseVideoProvider.php
```diff
@@ -179,11 +179,11 @@ protected function getMetadata(MediaInterface $media, string $url)
     {
         try {
             $response = $this->sendRequest('GET', $url);
-        } catch (\RuntimeException $e) {
+        } catch (\RuntimeException $exception) {
             throw new \RuntimeException(
                 sprintf('Unable to retrieve the video information for: %s', $url),
-                \is_int($e->getCode()) ? $e->getCode() : 0,
-                $e
+                $exception->getCode(),
```
VincentLanglet: For the psalm issue, you'll have to add it to the psalm baseline for now

jordisala1991:
I need to test more because I cant make a test fail.

jordisala1991:
In fact I can't reproduce the problem with the gallery or the gallery item on my project, only with a custom entity that has a relation with media.

Do you know what could happen @VincentLanglet ?

VincentLanglet:
> In fact I can't reproduce the problem with the gallery or the gallery item on my project, only with a custom entity that has a relation with media.
> 
> Do you know what could happen @VincentLanglet ?

Not sure to understand the issue and I don't know a lot with proxy, but I often get issue with proxy when it's about relation.
Something like when Foo is fully load, `Foo->getBar` is a Proxy, and is only loaded as soon as Foo->getBar->getBaz` is done.

jordisala1991:
I was trying this bundle on a project I have and I faced several issues (I will try to fix all of them)

One of it was trying to create an object through the admin, and that object relates to a media. The I click on edit and delete and it gives me an error saying that I cant get a Provider without a provider name. Debugging the code I ended up on the preRemove where , for some reason, The media entity was there but only with the id, I could call all its methods and none of them returned anything but null. I found that issue on the orm, and solved with a load when it is not initialized.

The thing is I dont seem to be able to reproduce this on Sonata Gallery or GalleryItem, and it should be the same.

Maybe there is another problem because not only I was unable to delete, but the media didnt show on the related entity... I don't know how to reproduce it yet, but its weird, Sonata Media 3 was working (not tested with latest patches)

tests/App/Entity/Foo.php
```diff
@@ -0,0 +1,54 @@
...
+namespace Sonata\MediaBundle\Tests\App\Entity;
+
+use Doctrine\ORM\Mapping as ORM;
+use Sonata\MediaBundle\Model\MediaInterface;
+
+/**
+ * @ORM\Entity
+ * @ORM\Table(name="media__foo")
+ */
+class Foo
```
jordisala1991: This is how I have my entity in my project and it fails when deleting, but I can't make it fail on tests. Not sure how to configure doctrine to use proxies, it should be enabled by default right? but on tests I always receive the real object, not the proxy. (I tried clearing the entity manager or removing the object from the unit of work, but nothing works.

VincentLanglet: Maybe @greg0ire can help

VincentLanglet:
Now the final keyword is removed, should we close this ?

jordisala1991:
> Not the final keyword is removed, should we close this ?

I will try to keep the test and improve them to have better coverage

jordisala1991:
wdyt? Might be too much functional tests, but this way we test a lot of templates, configurations and admins.

composer.json
```diff
@@ -65,11 +65,13 @@
     },
     "require-dev": {
         "aws/aws-sdk-php": "^3.0",
+        "dama/doctrine-test-bundle": "^6.7",
         "doctrine/doctrine-bundle": "^2.3.2",
         "doctrine/mongodb-odm": "^2.2",
         "doctrine/orm": "^2.9",
         "knplabs/knp-menu-bundle": "^3.0",
         "liip/imagine-bundle": "^2.0",
+        "masterminds/html5": "^2.7",
```
jordisala1991, line 74: Needed to correctly parse some html with a lot of javascript

VincentLanglet (approved):

PR #2264, opened 2022-01-19
Avoid using final for getters/setters

### LINKED ISSUE #2232: images dont render
Today i make composer upgrade and after that images disappeared. When i dump media object i get
```
Proxies\__CG__\App\Entity\Media\SonataMediaMedia {#19025 ▼
  +__isInitialized__: false
  #id: 8
  #name: null
  #description: null
  #enabled: false
  #providerName: null
  #providerStatus: null
  #providerReference: null
  #providerMetadata: []
  #width: null
  #height: null
  #length: null
  #copyright: null
  #authorName: null
  #context: null
  #cdnStatus: null
  #cdnIsFlushable: false
  #cdnFlushIdentifier: null
  #cdnFlushAt: null
  #updatedAt: null
  #createdAt: null
  #binaryContent: null
  #previousProviderReference: null
  #contentType: null
  #size: null
  #category: null
   …2
}
```
And if i tried to upload image i get 'error.image_too_small'
media-bundle version - 4.0.0-RC2

jordisala1991:
For some reason it messes up with doctrine proxy loading

## Subject

I am targeting this branch, because this is a bug fix.

Closes #2232 

## Changelog

```markdown
### Removed
- Final modifier on all the getters / setters of the entities.

### Fixed
- Loading Media / Gallery object from the database.
```

### DIFF
--- a/psalm-baseline.xml
+++ b/psalm-baseline.xml
@@ -1,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<files psalm-version="v4.15.0@a1b5e489e6fcebe40cb804793d964e99fc347820">
+    <file src="src/Provider/BaseVideoProvider.php">
+        <InvalidScalarArgument occurrences="1">
+            <code>$exception-&gt;getCode()</code>
+        </InvalidScalarArgument>
+    </file>
+</files>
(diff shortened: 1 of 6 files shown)


jordisala1991:
Also for more information:

https://www.doctrine-project.org/projects/doctrine-orm/en/2.9/reference/architecture.html#entities

> An entity class must not be final or contain final methods.

VincentLanglet:
Phpstan have a `EntityNotFinal` rule https://github.com/phpstan/phpstan-doctrine/blob/master/src/Rules/Doctrine/ORM/EntityNotFinalRule.php

Can you add phpstan/phpstan-doctrine then ?

I think it only check final class, but we could request a feature for final method too.

core23 (approved):

jordisala1991:
> Phpstan have a `EntityNotFinal` rule https://github.com/phpstan/phpstan-doctrine/blob/master/src/Rules/Doctrine/ORM/EntityNotFinalRule.php
> 
> Can you add phpstan/phpstan-doctrine then ?
> 
> I think it only check final class, but we could request a feature for final method too.

It does not work for methods. Maybe it is not worth adding in this PR since it cannot detect the current issue, but can be good at a later point if they introduce for final methods too. (I have installed and it does not detect anything new, and tried to add final  on methods)

PR #2267, opened 2022-01-20
Use jsonb to avoid problems on batch delete

jordisala1991:
## Subject

I am targeting this branch, because this is the safe branch to do it.

## Changelog

```markdown
### Fixed
- Fixed usage of this bundle with PostgreSQL.
```

### DIFF
--- a/src/Resources/config/doctrine/BaseMedia.orm.xml
+++ b/src/Resources/config/doctrine/BaseMedia.orm.xml
@@ -7,7 +7,11 @@
         <field name="providerName" column="provider_name" type="string" length="255"/>
         <field name="providerStatus" column="provider_status" type="integer"/>
         <field name="providerReference" column="provider_reference" type="string" length="255"/>
-        <field name="providerMetadata" column="provider_metadata" type="json" nullable="true"/>
+        <field name="providerMetadata" column="provider_metadata" type="json" nullable="true">
+            <options>
+                <option name="jsonb">true</option>
+            </options>
+        </field>
         <field name="width" column="width" type="integer" nullable="true"/>
         <field name="height" column="height" type="integer" nullable="true"/>
         <field name="length" column="length" type="decimal" nullable="true"/>
--- a/UPGRADE-4.0.md
+++ b/UPGRADE-4.0.md
@@ -12,6 +12,9 @@ Assuming your media table is called `media__media`, you could execute:
 UPDATE media__media SET cdnIsFlushable = false WHERE cdnIsFlushable IS NULL;
 ```
 
+We also added an option to use `jsonb` for the `providerMetadata` field. This will only affect you if you use PostgreSQL
+database. Please make sure you do the appropiate changes to your schema in order to be in sync.
+
 ### Removed unused JavaScript
 
 SonataMediaBundle 3 included a JavaScript dependency: Nivo Gallery. This dependency is not used on Sonata code,


jordisala1991:
@lukepass do you use MySQL? if so, can you check if this PR breaks your database schema in any way?

jordisala1991:
Maybe it is better to just add the jsonb option to not do more hard BC breaks. @VincentLanglet 

(we have more usages of json fields in SonataPageBundle)
(The roles field will become an array again , that was a mistake when removing FOSUserBundle)

lukepass:
Should I just go on this branch and launch migrations or make some media upload?

jordisala1991:
Yes please, go to this branch and try to launch migrations. Also if you could remove your media table and try to update the schema again, to ensure nothing changes, would be helpful.

Thank you :)

src/Resources/config/doctrine/BaseMedia.orm.xml
```diff
@@ -7,7 +7,11 @@
         <field name="providerName" column="provider_name" type="string" length="255"/>
         <field name="providerStatus" column="provider_status" type="integer"/>
         <field name="providerReference" column="provider_reference" type="string" length="255"/>
-        <field name="providerMetadata" column="provider_metadata" type="json" nullable="true"/>
+        <field name="providerMetadata" column="provider_metadata" type="json" nullable="true">
+            <options>
+                <option name="jsonb">true</option>
```
ChristianGripp, line 12: Should we add tests for this?

core23 (approved):
Did not test this, but looks okay so far

core23:
IMHO we can release the stable 4.0 release after merging this. The last RC release was in November.

VincentLanglet:
Yes, I just wait for someone, like lukepass, to test this.

lukepass:
Yes! Sorry monday I will test it. Thanks!

Il sab 22 gen 2022, 19:06 Vincent Langlet ***@***.***> ha
scritto:

src/Resources/config/doctrine/BaseMedia.orm.xml
jordisala1991: I don't think it is really easy to test. The problem with the distinct query is only when you do a batch delete.

I think we might start with testing on different databases instead of trying to reproduce the problem that this PR solves.

Anyway we shouldn't block the PR for its test IMO. I guess if it does not break with Sqlite we are fine.

VincentLanglet: I personally don't block the PR for unit/functional test, but for real tests.

I'd like to avoid to release the 4.0 with a major bug for mysql user and then be in the situation where reverting the change introduce a bug for mongo users.

lukepass:
It generated a migration:

```
<?php

declare(strict_types=1);

namespace DoctrineMigrations;

use Doctrine\DBAL\Schema\Schema;
use Doctrine\Migrations\AbstractMigration;

/**
 * Auto-generated Migration: Please modify to your needs!
 */
final class Version20220124085858 extends AbstractMigration
{
    public function getDescription(): string
    {
        return '';
    }

    public function up(Schema $schema): void
    {
        // this up() migration is auto-generated, please modify it to your needs
        $this->addSql('ALTER TABLE media__media CHANGE provider_metadata provider_metadata JSON DEFAULT NULL');
    }

    public function down(Schema $schema): void
    {
        // this down() migration is auto-generated, please modify it to your needs
        $this->addSql('ALTER TABLE media__media CHANGE provider_metadata provider_metadata JSON DEFAULT NULL');
    }
}
```

I executed it and made some tests, it seems all ok! The schema didn't break and the previous data didn't corrupt.

VincentLanglet (approved):

PR #2268, opened 2022-01-20
DevKit updates for 4.x branch

SonataCI:


VincentLanglet:
We need to add `doctrine_test` in the dev-kit config.

But since it's a project config (and not branch) we need to install it in 3.x even if it's not used.

PR #2271, opened 2022-01-25
Backport ImageProviderInterface from 4.x

jorrit:
## Subject

I am preparing to migrate to Symfony 5 and all new Sonata bundle, including Media Bundle 4. One issue is that I have a class that inherits from `ImageProvider`. This is no longer possible in 4.0, where a new interface `ImageProviderInterface` is introduced. It would be convenient if I could migrate my implementation to this new interface before migrating every Sonata/Symfony package.

I am targeting this branch, because it backports a feature from 4.0.

## Changelog

```markdown
### Added
- Added `ImageProviderInterface` so `ImageProvider` can be replaced without extending the class.
```

### DIFF
--- a/src/Validator/Constraints/ImageUploadDimensionValidator.php
+++ b/src/Validator/Constraints/ImageUploadDimensionValidator.php
@@ -15,7 +15,7 @@ namespace Sonata\MediaBundle\Validator\Constraints;
 
 use Imagine\Image\ImagineInterface;
 use Sonata\MediaBundle\Model\MediaInterface;
-use Sonata\MediaBundle\Provider\ImageProvider;
+use Sonata\MediaBundle\Provider\ImageProviderInterface;
 use Symfony\Component\Validator\Constraint;
 use Symfony\Component\Validator\ConstraintValidator;
 use Symfony\Component\Validator\Exception\UnexpectedTypeException;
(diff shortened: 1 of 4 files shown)


src/Validator/Constraints/ImageUploadDimensionValidator.php
```diff
@@ -28,11 +28,11 @@ final class ImageUploadDimensionValidator extends ConstraintValidator
     private $imagineAdapter;
 
     /**
-     * @var ImageProvider
+     * @var ImageProviderInterface
      */
     private $imageProvider;
 
-    public function __construct(ImagineInterface $imagineAdapter, ImageProvider $imageProvider)
+    public function __construct(ImagineInterface $imagineAdapter, ImageProviderInterface $imageProvider)
```
jordisala1991, line 35: Not sure but this change looks like BC break to me.

jorrit: I can only think of one situation, namely when someone extends the validator and its constructor.

VincentLanglet: Why wouldn't it be BC @jordisala1991 ?

jorrit: I could also remove the explicit parameter type and check it at runtime.

VincentLanglet (approved):

core23 (approved):

VincentLanglet:
If it's okay for you @jordisala1991 we'll merge this

src/Validator/Constraints/ImageUploadDimensionValidator.php
jordisala1991: If it is BC, fine by me :)

PR #2299, opened 2022-03-31
Use lowercase for extension and mimetype

### LINKED ISSUE #2297: The image extension "JPG" is not one of the allowed
#### Sonata packages
<details>
<p>
```
sonata-project/media-bundle  3.38.1 
```
</p>
</details>

## Subject
The image extension "JPG" is not one of the allowed ("jpg", "png", "jpeg", "gif").

## Steps to reproduce
Upload file with JPG extension in name.

## Expected results
Restore strtolower() before allowed extensions check. It has been removed in last [commit](https://github.com/sonata-project/SonataMediaBundle/commit/e875289e0325bec8b75910867b6b449ceda00f91#diff-67180f223f1bd847e875cb4b6fc8e2df5bb0dfe73306b9c5a1ccf94075b67c0bL231)

It has been removed also in version 4.x and 5.x.

## Actual results
The image extension "JPG" is not one of the allowed.

VincentLanglet:
## Subject

I am targeting this branch, because patch (?).

Closes #2297.

I added the missing strtolowercase back @jordisala1991, but I also added a transformer on the config WDYT about it ?

## Changelog
```markdown
### Fixed
- Check for allowed extensions and mimetypes are case insensitive.
```

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -290,6 +290,12 @@ final class Configuration implements ConfigurationInterface
                                 ->scalarNode('generator')->defaultValue('sonata.media.generator.default')->end()
                                 ->scalarNode('thumbnail')->defaultValue('sonata.media.thumbnail.file')->end()
                                 ->arrayNode('allowed_extensions')
+                                    ->beforeNormalization()
+                                    ->ifArray()
+                                        ->then(static function (array $allowedExtensions): array {
+                                            return array_map('strtolower', $allowedExtensions);
+                                        })
+                                    ->end()
                                     ->prototype('scalar')->end()
                                     ->defaultValue([
                                         'pdf', 'txt', 'rtf',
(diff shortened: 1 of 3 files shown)


jordisala1991:
Do you think this is needed for mime types? I never saw upper case for that metadata

VincentLanglet:
> Do you think this is needed for mime types? I never saw upper case for that metadata

I'm not sure it will be needed indeed
https://developer.mozilla.org/es/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

If I add the strlowercase on the extension,
I thought it could be better to add a normalizer on the allowed_extensions config then.

But then, if we normalize the config for allowed_extensions, it could be natural we do the same for mime type.

The only question could be about the call inside the FileProvider and ImageProvider.
When looking on google about uppercase mimetype, I found message like this https://lists.freedesktop.org/archives/xdg/2013-March/012711.html

So I thought it was safer...

VincentLanglet:
Is it ok for you @jordisala1991 ?

jordisala1991 (approved):

PR #2328, opened 2022-08-04
Correction sf6

MahefaAbel:
## Subject

I am targeting this branch, because there are an error when using this package on symfony 6.

Closes #{put_issue_number_here}.

## Changelog

```markdown
### Added
- Added condition $provider['resizer'] when it is not string (do not have a value)

### Changed

### Deprecated

### Removed

### Fixed

### Security
```


VincentLanglet:
Why would the resizer be false ?
Isnt it a bad config instead ?
```
resizer: false
```
instead of 
```
resizer: ~
```
for instance ?

jordisala1991:
Yes, I think the issue is because the config needs to be updated when upgrading from v3.

PR #2347, opened 2022-11-02
Fix phpstan if not ClassificationBundle not installed

### LINKED ISSUE #2344: phpstan throws setCategory() has invalid type
#### Sonata packages

<details><summary>show</summary>
<p>

```
$ composer show --latest 'sonata-project/*'
Direct dependencies required in composer.json:
sonata-project/admin-bundle              4.20.0 4.20.0 The missing Symfony Admin Generator
sonata-project/doctrine-orm-admin-bundle 4.8.0  4.8.0  Integrate Doctrine ORM into the SonataAdminBundle
sonata-project/exporter                  2.13.0 3.0.0  Lightweight Exporter library
sonata-project/media-bundle              4.5.1  4.5.1  Symfony SonataMediaBundle

Transitive dependencies not required in composer.json:
sonata-project/block-bundle              4.19.0 4.19.0 Symfony SonataBlockBundle
sonata-project/cache                     2.2.0  2.2.0  Cache library
Package sonata-project/cache is abandoned, you should avoid using it. No replacement was suggested.
sonata-project/doctrine-extensions       2.0.1  2.0.1  Doctrine2 behavioral extensions
sonata-project/form-extensions           1.18.0 1.18.0 Symfony form extensions
sonata-project/twig-extensions           1.12.0 2.0.0  Sonata twig extensions

```

</p>
</details>

#### Symfony packages

<details><summary>show</summary>
<p>

```
Direct dependencies required in composer.json:
symfony/monolog-bundle           v3.8.0  v3.8.0  Symfony MonologBundle
symfony/phpunit-bridge           v4.4.47 v6.1.6  Provides utilities for PHPUnit, especially user deprecation notices management
symfony/symfony                  v4.4.47 v6.1.7  The Symfony PHP framework

Transitive dependencies not required in composer.json:
symfony/contracts                v1.1.13 v3.1.1  A set of abstractions extracted out of the Symfony components
symfony/deprecation-contracts    v3.1.1  v3.1.1  A generic function and convention to trigger deprecation notices
symfony/polyfill-ctype           v1.26.0 v1.26.0 Symfony polyfill for ctype functions
symfony/polyfill-intl-grapheme   v1.26.0 v1.26.0 Symfony polyfill for intl's grapheme_* functions
symfony/polyfill-intl-icu        v1.26.0 v1.26.0 Symfony polyfill for intl's ICU-related data and classes
symfony/polyfill-intl-idn        v1.26.0 v1.26.0 Symfony polyfill for intl's idn_to_ascii and idn_to_utf8 functions
symfony/polyfill-intl-normalizer v1.26.0 v1.26.0 Symfony polyfill for intl's Normalizer class and related functions
symfony/polyfill-mbstring        v1.26.0 v1.26.0 Symfony polyfill for the Mbstring extension
symfony/polyfill-php72           v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 7.2+ features to lower PHP versions
symfony/polyfill-php73           v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 7.3+ features to lower PHP versions
symfony/polyfill-php80           v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 8.0+ features to lower PHP versions
symfony/polyfill-php81           v1.26.0 v1.26.0 Symfony polyfill backporting some PHP 8.1+ features to lower PHP versions
symfony/security-acl             v3.3.2  v3.3.2  Symfony Security Component - ACL (Access Control List)
symfony/string                   v5.4.14 v6.1.7  Provides an object-oriented API to strings and deals with bytes, UTF-8 code points and grapheme clusters i...
```
</p>
</details>

#### PHP version

```
$ php -v
PHP 8.1.10 (cli) (built: Sep  1 2022 21:43:31) (NTS)
Copyright (c) The PHP Group
Zend Engine v4.1.10, Copyright (c) Zend Technologies
    with Zend OPcache v8.1.10, Copyright (c), by Zend Technologies
    with Xdebug v3.1.5, Copyright (c) 2002-2022, by Derick Rethans

```

## Steps to reproduce
- Create class implements `MediaInterface`, for example, `ExampleMedia`
- Create method `public function setCategory(?object $category = null): void`
- run phpstan analyse

## Expected results
no errors
## Actual results
`Parameter $category of method ExampleMedia::setCategory() has invalid type  
         Sonata\ClassificationBundle\Model\CategoryInterface.    `

This happened because I didn't install `sonata-project/classification-bundle`, because I don't need it.
What to do?

dmitryuk:
## Subject

Closes #2344

## Changelog
```markdown
### Changed
- `BaseMediaAdmin::__construct` changed signature
- `MediaInterface` methods `getCategory` and `setCategory` changed phpdoc
```
## To do

- [ ] Update the tests;
- [ ] Update the documentation;
- [ ] Add an upgrade note.

### DIFF
--- a/src/Admin/BaseMediaAdmin.php
+++ b/src/Admin/BaseMediaAdmin.php
@@ -40,44 +40,16 @@ abstract class BaseMediaAdmin extends AbstractAdmin
 
     protected $classnameLabel = 'Media';
 
-    /**
-     * NEXT_MAJOR: Change to (Pool, ?CategoryManagerInterface, ?ContextManagerInterface).
-     *
-     * @param Pool|string                          $pool
-     * @param CategoryManagerInterface|string|null $categoryManager
-     * @param ContextManagerInterface|string|null  $contextManager
-     *
-     * @phpstan-param CategoryManagerInterface|class-string<MediaInterface>|null $categoryManager
-     */
     public function __construct(
-        $pool,
-        $categoryManager = null,
-        $contextManager = null,
-        ?Pool $deprecatedPool = null,
-        ?CategoryManagerInterface $deprecatedCategoryManager = null,
-        ?ContextManagerInterface $deprecatedContextManager = null
+        Pool $pool,
+        ?CategoryManagerInterface $categoryManager = null,
+        ?ContextManagerInterface $contextManager = null
     ) {
-        // NEXT_MAJOR: Keep the if part.
-        if ($pool instanceof Pool) {
-            \assert(!\is_string($categoryManager));
-            \assert(!\is_string($contextManager));
+        parent::__construct();
 
-            parent::__construct();
-
-            $this->pool = $pool;
-            $this->categoryManager = $categoryManager;
-            $this->contextManager = $contextManager;
-        } else {
-            \assert(\is_string($categoryManager));
-            \assert(\is_string($contextManager));
-            \assert(null !== $deprecatedPool);
-
-            parent::__construct($pool, $categoryManager, $contextManager);
-
-            $this->pool = $deprecatedPool;
-            $this->categoryManager = $deprecatedCategoryManager;
-            $this->contextManager = $deprecatedContextManager;
-        }
+        $this->pool = $pool;
+        $this->categoryManager = $categoryManager;
+        $this->contextManager = $contextManager;
     }
 
     final public function getObjectMetadata(object $object): MetadataInterface
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -13,6 +13,7 @@ declare(strict_types=1);
 
 namespace Sonata\MediaBundle\DependencyInjection;
 
+use Sonata\ClassificationBundle\Model\CategoryInterface;
 use Sonata\Doctrine\Mapper\Builder\OptionsBuilder;
 use Sonata\Doctrine\Mapper\DoctrineCollector;
 use Symfony\Component\Config\Definition\Processor;
@@ -36,6 +37,10 @@ final class SonataMediaExtension extends Extension implements PrependExtensionIn
 
     public function load(array $configs, ContainerBuilder $container): void
     {
+        // Fix strict type if ClassificationBundle is not installed
+        if (!interface_exists(CategoryInterface::class)) {
+            class_alias(CategoryInterface::class, \stdClass::class);
+        }
         $processor = new Processor();
         $configuration = new Configuration();
         $config = $processor->processConfiguration($configuration, $configs);
--- a/src/Model/Media.php
+++ b/src/Model/Media.php
@@ -77,10 +77,7 @@ abstract class Media implements MediaInterface
      */
     protected Collection $galleryItems;
 
-    /**
-     * @var CategoryInterface|null
-     */
-    protected ?object $category = null;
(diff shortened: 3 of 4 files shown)


VincentLanglet (changes requested):
IMHO, that's not the right solution:
- Removing the phpdoc from getCategory and setCategory will allow any object to be get/set, but we don't support this.
- In the same way allowing `?object` instead of `?CategoryManager` in the construct of the Admin, will allow any object to be passed in 2nd and 3rd argument, and again we don't support it.

You're solving your issue because you're not using those method/arguments, but you're creating issues for people using it.

Did you open a phpstan discussion to see if solution are proposed for such situation https://github.com/phpstan/phpstan/discussions ? Because the php code is fully working atm.
So it's maybe something to do for phpstan (class alias ? https://phpstan.org/user-guide/discovering-symbols#class-aliases)

For the admin we might move the category-related logic to an admin extension.
For the entity, I dunno if it would be possible to have two classes/interfaces.

dmitryuk:
@VincentLanglet thanks for your answer, but:
- `getCategory` and `setCategory` before allowed to pass any object, phpdoc only do hints for IDE, but in this case it also didn't work:
image
- You  can pass any object to constructor, but it will throw an exception, because where are `asserts` checks

VincentLanglet:
> but in this case it also didn't work:
>   image
> * You  can pass any object to constructor, but it will throw an exception, because where are `asserts` checks

`assert` are generally throwing warning in development mode and doing nothing in production mode.
We can't rely on this. 

> @VincentLanglet thanks for your answer, but:
> 
> * `getCategory` and `setCategory` before allowed to pass any object

Even in 3.x it didn't support any object
cf: https://github.com/sonata-project/SonataMediaBundle/blob/3.x/src/Model/Media.php#L155-L166

> phpdoc only do hints for IDE

I never say I liked more this solution. But at least it never throw error on runtime.
PHPStan error is a small issue and you can still ignore it.

Moreover, if it's a *PHPStan* issue, it could be solved in phpstan.
I proposed:
- to open a discussion https://github.com/phpstan/phpstan/discussions
- to try class alias https://phpstan.org/user-guide/discovering-symbols#class-aliases
Did you try both ?

Also I proposed to move the Category-related code from the Admin to an AdminExtension. This should be doable with something like https://github.com/sonata-project/SonataMediaBundle/pull/2348 ; it would require to dynamically load the extension config in DependencyInjection/SonataMediaExtension if classification is enabled.
What do you think about this @jordisala1991 ?

For the MediaInterface, I dunno if we can solve the issue with two different interfaces.

dmitryuk:
@VincentLanglet please check the code again, I fixed it with class_alias.
I found that phpstan on `BaseMediaAdmin` don't throw an error if I don't pass `null` to arguments from `ClassificationBundle`

VincentLanglet:
> @VincentLanglet please check the code again, I fixed it with class_alias.

Glad to hear this, this could be an easy solution.

Can you reopen a PR on 4.x instead ?
We Might add class alias for
CategoryManagerInterface
ContextManagerInterface
too.

dmitryuk:
@VincentLanglet I can create two MRs:
1. 4.x - only add class_aliases
2. 5.x - change MediaInterface to use `CategoryInterface` as return type and argument; remove extra arguments from constructor of BaseMediaAdmin + add types

PR #2358, opened 2023-02-16
Configurable s3 adapter

### LINKED ISSUE #2357: Configurable Gaufrette S3 adapter class
## Feature Request

It would be very useful if I could set the Gaufrette S3 adapter class via the bundle configuration. Right now I can see that `Gaufrette\Adapter\AwsS3` is hardcoded in the `gaufrette.php`. But what if I wanted to use `Gaufrette\Adapter\AsyncAwsS3`?

I was trying to override the `sonata.media.adapter.filesystem.s3` service, but since it's not that easy to pass all needed arguments there (which are literally being built in the bundle's compiler pass), I didn't succeed. I could make a custom compiler pass in my code to redefine this service, but it looks overhead to me, if there was a possibility to change the class from the configs in the first place. I also couldn't find anything about it in the docs, [amazon s3 configuration reference](https://docs.sonata-project.org/projects/SonataMediaBundle/en/3.x/reference/amazon_s3/) just shows default configuration for the Gaufrette adapter in general, with no option to override the actual class.

So, is it possible to make the adapter class configurable? Or maybe there's an easy way to override it right now, so it's not necessary to add an additional configuration for this? Perhaps, I just didn't find an existing quick solution?

Thank you.

ozahorulia:
## Added configuration to set AWS S3 adapter class

I am targeting this branch, because it adds a configuration feature with full BC.

Closes #2357 .

## Changelog

```markdown
### Added
- Added new configuration parameter `sonata_media.filesystem.s3.async`
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,7 +107,8 @@
         "sonata-project/doctrine-orm-admin-bundle": "<4.0"
     },
     "suggest": {
-        "async-aws/simple-s3": "If you want to use async AWS S3 adapter",
+        "async-aws/simple-s3": "If you want to use async Amazon S3 adapter",
+        "aws/aws-sdk-php": "If you want to use Amazon S3",
         "liip/imagine-bundle": "If you want on-the-fly thumbnail generation or image filtering (scale, crop, watermark...).",
         "sonata-project/classification-bundle": "If you want to categorize your media items.",
         "sonata-project/doctrine-orm-admin-bundle": "If you want to persist entities.",
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "twig/twig": "^2.14 || ^3.0"
     },
     "require-dev": {
-        "async-aws/simple-s3" : "^1.1",
+        "async-aws/simple-s3": "^1.0",
         "aws/aws-sdk-php": "^3.0",
         "dama/doctrine-test-bundle": "^6.7",
         "doctrine/doctrine-bundle": "^2.3.2",
--- a/composer.json
+++ b/composer.json
@@ -64,6 +64,7 @@
         "twig/twig": "^2.14 || ^3.0"
     },
     "require-dev": {
+        "async-aws/simple-s3" : "^1.1",
         "aws/aws-sdk-php": "^3.0",
         "dama/doctrine-test-bundle": "^6.7",
         "doctrine/doctrine-bundle": "^2.3.2",
@@ -89,7 +90,6 @@
         "sonata-project/classification-bundle": "^4.0",
         "sonata-project/doctrine-orm-admin-bundle": "^4.0",
         "symfony/browser-kit": "^4.4 || ^5.4 || ^6.0",
-        "symfony/filesystem": "^4.4 || ^5.4 || ^6.0",
         "symfony/messenger": "^4.4 || ^5.4 || ^6.0",
         "symfony/phpunit-bridge": "^6.1",
         "symfony/security-csrf": "^4.4 || ^5.4 || ^6.0",
@@ -107,6 +107,7 @@
         "sonata-project/doctrine-orm-admin-bundle": "<4.0"
     },
     "suggest": {
+        "async-aws/simple-s3": "If you want to use async AWS S3 adapter",
         "liip/imagine-bundle": "If you want on-the-fly thumbnail generation or image filtering (scale, crop, watermark...).",
         "sonata-project/classification-bundle": "If you want to categorize your media items.",
         "sonata-project/doctrine-orm-admin-bundle": "If you want to persist entities.",
--- a/composer.json
+++ b/composer.json
@@ -96,6 +96,7 @@
         "vimeo/psalm": "^4.7.2"
     },
     "conflict": {
+        "async-aws/simple-s3": "<1.0",
         "aws/aws-sdk-php": "<3.0",
         "doctrine/mongodb-odm": "<2.2",
         "doctrine/orm": "<2.9",
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -111,7 +111,7 @@ Full configuration options:
                 cache_control: max-age=86400 # or any other
                 meta:
                     key1: value1 #any amount of metas(sent as x-amz-meta-key1 = value1)
-                async: true
+                async: false
 
             replicate:
                 primary: sonata.media.adapter.filesystem.s3
--- a/docs/reference/advanced_configuration.rst
+++ b/docs/reference/advanced_configuration.rst
@@ -111,6 +111,7 @@ Full configuration options:
                 cache_control: max-age=86400 # or any other
                 meta:
                     key1: value1 #any amount of metas(sent as x-amz-meta-key1 = value1)
+                async: true
 
             replicate:
                 primary: sonata.media.adapter.filesystem.s3
--- a/src/DependencyInjection/SonataMediaExtension.php
+++ b/src/DependencyInjection/SonataMediaExtension.php
@@ -266,9 +266,9 @@ final class SonataMediaExtension extends Extension implements PrependExtensionIn
         if ($container->hasDefinition('sonata.media.adapter.filesystem.s3') && isset($config['filesystem']['s3'])) {
             $async = true === $config['filesystem']['s3']['async'];
             if (true === $async && !class_exists(SimpleS3Client::class)) {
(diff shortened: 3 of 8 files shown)


ozahorulia:
On a real project it fails because of this error:

```
Gaufrette\Adapter\AsyncAwsS3::__construct(): Argument #1 ($service) must be of type AsyncAws\SimpleS3\SimpleS3Client, ContainerEmExVd8\S3Client_76b2022 given,
```

I have to think how to pass async client instead of the simple one

VincentLanglet:
> On a real project it fails because of this error:
> 
> ```
> Gaufrette\Adapter\AsyncAwsS3::__construct(): Argument #1 ($service) must be of type AsyncAws\SimpleS3\SimpleS3Client, ContainerEmExVd8\S3Client_76b2022 given,
> ```
> 
> I have to think how to pass async client instead of the simple one

Is there other adapters ?
If not, you may use a config `async: true/false` instead of the class.
And then if config = true, you define the async, else you define the original one.
This allows to have different constructors.

ozahorulia:
> Is there other adapters ? If not, you may use a config `async: true/false` instead of the class. And then if config = true, you define the async, else you define the original one. This allows to have different constructors.

Yes, this is exactly what I'm going to do. Since there're indeed no other adapters, `async` options looks much more useful. I'm working on it.

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -260,8 +262,15 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
 
         // add the default configuration for the S3 filesystem
         if ($container->hasDefinition('sonata.media.adapter.filesystem.s3') && isset($config['filesystem']['s3'])) {
+            $async = (bool) $config['filesystem']['s3']['async'];
```
VincentLanglet: Instead of casting, we are solving the phpstan issue with

True === $config...

docs/reference/advanced_configuration.rst
```diff
@@ -111,6 +111,7 @@ Full configuration options:
                 cache_control: max-age=86400 # or any other
                 meta:
                     key1: value1 #any amount of metas(sent as x-amz-meta-key1 = value1)
+                async: true
```
VincentLanglet: False for the default config doc

tests/Controller/MediaAdminControllerTest.php
```diff
@@ -128,6 +128,10 @@ public function testListAction(): void
         $this->configureSetFormTheme($formView, ['filterTheme']);
         $this->configureSetCsrfToken('sonata.batch');
         $this->configureRender('templateList', 'renderResponse');
+
+        /**
+         * @psalm-suppress DeprecatedClass
```
VincentLanglet: Can you solve the deprecation instead ? :)

src/Resources/config/gaufrette.php
```diff
@@ -75,4 +76,10 @@
             ->args([[]])
 
         ->set('sonata.media.metadata.noop', NoopMetadataBuilder::class);
+
+    if (class_exists(SimpleS3Client::class)) {
```
VincentLanglet, line 83: Is there an error without the class exist check ? Because we might lack of a class exist check for the non async S3Client then

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -276,22 +285,37 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
+                        'secret' => $config['filesystem']['s3']['secretKey'],
+                        'key' => $config['filesystem']['s3']['accessKey'],
+                    ];
+                }
             }
 
-            $container->getDefinition('sonata.media.adapter.service.s3')
-                ->replaceArgument(0, $arguments);
+            if ($async) {
+                $container->getDefinition('sonata.media.adapter.service.s3.async')
```
VincentLanglet, line 321: Am I wrong or the definition wont exist if the package is not installed ?

We should add an explicit exception "you need to install foo package to use async"

ozahorulia: I also just realized that for the full BC we must remain the old s3 client service even when async is turned on, because somebody might rely on that service outside of sonata context. So instead of selecting just one client, I'll configure both if async is turned on.

But I'm still adding errors for both simple and async adapters if the client doesn't exist, it makes sense since both clients are optional and not required by default.

src/Resources/config/gaufrette.php
ozahorulia: You are right

tests/Controller/MediaAdminControllerTest.php
ozahorulia: I'd leave it as it is right now, because phpunit [doesn't provide any replacement](https://github.com/sebastianbergmann/phpunit/issues/4026#issuecomment-825453794) for the `withConsecutive` method, and I'm way out of the context of this test case to thoroughly re-write the test. Any proposed workarounds, like using callbacks with match (switch if we still support 7.4) look very hideous to me.

src/DependencyInjection/SonataMediaExtension.php
VincentLanglet: I think you don't keep the old s3 client for BC reason because the developer has to set manually the `async` option to true in order to "change" the client service.

But it seems not a bad idea to allow having both clients.

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -260,8 +264,21 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
 
         // add the default configuration for the S3 filesystem
         if ($container->hasDefinition('sonata.media.adapter.filesystem.s3') && isset($config['filesystem']['s3'])) {
+            $async = true === $config['filesystem']['s3']['async'];
+            if (true === $async && !class_exists(SimpleS3Client::class)) {
```
VincentLanglet: can be simplify
```suggestion
            if ($async && !class_exists(SimpleS3Client::class)) {
```

src/DependencyInjection/SonataMediaExtension.php
```diff
@@ -260,8 +264,21 @@ public function configureFilesystemAdapter(ContainerBuilder $container, array $c
...
+                throw new \RuntimeException('You must install "async-aws/simple-s3" to use async S3 adapter');
+            } elseif (false === $async && !class_exists(S3Client::class)) {
```
VincentLanglet:
```suggestion
            } elseif (!$async && !class_exists(S3Client::class)) {
```

composer.json
```diff
@@ -64,6 +64,7 @@
         "twig/twig": "^2.14 || ^3.0"
     },
     "require-dev": {
+        "async-aws/simple-s3" : "^1.1",
```
VincentLanglet: Can we support `^1.0` ?
We also have to add "aws/aws-sdk-php": "<1.0", in the conflict section.

VincentLanglet: Also take a look at the failing build for the composer linter

ozahorulia: @VincentLanglet are you sure about "aws/aws-sdk-php": "<1.0" conflict? There is already "aws/aws-sdk-php": "<3.0" in the conflict section, and probably it's there for a reason? I'm not sure about it, but SDK's changelog is too huge to make the decision

VincentLanglet: I meant `async-aws/simple-s3 < 1.0` sorry

ozahorulia: @VincentLanglet Sure, makes sense. Added it.

composer.json
```diff
@@ -89,13 +90,13 @@
         "sonata-project/classification-bundle": "^4.0",
         "sonata-project/doctrine-orm-admin-bundle": "^4.0",
         "symfony/browser-kit": "^4.4 || ^5.4 || ^6.0",
-        "symfony/filesystem": "^4.4 || ^5.4 || ^6.0",
```
VincentLanglet, line 92: One last question, why this was removed ?

VincentLanglet (approved):

composer.json
ozahorulia: It's already in the "require" section with the same versions. Does it makes sense to have a duplicate entry in require-dev? I can roll it back if it was there for a reason :)

VincentLanglet: didn't see the require section. You were right to remove it then. thnaks

ozahorulia:
@VincentLanglet Thank you! May I ask you to create a minor release, so I can use it in my projects?

VincentLanglet:
Done

ozahorulia:
Thank you!

PR #2360, opened 2023-03-01
Drop older versions an update deps

jordisala1991:
## Subject

I am targeting this branch, because this should be BC.

## Changelog

```markdown
### Added
- Added support for `knplabs/gaufrette` ^0.11.

### Removed
- Drop support for `doctrine/persistence` ^2.0.
```


jordisala1991:
```
Error: src/Model/Gallery.php:147:14: RedundantCondition: Traversable<int, T:Sonata\MediaBundle\Model\Gallery as Sonata\MediaBundle\Model\GalleryItemInterface> cannot be identical to ArrayIterator (see https://psalm.dev/122)
```

Last error might get silenced too... I think psalm is kind of wrong here, wdyt @VincentLanglet ?

VincentLanglet:
> ```
> Error: src/Model/Gallery.php:147:14: RedundantCondition: Traversable<int, T:Sonata\MediaBundle\Model\Gallery as Sonata\MediaBundle\Model\GalleryItemInterface> cannot be identical to ArrayIterator (see https://psalm.dev/122)
> ```
> 
> Last error might get silenced too... I think psalm is kind of wrong here, wdyt @VincentLanglet ?

ArrayIterator is a SeekableIterator, which is an Iterator which is Traversable.
So I would say psalm is wrong indeed.

Can you open a psalm issue and add a link in the @psalm-suppress ?

jordisala1991:
I can't even reproduce on psalm.dev: https://psalm.dev/r/539e779252

jordisala1991:
I think the problem is somewhere else. Let's add a new commit.

jordisala1991:
It is related with the usage of template params. I have removed them, wdyt @VincentLanglet ? To me those made a little sense, you are not suposed to create galleries of another thing that is not galleryItemInterface, it is even enforced on the interfaces.

VincentLanglet:
> It is related with the usage of template params. I have removed them, wdyt @VincentLanglet ? To me those made a little sense, you are not suposed to create galleries of another thing that is not galleryItemInterface, it is even enforced on the interfaces.

Template are not here to "enforce" passing galleryItemInterface, it's made to be able to understand which galleryItem you are using. This way people are able to write `Gallery<CustomGalleryItem>` and be able to call custom methods on the gallery item. Like when you write `Collection<Foo>` rather than `Collection<object>`.
I dunno how it's useful, but removing template is kinda a BC-break for phpdoc/static analysis ; I would feel sad having to introduce such change just because psalm is messing with it. PHPStan has no errors...

But since it's related to template, the reproducer is https://psalm.dev/r/6f6a50ae98

jordisala1991:
Iterating your example: https://psalm.dev/r/7391c0ce78

I think it comes from the fact that we are doing T of .... in both class and interface.

VincentLanglet:
> Iterating your example: [psalm.dev/r/7391c0ce78](https://psalm.dev/r/7391c0ce78)
> 
> I think it comes from the fact that we are doing T of .... in both class and interface.

Indeed.
To me it's a psalm bug, I would say they try to compare
```
Traversable<int, T:Gallery as GalleryItemInterface>
```
and 
```
ArrayIterator<mixed>
```
to say it's different in the `instanceof` check...

Are you ok with the psalm issue or do you think it's something on our side ?

jordisala1991:
> > Iterating your example: [psalm.dev/r/7391c0ce78](https://psalm.dev/r/7391c0ce78)
> > I think it comes from the fact that we are doing T of .... in both class and interface.
> 
> Indeed. To me it's a psalm bug, I would say they try to compare
> 
> ```
> Traversable<int, T:Gallery as GalleryItemInterface>
> ```
> 
> and
> 
> ```
> ArrayIterator<mixed>
> ```
> 
> to say it's different in the `instanceof` check...
> 
> Are you ok with the psalm issue or do you think it's something on our side ?

I think I don't know enough to know if it is on our side, with the info I have I think it is not, so let's open an issue.

jordisala1991:
Issue created and PR updated @VincentLanglet

VincentLanglet (approved):

PR #2363, opened 2023-03-10
Add galley_item translation key for better SonatUserBundle roles display

gremo:
## Subject

I am targeting this branch, because this is a smally fix.

SonataUserBundle translates admin names. Gallery and media are translated:

Immagine 2023-03-10 130847

Gallery item is not, resulting in a row string:

Immagine 2023-03-10 130904

Immagine 2023-03-10 131008

## Changelog

```markdown
### Added
- Added gallery_item translation key (english)
```

### DIFF
--- a/src/Resources/translations/SonataMediaBundle.en.xliff
+++ b/src/Resources/translations/SonataMediaBundle.en.xliff
@@ -14,6 +14,10 @@
         <source>gallery</source>
         <target>Gallery</target>
       </trans-unit>
+      <trans-unit id="gallery_item">
+        <source>gallery_item</source>
+        <target>Gallery item</target>
+      </trans-unit>
       <trans-unit id="sonata_media">
         <source>sonata_media</source>
         <target>Media Library</target>


src/Resources/translations/SonataMediaBundle.en.xliff
```diff
@@ -14,6 +14,10 @@
         <source>gallery</source>
         <target>Gallery</target>
       </trans-unit>
+      <trans-unit id="gallery_item">
+        <source>gallery_item</source>
+        <target>Gallery item</target>
```
VincentLanglet, line 19: We have the following translation
```
<trans-unit id="form.label_gallery_items">
        <source>form.label_gallery_items</source>
        <target>Media</target>
      </trans-unit>
```
Should we use Media here then ?

gremo: Do you want to translate gallery_item into "Media", that is the same translation of media itself? it would be confusing in sonata user admin then.

VincentLanglet: I don't use this bundle and don't know the difference.
I just notice that if gallery_item is different from media, maybe the translation of `form.label_gallery_items` is wrong.

I think @jordisala1991 use this lib.

gremo: That key is actually used in the gallery admin (it's the **Media*** label):

https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Admin/GalleryAdmin.php#L101

Immagine 2023-03-10 133152

gremo:
Further comment: I don't know how we should label this. Actually it should be something like "Gallery-media association" o relation or something similar 👍 the "Gallery item" itself means nothing to me, but I may be wrong. Feel free to change it as you like.

src/Resources/translations/SonataMediaBundle.en.xliff
jordisala1991: I think this change is correct. This is in order to be changed on SonataUserbundle so better not use Media.

And you don't want the admin label to change on @gremo screenshot 👍

jordisala1991 (approved):

PR #2371, opened 2023-03-25
Use attributes on docs

jordisala1991:
Should be pedantic.

### DIFF
--- a/docs/reference/installation.rst
+++ b/docs/reference/installation.rst
@@ -125,63 +125,69 @@ And then create the corresponding entities, ``src/Entity/SonataMediaMedia``::
 
     // src/Entity/SonataMediaMedia.php
 
+    use Doctrine\DBAL\Types\Types;
     use Doctrine\ORM\Mapping as ORM;
     use Sonata\MediaBundle\Entity\BaseMedia;
 
-    /**
-     * @ORM\Entity
-     * @ORM\Table(name="media__media")
-     */
+    #[ORM\Entity]
+    #[ORM\Table(name: 'media__media')]
     class SonataMediaMedia extends BaseMedia
     {
-        /**
-         * @ORM\Id
-         * @ORM\GeneratedValue
-         * @ORM\Column(type="integer")
-         */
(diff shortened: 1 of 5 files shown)


docs/reference/installation.rst
```diff
@@ -125,63 +125,69 @@ And then create the corresponding entities, ``src/Entity/SonataMediaMedia``::
...
-        /**
-         * @ORM\Id
-         * @ORM\GeneratedValue
-         * @ORM\Column(type="integer")
-         */
-        protected $id;
+        #[ORM\Id]
+        #[ORM\Column(type: Types::INTEGER)]
+        #[ORM\GeneratedValue]
+        protected ?int $id = null;
```
phansys, line 162: IIMO, if we are forcing `id` to be `?int`, we should use the same declaration as return type in `getId()`.

docs/reference/installation.rst
```diff
@@ -125,63 +125,69 @@ And then create the corresponding entities, ``src/Entity/SonataMediaMedia``::
...
+
+        public function getId()
```
jordisala1991:
```suggestion
        public function getId(): ?int
```

docs/reference/installation.rst
```diff
@@ -125,63 +125,69 @@ And then create the corresponding entities, ``src/Entity/SonataMediaMedia``::
...
```
jordisala1991:
```suggestion
        public function getId(): ?int
```

docs/reference/installation.rst
```diff
@@ -125,63 +125,69 @@ And then create the corresponding entities, ``src/Entity/SonataMediaMedia``::
...
```
jordisala1991:
```suggestion
        public function getId(): ?int
```

docs/reference/installation.rst
jordisala1991: You’re right, applied.

VincentLanglet (approved):

phansys (approved):

PR #2372, opened 2023-03-27
Change web to public on next major

### LINKED ISSUE #2341: Local driver directory wrong set by default beginning from Symfony 4
## Subject
Sonata media Configuration has with default value
https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/DependencyInjection/Configuration.php#L195
```
->scalarNode('directory')->defaultValue('%kernel.project_dir%/web/uploads/media')->end()
```
But starting from Symfony 4.4 `web` folder moved to `public`. So the bundle not working as expected

## Possible solution
Use first exiting folder: `web` or `public`

What do you think?

jordisala1991:
## Subject

I am targeting this branch, because this is pedantic, just a comment for next major.

Closes #2341.

### DIFF
--- a/src/DependencyInjection/Configuration.php
+++ b/src/DependencyInjection/Configuration.php
@@ -192,6 +192,7 @@ final class Configuration implements ConfigurationInterface
                         ->arrayNode('local')
                             ->addDefaultsIfNotSet()
                             ->children()
+                                // NEXT_MAJOR: Change 'web' directory for 'public'.
                                 ->scalarNode('directory')->defaultValue('%kernel.project_dir%/web/uploads/media')->end()
                                 ->scalarNode('create')->defaultValue(false)->end()
                             ->end()


jordisala1991:
I dont think we can don much more without breaking bc and I dont think it is worth breaking BC.

VincentLanglet (approved):

VincentLanglet:
I may mis-understand the issue https://github.com/sonata-project/SonataMediaBundle/issues/2341#issue-1408970189 but doesn't it say that the bundle doesn't work anymore with the default config ?

jordisala1991:
That is correct. If you start a new project and install this bundle, you will need to edit this config option, otherwise you may end up with public and web directories and images not loading because your web server will point to public, and not to web.

The other option is to break Bc and deal with people with sf 4.4 and that still uses web directory (that is possible)

VincentLanglet:
>  deal with people with sf 4.4 and that still uses web directory (that is possible)

Didn't know it was possible

Maybe the suggestion could be a smoother way to avoid as many bc break as possible:
- If web folder exist, use it
- Else use public folder

jordisala1991:
> Didn't know it was possible

Yes, it is. When I did the migration from 3.4 to 4.4. I normally kept using web directory for a while and then when I could afford to deal with that change on the server, I did it, but you could use sf 6.2 with web directory if you want.

> Maybe the suggestion could be a smoother way to avoid as many bc break as possible:
> 
> - If web folder exist, use it
> 
> - Else use public folder

that check would be a bit clunky, or not? I would have to check like __DIR__.'../../..' to be outside vendor.

VincentLanglet:
Let's go with this so far

PR #2378, opened 2023-04-08
4.7.1

VincentLanglet:


jordisala1991 (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [4.7.1](https://github.com/sonata-project/SonataMediaBundle/compare/4.7.0...4.7.1) - 2023-04-09
+### Added
```
phansys, line 6: Are we introducing additions and changes in a PATCH release? :thinking:

VincentLanglet: To me it's kinda like "Fixing missing key"

phansys: Should we then move this item to the "Fixed" section?

jordisala1991: Probably yes. More like: Fixed translation of...

VincentLanglet: We're supposed to have a tool to auto-generate the changelog. The PR with this changelog is one month old, there was time to have this discussion or/and update the changelog. The release shouldn't be the place for this... It's kinda annoying having always to rewrite the sentence/changing the section when doing a release, doubling the time it takes for no real value or I would be happy no be the only one doing these PR/release.

Feel free to change what you want about the changelog message...

phansys: I understand your concerns.
AFAIK, even if it might be annoying, this stage is the last chance we have to provide feedback for the changelog included in each release.
In this case, I found time to provide my review today.
I agree that `dev-kit` should be responsible to prevent (or at least, to emit a warning) about the introduction of changes that are not compatible with the version we are trying to release. IMO, we should check the behavior regarding this.
That said, I disagree with your opinion about the value of the changelog. For me, it is important to provide clear and proper changelogs to let users know what was changed and what to expect in our releases.

VincentLanglet: > That said, I disagree with your opinion about the value of the changelog. For me, it is important to provide clear and proper changelogs to let users know what was changed and what to expect in our releases.

With the low number of contributors/maintainers ; there is things more important than changing an "Added" to "Fixed"... Like doing the release. Like I said this wasn't release in one full month. I'm not the only one in the release-manager team.

And the changelog can be edited here: https://github.com/sonata-project/SonataMediaBundle/pull/2363#issue-1618889389

Anyway, I'll drop this release

PR #2404, opened 2023-07-21
SimpleResizer calculation in `round()` can get return invalid value for image resizing.

silasjoisten:
## Subject

I am targeting this branch, because this is a bugfix.

the `round` method in the `SimpleResizer` can return 0 and thats invalid for image resizing.

## Changelog
```markdown
### Fixed
* `round()` can return 0 which is invalid for image resizing.

```

PR #2405, opened 2023-07-21
Fix: SimpleResizer calculation in round() can return invalid value for image resizing. (4.x)

silasjoisten:
## Subject

the `round` method in the `SimpleResizer` can return 0 and thats invalid for image resizing.

## Changelog
```markdown
### Fixed
* `round()` can return 0 which is invalid for image resizing.

```

ref #2404


VincentLanglet (approved):

VincentLanglet:
Thanks

PR #2420, opened 2023-11-19
Fix doctrine mapping

core23:
## Subject

The recent doctrine dbal update introduced a deprecation warning
```
 [FAIL] The entity-class App\Entity\Media\Media mapping is invalid:

 * The field 'App\Entity\Media\Media#length' has the property type 'float' that differs from the metadata field type 'string' returned by the 'decimal' DBAL type.

 [FAIL] The entity-class Sonata\MediaBundle\Entity\BaseMedia mapping is invalid:

 * The field 'Sonata\MediaBundle\Entity\BaseMedia#length' has the property type 'float' that differs from the metadata field type 'string' returned by the 'decimal' DBAL type.

 [FAIL] The entity-class Core23\MediaBundle\Entity\BaseMedia mapping is invalid:

 * The field 'Core23\MediaBundle\Entity\BaseMedia#length' has the property type 'float' that differs from the metadata field type 'string' returned by the 'decimal' DBAL type.
```

## Changelog

```markdown
### Fixed
- Fix doctrine mapping for media length
```

### DIFF
--- a/src/Model/Media.php
+++ b/src/Model/Media.php
@@ -41,7 +41,7 @@ abstract class Media implements MediaInterface, \Stringable
 
     protected ?int $height = null;
 
-    protected ?float $length = null;
+    protected ?string $length = null;
 
     protected ?string $copyright = null;
 
@@ -184,12 +184,12 @@ abstract class Media implements MediaInterface, \Stringable
 
     public function setLength(?float $length): void
     {
-        $this->length = $length;
+        $this->length = null !== $length ? (string) $length : null;
     }
 
     public function getLength(): ?float
     {
-        return $this->length;
+        return null !== $this->length ? (float) $this->length : null;
     }
 
     public function setCopyright(?string $copyright): void


src/Model/Media.php
```diff
@@ -184,12 +184,12 @@ public function getHeight(): ?int
 
     public function setLength(?float $length): void
     {
-        $this->length = $length;
+        $this->length = (string) $length;
```
VincentLanglet:
```suggestion
        $this->length = null !== $length ? (string) $length : null;
```

src/Model/Media.php
```diff
@@ -184,12 +184,12 @@ public function getHeight(): ?int
...
     }
 
     public function getLength(): ?float
     {
-        return $this->length;
+        return (float) $this->length;
```
VincentLanglet:
```suggestion
        return null !== $this->length ? (float) $this->length : null;
```

src/Model/Media.php
VincentLanglet: Maybe this would require a unit test to ensure
```
$media->setLenght(null);
static::assertNull($media->getLength());
```

VincentLanglet (changes requested):

VincentLanglet (approved):

dmitryuk:
Please make a new release after merge

dmitryuk:
@VincentLanglet can you make a new release?

VincentLanglet:
> @VincentLanglet can you make a new release?

Done https://github.com/sonata-project/SonataMediaBundle/releases/tag/4.10.5

PR #2448, opened 2024-03-17
require-dev compatibility for Doctrine ORM 3

dmaicher:
## Subject

I am targeting this branch, because its BC.

Part of https://github.com/sonata-project/dev-kit/issues/2674

## Changelog

```markdown
### Added
- compatibility with ORM 3
```


src/Resources/config/doctrine/BaseMedia.orm.xml
```diff
@@ -14,7 +14,7 @@
         </field>
         <field name="width" column="width" type="integer" nullable="true"/>
         <field name="height" column="height" type="integer" nullable="true"/>
-        <field name="length" column="length" type="decimal" nullable="true"/>
+        <field name="length" column="length" type="decimal" nullable="true" precision="10"/>
```
dmaicher, line 17: This is needed with ORM 3. Otherwise will lead to

```
Doctrine\DBAL\Exception\InvalidColumnType\ColumnPrecisionRequired: Column precision is not specified
```

https://www.doctrine-project.org/projects/doctrine-dbal/en/4.0/reference/schema-representation.html#portable-options

says

> Defaults to 10.

Still going to verify this

dmaicher: Schema create SQL

ORM 2 without specified `precision`: `length NUMERIC(10, 0) DEFAULT NULL` 
ORM 3 with specified `precision = 10`: `length NUMERIC(10, 0) DEFAULT NULL`

 So this change should keep BC

VincentLanglet (approved):

PR #2476, opened 2025-02-21
Use `BIGINT` for `Media::$contentSize`

OskarStark:
Otherwise you cannot upload 3GB files and add the content size

## Changelog

```markdown
### Changed
- Use BIGINT for Media::$contentSize
```

### DIFF
--- a/src/Resources/config/doctrine/BaseMedia.orm.xml
+++ b/src/Resources/config/doctrine/BaseMedia.orm.xml
@@ -16,7 +16,7 @@
         <field name="height" column="height" type="integer" nullable="true"/>
         <field name="length" column="length" type="decimal" nullable="true" precision="10"/>
         <field name="contentType" column="content_type" type="string" nullable="true" length="255"/>
-        <field name="size" column="content_size" type="integer" nullable="true"/>
+        <field name="size" column="content_size" type="integer" length="5" nullable="true"/>
         <field name="copyright" column="copyright" type="string" nullable="true"/>
         <field name="authorName" column="author_name" type="string" nullable="true"/>
         <field name="context" column="context" type="string" nullable="true" length="64"/>
--- a/src/Resources/config/doctrine/BaseMedia.orm.xml
+++ b/src/Resources/config/doctrine/BaseMedia.orm.xml
@@ -16,7 +16,7 @@
         <field name="height" column="height" type="integer" nullable="true"/>
         <field name="length" column="length" type="decimal" nullable="true" precision="10"/>
         <field name="contentType" column="content_type" type="string" nullable="true" length="255"/>
-        <field name="size" column="content_size" type="integer" length="5" nullable="true"/>
+        <field name="size" column="content_size" type="bigint" nullable="true"/>
         <field name="copyright" column="copyright" type="string" nullable="true"/>
         <field name="authorName" column="author_name" type="string" nullable="true"/>
         <field name="context" column="context" type="string" nullable="true" length="64"/>


src/Resources/config/doctrine/BaseMedia.orm.xml
```diff
@@ -16,7 +16,7 @@
         <field name="height" column="height" type="integer" nullable="true"/>
         <field name="length" column="length" type="decimal" nullable="true" precision="10"/>
         <field name="contentType" column="content_type" type="string" nullable="true" length="255"/>
-        <field name="size" column="content_size" type="integer" nullable="true"/>
+        <field name="size" column="content_size" type="integer" length="5" nullable="true"/>
```
VincentLanglet: Updating the existing config would trigger a migration for all existing user no ? Could it be considered as a BC break ?

OskarStark: Yes it will, but we are widening it, so should be ok, also if you don't execute a new migration the code would still work like before.

VincentLanglet: We had some discussion about such schema changes and it was unclear about if it's okay or not to introduce a migration for users https://github.com/sonata-project/SonataUserBundle/pull/1677

I assume that changing array to json is a BC break because the migration is required but maybe changing int to bigint is ok if the code still work without the migration (?).

@core23 @jordisala1991 WDYT ?

OskarStark: Ping @dmaicher

OskarStark: Ping @dmaicher

VincentLanglet:
You can rebase to solve the CI, I made https://github.com/sonata-project/SonataMediaBundle/pull/2477

OskarStark:
Done

src/Resources/config/doctrine/BaseMedia.orm.xml
VincentLanglet: Should we use an even bigger length ?

VincentLanglet (dismissed):
It's fine for me ; but i'm not against another opinion.

VincentLanglet:
Maybe we should do a similar change for the Mongo support to be consistent @OskarStark https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Resources/config/doctrine/BaseMedia.mongodb.xml#L15 ? (Im not familiar if there is length issue with this db)

src/Resources/config/doctrine/BaseMedia.orm.xml
OskarStark: There is an internal mapping inside doctrine where length > 4 maps to BIGINT

OskarStark:
cc @alcaeus @GromNaN

GromNaN:
> Maybe we should do a similar change for the Mongo support to be consistent @OskarStark https://github.com/sonata-project/SonataMediaBundle/blob/4.x/src/Resources/config/doctrine/BaseMedia.mongodb.xml#L15 ? (Im not familiar if there is length issue with this db)

I don't think there is anything to do for MongoDB. To quote the [request for support of 64 bytes int](https://github.com/doctrine/mongodb-odm/issues/2190#issuecomment-635979300):
> There is no special handling necessary to get the driver to store a 64-bit value, as long as you are running on a 64-bit platform.

alcaeus:
Correct. If you're in need of storing files larger than 8 PiB, please use a Decimal128, accepting potential losses in detail.

src/Resources/config/doctrine/BaseMedia.orm.xml
dmaicher: Out of curiosity: Why not use type `bigint`?
```suggestion
        <field name="size" column="content_size" type="bigint" nullable="true"/>
```

OskarStark: Does this work with Postgres?

dmaicher: Hm I guess so :thinking: 

See https://www.doctrine-project.org/projects/doctrine-dbal/en/4.2/reference/types.html#bigint

Not sure which DBAL version we support here and if that makes a difference though

OskarStark:
Thanks @alcaeus and @GromNaN 🙂👍🏻

src/Resources/config/doctrine/BaseMedia.orm.xml
OskarStark: Ok, already supported in DBAL 3, so lets use that

VincentLanglet (approved):

dmaicher (approved):

OskarStark:
Thanks guys, can you please make a release? 🥳

VincentLanglet:
https://github.com/sonata-project/SonataMediaBundle/releases/tag/4.16.0

PR #2490, opened 2025-06-13
Fix async S3 filesystem adapter throwing error if the full AWS SDK is not installed, even though it is not used in this case

zyberspace:
## Subject

When i followed the [S3 setup documentation](https://docs.sonata-project.org/projects/SonataMediaBundle/en/4.x/reference/amazon_s3), I ran into an error that the `sonata.media.adapter.service.s3` does not exist.

As it turns out, even when using the `async-aws/simple-s3` client like described in the documentation and therefore the `sonata.media.adapter.service.s3.async` service, the container builder still expects the `sonata.media.adapter.service.s3` to exist, even though this service is not used in this case.

The full error i received:
> You have requested a non-existent service "sonata.media.adapter.service.s3".

I added a failing test that reproduces this error. If you run the PHPUnit tests with the commit a9ebea4 you can see the failing test and the error described above is thrown.

After verifying that the new test "successfully" fails, i added another commit that fixes this.

I actually had a bit of trouble preventing the AWS SDK S3 service from registering because both clients are installed in the test environment, hence why the tests were also not throwing any errors before. If there is a better way to prevent that service from being registered in the test environment, please let me know.

*(I also renamed the S3 tests from `...V3..` to `...S3...`. If the `V3` name was on purpose, i am happy to change that back)*

I am targeting this branch, because this is a backwards compatible fix.

## Changelog

```markdown
### Fixed
- Using the `async: true` S3 filesystem adapter throws an error if the full AWS SDK is not installed, even though it is not used in this case.
```

### DIFF
--- a/src/Resources/config/gaufrette.php
+++ b/src/Resources/config/gaufrette.php
@@ -76,13 +76,19 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 
         ->set('sonata.media.metadata.noop', NoopMetadataBuilder::class);
 
-    if (class_exists(S3Client::class)) {
+    if (
+        class_exists(S3Client::class)
+        && ($_ENV['PHPUNIT_TESTS_SONATA_MEDIA_FORCE_DISABLE_S3_CLIENT'] ?? 'false') !== 'true'
+    ) {
         $containerConfigurator->services()
             ->set('sonata.media.adapter.service.s3', S3Client::class)
             ->args([abstract_arg('settings')]);
     }
 
-    if (class_exists(SimpleS3Client::class)) {
+    if (
+        class_exists(SimpleS3Client::class)
+        && ($_ENV['PHPUNIT_TESTS_SONATA_MEDIA_FORCE_DISABLE_S3_ASYNC_CLIENT'] ?? 'false') !== 'true'
+    ) {
         $containerConfigurator->services()
             ->set('sonata.media.adapter.service.s3.async', SimpleS3Client::class)
             ->args([abstract_arg('settings')]);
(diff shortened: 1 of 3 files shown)


tests/DependencyInjection/SonataMediaExtensionTest.php
```diff
@@ -271,7 +305,7 @@ public function testLoadWithFilesystemConfigurationV3(
     /**
      * @phpstan-return iterable<array{array<string, mixed>, array<string, mixed>}>
      */
-    public function provideLoadWithFilesystemConfigurationV3Cases(): iterable
+    public function provideLoadWithFilesystemConfigurationS3Cases(): iterable
```
phansys, line 308: "V3" refers to the SDK's version.

zyberspace: Hey, thanks for the reply. So i drop the name change?

Maybe something like `provideLoadWithS3FilesystemConfigurationV3Cases` would work?

Tests would then be `testLoadWithS3FilesystemConfigurationV3` and `testLoadWithS3FilesystemConfigurationAsync`.

src/Resources/config/gaufrette.php
```diff
@@ -76,13 +76,19 @@
 
         ->set('sonata.media.metadata.noop', NoopMetadataBuilder::class);
 
-    if (class_exists(S3Client::class)) {
+    if (
+        class_exists(S3Client::class)
+        && ($_ENV['PHPUNIT_TESTS_SONATA_MEDIA_FORCE_DISABLE_S3_CLIENT'] ?? 'false') !== 'true'
```
VincentLanglet, line 81: This feels odd having a code/env value for test in prod config. Isn't any other way to do it @zyberspace ?

VincentLanglet:
If you rebase, the rector build will be fixed.

PR #2492, opened 2025-08-27
Add support for Azure storage

jorrit:
## Subject

I need support for Azure Blob Storage. I copied the S3 code and modified it a bit to connect using the Gaufrette Azure adapter. I know that Gaufrette is deprecated and Sonata Media should migrate to Flysystem, but this had to be done quickly.

The Gaufrette Azure adapter has different behavior for setting meta data compared to S3, so I left that out for now. With Azure, the meta data should be set after uploading the blob.

## Changelog

```markdown
### Added
- Basic support for using Azure Blob Storage as filesystem
```


VincentLanglet (commented):
Thanks, could you add some test ? :)

jorrit:
> Thanks, could you add some test ? :)

Done, insofar possible with an external service.

docs/reference/advanced_configuration.rst
```diff
@@ -113,6 +113,11 @@ Full configuration options:
                     key1: value1 #any amount of metas(sent as x-amz-meta-key1 = value1)
                 async: false
 
+            azure:
+                containerName:
+                connectionString:
```
VincentLanglet: Except accessKey and secretKey, I feel like we're using snake case everywhere.
I think those "key" were a mistake and we should stick with snake case, WDYT ?

So container_name and connection_string (and create_container)

VincentLanglet (changes requested):

VincentLanglet:
Also, please rebase

dmaicher (commented):
:+1: LGTM if you run CS fixer and the build is green

dmaicher (approved):

VincentLanglet (approved):

jorrit:
Thanks all!

PR #2499, opened 2025-09-30
Deprecate MediaProviderInterface::validate()

dmaicher:
## Subject

I am targeting this branch, because {reason}.

Relates to https://github.com/sonata-project/form-extensions/issues/333

## Changelog

```markdown
### Deprecated
- `MediaProviderInterface::validate()` is deprecated. Implement `validateMedia()` instead. 
```

## To do

- [ ] Add more tests
- [ ] documentation update?

@sonata-project/contributors this is a draft for now. Before I continue to finalize it I wanted to get your feedback on it.

To achieve https://github.com/sonata-project/form-extensions/issues/333 we need to get rid of `ErrorElement` and `InlineConstraint` usages here.

From what I see we can easily achieve the validation with the proposed approach here by introducing a new `validateMedia()` method.

WDYT?


src/Provider/FileProvider.php
```diff
@@ -250,8 +250,11 @@ public function getDownloadResponse(MediaInterface $media, string $format, strin
         );
     }
 
-    public function validate(ErrorElement $errorElement, MediaInterface $media): void
+    public function validateMedia(ExecutionContextInterface $context, MediaInterface $media): void
```
VincentLanglet, line 253: There is no BC break issue if someone is extending the validate method ?

VincentLanglet (commented):
Seems an interesting solution, the only issue I see
- If it's really Backward compatible for someone overriding validate method (and eventually rely on parent::validate)

- We have BlockBundle using it a lot https://github.com/search?q=repo%3Asonata-project%2FSonataBlockBundle%20ErrorElement&type=code and every bundle implementing those interface. Like SonataPageBundle or SonataClassificationBundle.

- More important, with our lack of maintainer/contributor, I dunno if there will be a new major one day and when...

src/Provider/FileProvider.php
dmaicher: It should be compatible as I added those checks if a child class overwrites them inside `BaseProvider` :thinking:

dmaicher:
>     * If it's really Backward compatible for someone overriding validate method (and eventually rely on parent::validate)

Should be. See https://github.com/sonata-project/SonataMediaBundle/pull/2499#discussion_r2391993200


>     * We have BlockBundle using it a lot https://github.com/search?q=repo%3Asonata-project%2FSonataBlockBundle%20ErrorElement&type=code and every bundle implementing those interface. Like SonataPageBundle or SonataClassificationBundle.

Yeah that would indeed be the bigger work to deprecate and fix those :cry: 

>     * More important, with our lack of maintainer/contributor, I dunno if there will be a new major one day and when...

Valid point. Maybe its not worth it for now anyway. I don't use this bundle myself anyway.

PR #2511, opened 2026-01-03
Replace Gaufrette with Flysystem

### LINKED ISSUE #1985: Replace Gauffrete?
## Feature Request

I have been looking at gauffrete and I don't know if it is still maintained or we should move away from it.

There are other very popular alternatives like: https://github.com/thephpleague/flysystem

IMHO looking at the future. The best would be to not take care for the FileSystem. I mean, if we are integrating with 3rd party bundle/package we should provide only the default filesystem integration and provide a way to configure an alternative filesystem service. This service can be configured with the 3rd party tool.

This way we don't end up having to replicate every single option of the different adapters, like: https://github.com/sonata-project/SonataMediaBundle/pull/1817

This is probably something to tackle for 5.0, since we don't have many resources to do it here, and there are other importants things to do.

WDYT @sonata-project/contributors . Do you know a better alternative to gauffrete? Do you use it for your own projects?

jorrit:
## Subject

This PR replaces the Gaufrette filesystem abstraction layer with Flysystem in a non-backwards compatible way.

I chose to also remove the metadata code, I hope this can be replaced by configuring Flysystem directly. Maybe I have to revert this.
I also could not port the download mode code (`http`, `X-Sendfile`, `X-Accel-Redirect`) because Flysystem does not allow access to the real file path for local adapters.

I am targeting this branch, because it is very backwards incompatible.

Closes #1985

## Changelog

```markdown
### Changed
- Filesystem abstraction layer from Gaufrette to Flysystem

### Removed
- Metadata generator
- download mode setting
```

## To do

- [x] Update the tests;
- [ ] Update the documentation;
- [ ] Add an upgrade note.


jorrit:
@VincentLanglet @jordisala1991 This is a work in progress, but I am hoping you could already provide some feedback on my approach. Thanks!

PR #2513, opened 2026-01-29
fix: don't call s3 reference while s3.async is used

treehousedevelopers:
## Subject

I am targeting this branch, to fix configuration issue while S3 is used with async client. In this case there is no reference `sonata.media.adapter.service.s3` and should not be called in configuration.

## Changelog

[fix: don't call s3 reference while s3.async is used](https://github.com/sonata-project/SonataMediaBundle/commit/0737cf3c84982b90910cd40f3bd2ff5b9a166e6c)


treehousedevelopers:
I think, it duplicates : https://github.com/sonata-project/SonataMediaBundle/pull/2490, I missed it.

PR #2514, opened 2026-01-29
Fix: check interface instead of class

jorenMartens:
## Subject

Check interface instead of class

I am targeting this branch, because this is a backwards compatible fix.

## Changelog

```markdown
### Fixed
- Check FileProviderInterface instead of FileProvider when adding providers in CleanMediaCommand
```


esserj:
Great!, this wil fix the issue where decorating the file provider would result in all images being deleted

